PR #3, opened 2022-04-19
Next-Build Change to Src Directory

amponce:
In order to provide a clean developer experience we separated the app from the root by adding a src folder to contain the files.


timcosgrove (approved):
r

PR #5, opened 2022-04-19
Incorporate new next-build client with custom fetcher for proxy.

timcosgrove:
This takes advantage of the new `client` class in next-build 1.3, which allows us to pass in a custom fetcher. This allows us to conditionally route the API requests through our local proxy, when that would be warranted (when the remote is behind the firewall).

This also adds the same self-signed cert that is used in content-build for a similar purpose. I followed the pattern from there of committing the cert. We may want to remove this in the future.


src/pages/[...slug].tsx
```diff
@@ -20,26 +16,21 @@ export default function NodePage({ node }) {
...
 
 export async function getStaticProps(context) {
 
-    const type = await getResourceTypeFromContext(context);
-    if (!type) {
-        return {
-            notFound: true,
-        };
-    }
+    const type = 'node--q_a';
```
Comment (timcosgrove) [inline src/pages/[...slug].tsx:26]: In the other cases, the previously available methods are now part of `client`, but there was no replacement for `getResourceTypeFromContext` that I could see.

amponce (approved):
🚀

PR #7, opened 2022-04-20
Vacms 8638 benefits hub links

timcosgrove:
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/8638


src/components/partials/benefitHubsLinks/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+        return (
+            <>
+                <section className="vads-u-padding-y--3 vads-u-display--flex vads-u-flex-direction--column vads-u-padding-x--1 large-screen:vads-u-padding-x--0">
+                    <h2 className="vads-u-margin-y--0 vads-u-font-size--h3">
+                        VA benefits
+                    </h2>
+
+                    <ul className="usa-unstyled-list" role="list">
+                        {nodes.map((node, index) => (
+                            <li key={index} className="vads-u-margin-y--2">
```
amponce, line 17: `{index}` is fine for this case. We also have uuidv4 installed as an alternate if needed
`import { v4 as uuidv4 } from 'uuid';`
```
// uuidv4(); // ⇨ '9b1deb4d-3b7d-4bad-9bdd-2b0d7b3dcb6d'
<li key={uuidv4()} className="vads-u-margin-y--2">
```

src/pages/demo/partials/benefitHubsLinks.tsx
```diff
@@ -0,0 +1,42 @@
+import { drupalClient } from '@/utils/drupalClient'
+import BenefitsHubLinks from '@/components/partials/benefitHubsLinks'
+import { DrupalJsonApiParams } from 'drupal-jsonapi-params'
+
+const DemoPage = ({ nodes }) => {
+    return nodes.map((node) => (
```
amponce, line 6: might just want to null check nodes in case it's empty

amponce (approved):

src/components/partials/benefitHubsLinks/index.tsx
timcosgrove: Each node has a uuid; maybe we should use that?

src/components/partials/benefitHubsLinks/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+        return (
+            <>
+                <section className="vads-u-padding-y--3 vads-u-display--flex vads-u-flex-direction--column vads-u-padding-x--1 large-screen:vads-u-padding-x--0">
+                    <h2 className="vads-u-margin-y--0 vads-u-font-size--h3">
+                        VA benefits
+                    </h2>
+
+                    <ul className="usa-unstyled-list" role="list">
+                        {nodes.map((node, index) => (
+                            <li key={node.id} className="vads-u-margin-y--2">
```
timcosgrove, line 17: Changed this to use the node's own uuid

src/pages/demo/partials/benefitHubsLinks.tsx
```diff
@@ -0,0 +1,44 @@
+import { drupalClient } from '@/utils/drupalClient'
+import BenefitsHubLinks from '@/components/partials/benefitHubsLinks'
+import { DrupalJsonApiParams } from 'drupal-jsonapi-params'
+
+const DemoPage = ({ nodes }) => {
+    if (nodes) {
```
timcosgrove, line 6: Wrapped in check for data.

PR #13, opened 2022-04-29
node--person_profile and paragraph--staff_profile 

amponce:
**PR includes**: WIP

- paragraph--staff_profile 
- node--person_profile


src/components/node/person_profile/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+                Download full bio (PDF)
+              </a>
+            </div>
+          </div>
+        )}
+      </article>
+      <hr />
+    </div>
+  )
+}
```
amponce, line 118: Person profile component with conditionals to match [person_profile](https://github.com/department-of-veterans-affairs/content-build/blob/master/src/site/layouts/person_profile.drupal.liquid) layout on content-build

src/pages/demo/partials/personProfiles.tsx
```diff
@@ -0,0 +1,57 @@
+import { drupalClient } from '@/utils/drupalClient'
+import { PersonProfile } from '@/components/node/person_profile'
+import { DrupalJsonApiParams } from 'drupal-jsonapi-params'
+import Container from '@/components/container'
+
+const PersonProfilePage = ({ nodes }) => {
+  console.log(nodes)
```
amponce, line 7: need to remove

src/pages/demo/partials/personProfiles.tsx
```diff
@@ -0,0 +1,57 @@
...
+        'filter[status][value]': '1',
+      },
+    }
+  )
+  return {
+    props: {
+      nodes: nodes || null,
+    },
+  }
+}
```
amponce, line 57: I might consider rethinking how this works because staff profile has redundant logic, I wonder if there is a way to just have one page render the node data for both staff profile and person profile

src/pages/demo/partials/staffProfiles.tsx
```diff
@@ -0,0 +1,60 @@
...
+    <>
+      <Container className="container">
+        <div id="content" className="interior">
+          <main className="va-l-detail-page va-facility-page">
+            <div className="usa-grid usa-grid-full">
+              <div className="usa-width-three-fourths">
+                {nodes.map((node, index) => (
+                  <div key={index}>
+                    <StaffProfiles node={node} />
+                    <StaffNewsProfiles node={node} />
```
amponce, line 21: Added news profile to this section for demo purposes, we need to think about how we want to showcase the components we build

.prettierrc.json
```diff
@@ -2,5 +2,7 @@
   "trailingComma": "es5",
   "tabWidth": 2,
   "semi": false,
-  "singleQuote": true
+  "singleQuote": true,
+  "bracketSpacing": true,
+  "proseWrap": "preserve"
```
amponce, line 7: I was having some formatting issues when I saved so I added proseWrap preserve, let me know if this causes any issues.

.github/workflows/test.yml
```diff
@@ -1,35 +1,35 @@
...
-      - '*'
-  pull_request:
-    branches:
-      - '*'
+push:
+branches:
+- '*'
+pull_request:
+branches:
+- '*'
```
timcosgrove, line 8: In general for this file: are the changes intentional? It looks like there are structural changes to the yaml file here.

src/components/paragraph/staff_profile/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  return (
+    <div key={uuid}>
+      <div className="authored-by-line vads-u-margin-bottom--0p5 vads-u-font-weight--bold">
+        {title} {fieldDescription ? fieldDescription : null}
+      </div>
+      <hr />
+    </div>
+  )
+}
```
timcosgrove, line 111: I feel like this better belongs in components/node/person_profile. The data that will be passed to this is a node or portion of node data. When we render a paragraph, it will receive paragraph data.

src/components/paragraph/staff_profile/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+export const StaffNewsProfiles = ({ node }): JSX.Element => {
+  if (!node) return
+  const { uuid, title, field_description: fieldDescription } = node
+
+  return (
+    <div key={uuid}>
```
timcosgrove, line 104: `node.id` will always be the `uuid` for the node. We could use that here, though as I understand it, React mostly just cares that it has an ID for each thing it's iterating over, right?

src/components/node/person_profile/index.tsx
```diff
@@ -0,0 +1,118 @@
+import Image from '@/components/image'
+import { DEV_PATH } from '@/lib/constants'
```
timcosgrove, line 2: I think this is fine for completing the component but I'd like to at least mark one of the uses of `DEV_PATH` here for removal, maybe with a simple `@todo`. We'll want to handle image paths differently eventually.

src/components/paragraph/staff_profile/index.tsx
```diff
@@ -0,0 +1,111 @@
+import Image from '@/components/image'
+import { DEV_PATH } from '@/lib/constants'
+
+export const StaffProfiles = ({ node }): JSX.Element => {
```
timcosgrove, line 4: So, a slightly tricky thing here is that, if we're rendering out a staff_profile _paragraph_, what is actually going to get passed to this component is paragraph data. The node that actually contains the person's name, biography, phone number, etc, is attached to a field on that paragraph.

The existing liquid template deals with this in a sort of simple way, by drilling into that field value to get the attached node and assigning it to a variable:
https://github.com/department-of-veterans-affairs/content-build/blob/master/src/site/paragraphs/staff_profile.drupal.liquid#L1 

We could do something similar here: 
```
export const StaffProfiles = ({ paragraph }): JSX.Element => {
  const node = paragraph.field_staff_profile;
  if (!node) return;
  <snip>
```

Then the rest of the component operates as-is.

src/pages/demo/staffProfiles/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+  StaffNewsProfiles,
+} from '@/components/paragraph/staff_profile'
+import { DrupalJsonApiParams } from 'drupal-jsonapi-params'
+import Container from '@/components/container'
+
+const StaffProfilePage = ({ nodes }) => {
+  if (!nodes) return
+  return (
+    <>
+      <Container className="container">
```
timcosgrove, line 13: I think it would be good to provide a divider between each iteration, just to be clear where a pair begins and ends. An `<hr />` would be fine.

src/pages/demo/personProfiles/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+    .addFilter('status', '1')
+    .addFilter('field_office', null, 'IS NOT NULL')
+    .addFilter('field_media', null, 'IS NOT NULL')
+    .addInclude(['field_office', 'field_media'])
+
+  const nodes = await drupalClient.getResourceCollectionFromContext(
+    'node--person_profile',
+    context,
+    {
+      params: {
```
timcosgrove, line 43: Note that this is the same issue I found I was doing incorrectly earlier: this creates the `params` variable above but then doesn't use it. If we want to use the `DrupalJsonApiParams()` object (and I think it might be easier to work with than constructing the params directly as JSON), you could do the things here - adding the thumbnail, the media image, etc - up above, and then replace this with:

```
params.getQueryObject()
```

which will generate an object like this.

src/pages/demo/personProfiles/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+    </>
+  )
+}
+
+export default PersonProfilePage
+
+export async function getStaticProps(context) {
+  const params = new DrupalJsonApiParams()
+  params
+    .addFilter('status', '1')
```
timcosgrove, line 34: See below, re: this `params` variable.

src/pages/demo/personProfiles/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+    .addFilter('field_office', null, 'IS NOT NULL')
+    .addFilter('field_media', null, 'IS NOT NULL')
+    .addInclude(['field_office', 'field_media'])
```
timcosgrove, line 37: It may be good for us to start putting limits on how many objects we retrieve for these demo pages. We can add this with `addPage()`, i.e. `.addPage(10)` to limit ourselves to 10 objects.

timcosgrove (commented):
A lot of good work here; just a few questions and comments.

.github/workflows/test.yml
amponce: This is an issue with prettier when I run format, I'm not sure I'll compare the changes before this was modified and revert back.

src/components/paragraph/staff_profile/index.tsx
amponce: Yeah, it primarily cares that it's a unique identifier of some kind, I decided to sub in uuidv4 for this one

timcosgrove (approved):

PR #16, opened 2022-05-10
VA-8676 Component Media--image 

amponce:
The media image component is built to accept an image style that allows users to pass a prop to modify the image size.

- 1_1_square_medium_thumbnail:
- 2_1_large 
- 2_1_medium_thumbnail
- 2_3_medium_thumbnail
- 3_2_medium_thumbnail
- 7_2_medium_thumbnail
- full_content_width
- large
- self
- viewport_width

mediaComponent

### DIFF
--- a/src/components/media/index.test.tsx
+++ b/src/components/media/index.test.tsx
@@ -1,0 +1,57 @@
+import { screen, render } from '@testing-library/react'
+import '@testing-library/jest-dom/extend-expect'
+import '@testing-library/jest-dom'
+import { MediaImage } from '.'
+
+// Language: typescript
+// Path: src/components/media/index.test.tsx
+
+const media = {
+  image: {
+    uri: {
+      url: '/sites/default/files/hub_promos/pension.png',
+    },
+    links: {
+      full_content_width: {
+        href: 'https://prod.cms.va.gov/sites/default/files/styles/full_content_width/public/hub_promos/pension.png',
+        meta: {
+          height: '304',
+          width: '456',
+        },
+      },
+      large: {
+        href: 'https://prod.cms.va.gov/sites/default/files/styles/large/public/hub_promos/pension.png',
+        meta: {
+          height: '304',
+          width: '456',
+        },
+      },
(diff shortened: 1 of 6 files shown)


src/pages/demo/partials/benefitHubsLinks.tsx
```diff
@@ -19,24 +25,21 @@ const DemoPage = ({ nodes }) => {
...
 export default DemoPage
 
-export async function getStaticProps(context) {
+export async function getStaticProps(
+  context: GetStaticPropsContext
+): Promise<GetStaticPropsResult<BenefitsHubLinksProps>> {
   const params = new DrupalJsonApiParams()
   params
     .addFilter('status', '1')
     .addFilter('field_related_benefit_hubs', null, 'IS NOT NULL')
```
penny-lischer, line 34: Instead of 'IS NOT NULL' you can use Operators.isNotNull

amponce: perfect, nice find we should probably go back and update the some of the other original components with that fix

src/components/media/index.test.tsx
```diff
@@ -0,0 +1,57 @@
+import { screen, render } from '@testing-library/react'
+import '@testing-library/jest-dom/extend-expect'
+import '@testing-library/jest-dom'
+import { MediaImage } from '.'
+
+// Language: typescript
+// Path: src/components/media/index.test.tsx
+
+const media = {
```
penny-lischer, line 9: Do we want to name constants uppercase?

src/components/media/index.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+    resourceIdObjMeta: {
+      alt: 'pension',
+      title: 'pension',
+      height: '304',
+      width: '456',
+    },
+  },
+}
+
+test('<MediaImage> renders', () => {
```
penny-lischer, line 39: Should wrap the tests in a describe

PR #17, opened 2022-05-11
[WIP] 8610 paragraph--button implementation

penny-lischer:
Screen Shot 2022-05-11 at 2 27 43 PM

### DIFF
--- a/src/components/paragraph/button/index.tsx
+++ b/src/components/paragraph/button/index.tsx
@@ -1,0 +1,23 @@
+import { v4 as uuidv4 } from 'uuid'
+import Link from 'next/link'
+
+const Button = ({ paragraph }): JSX.Element => {
+  if (!paragraph) return
+
+  const {
+    field_button_label: fieldButtonLabel,
+    field_button_link: fieldButtonLink,
+  } = paragraph
+
+  return (
+    <div key={uuidv4()}>
+      <Link href={fieldButtonLink ? fieldButtonLink.uri : ''} passHref>
+        <a className="vads-c-action-link--blue">
+          {fieldButtonLabel ? fieldButtonLabel : ''}
+        </a>
+      </Link>
+    </div>
+  )
+}
+
+export default Button
--- a/src/components/paragraph/button/index.tsx
+++ b/src/components/paragraph/button/index.tsx
@@ -1,8 +1,15 @@
 import { v4 as uuidv4 } from 'uuid'
 import Link from 'next/link'
 
+function isRequestValid(paragraph: any) {
+  return (
+    paragraph.field_button_label !== null &&
+    paragraph.field_button_link.uri !== null
+  )
+}
+
 const Button = ({ paragraph }): JSX.Element => {
-  if (!paragraph) return
+  if (!paragraph || !isRequestValid(paragraph)) return
 
   const {
     field_button_label: fieldButtonLabel,
--- a/src/components/paragraph/button/index.tsx
+++ b/src/components/paragraph/button/index.tsx
@@ -12,15 +12,12 @@ function isRequestValid(paragraph: any) {
 const Button = ({ paragraph }): JSX.Element => {
   if (!paragraph || !isRequestValid(paragraph)) return
 
-  const {
-    field_button_label: fieldButtonLabel,
-    field_button_link: fieldButtonLink,
-  } = paragraph
-
   return (
-    <div key={uuidv4()}>
-      <Link href={fieldButtonLink?.uri} passHref>
-        <a className="vads-c-action-link--blue">{fieldButtonLabel}</a>
+    <div key={paragraph.id}>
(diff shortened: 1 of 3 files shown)


src/pages/demo/button/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+const ButtonPage = ({ buttons }: ButtonPageProps) => {
+  console.log('buttons ', buttons)
+  if (!buttons) buttons = []
+
+  // TODO: Remove once the DrupalJsonApiParams can be implemented
+  const filteredButtons = _.filter(buttons, function (button) {
+    return (
+      button.field_button_label !== null && button.field_button_link !== null
+    )
+  })
```
amponce, line 24: Not a big deal but I think we can avoid and just filter out anything missing a label if don't get the DrupalJsonApiParams working

```
  const filteredButton = buttons.filter(
    ({ field_button_label }) => field_button_label !== null
  )
```

src/pages/demo/button/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+export async function getStaticProps(
+  context: GetStaticPropsContext
+): Promise<GetStaticPropsResult<ButtonPageProps>> {
+  const params = new DrupalJsonApiParams()
+  params
+    // .addGroup('button_group', Operators.and)
+    // .addFilter('field_button_link', null, Operators.isNotNull, 'button_group')
+    // .addFilter('field_button_label', null, 'IS NOT NULL', 'button_group')
+    .addPageLimit(20)
+
```
amponce, line 48: I just spent 20 min testing these and couldn't find a combination that worked, maybe it only works on included fields?

src/components/paragraph/button/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+    field_button_link: fieldButtonLink,
+  } = paragraph
+
+  return (
+    <div key={uuidv4()}>
+      <Link href={fieldButtonLink ? fieldButtonLink.uri : ''} passHref>
+        <a className="vads-c-action-link--blue">
+          {fieldButtonLabel ? fieldButtonLabel : ''}
+        </a>
+      </Link>
```
amponce, line 18: I wonder if filtering null values in the parent means we won't need the ternaries anymore? On the other hand if the Button component is being used by a different page the dev doesn't filter null values early on it could cause some broken links so maybe leaving them in is a better idea, idk?

amponce:
Nice work! This looks solid, just a few comments, nothing blocking

src/pages/demo/button/index.tsx
penny-lischer: Same. I couldn't find a way even tried doing it via the inline params and that didn't work either.

amponce (approved):
:shipit:  Love the test! 💪

src/components/paragraph/button/index.tsx
penny-lischer: @amponce I added more validation and removed the ternaries.

src/components/paragraph/button/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+const Button = ({ paragraph }): JSX.Element => {
+  if (!paragraph || !isRequestValid(paragraph)) return
+
+  const {
+    field_button_label: fieldButtonLabel,
+    field_button_link: fieldButtonLink,
+  } = paragraph
+
+  return (
+    <div key={uuidv4()}>
```
timcosgrove, line 21: `paragraph.id` exists and would also work here (or add it to the destructuring). Any Drupal data structure passed in will have its own uuid at `*.id`.  It might make sense to just use that.

src/components/paragraph/button/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+    paragraph.field_button_link !== null &&
+    paragraph.field_button_link?.uri !== null
+  )
+}
+
+const Button = ({ paragraph }): JSX.Element => {
+  if (!paragraph || !isRequestValid(paragraph)) return
+
+  const {
+    field_button_label: fieldButtonLabel,
```
timcosgrove, line 16: I don't object to this pattern, but, I also don't know if it's necessary.

Would it be possible create a method that would destructure an object into consts from its properties like this and convert them to camelCase at the same time? i.e. something 'magic'. I don't think we need it for this ticket, but I wonder how possible that would be.

timcosgrove (approved):
If you could change the `paragraph.id` bit that would be great, and then this is gtg.

src/components/paragraph/button/index.tsx
penny-lischer: Done

src/components/paragraph/button/index.tsx
penny-lischer: Removed

timcosgrove (approved):

PR #21, opened 2022-05-14
First pass for news_story / generalized node component framework.

timcosgrove:
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/8649


src/pages/[...slug].tsx
```diff
@@ -4,80 +4,69 @@ import {
...
+      resourceTypes,
       context,
       {
-        params: {
-          filter: {
-            'field_site.meta.drupal_internal__target_id':
-              process.env.DRUPAL_SITE_ID,
-          },
-        },
+        params: {},
```
Comment (timcosgrove) [inline src/pages/[...slug].tsx:32]: Unsure of the need for this; removed for the moment.

src/components/node/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+/** We expect each Node component to define and export information about itself to help with discoverability. */
+export interface NodeMetaInfo {
+  resource: string
+  component: ({ node, viewMode, ...props }: NodeProps) => JSX.Element
+  params: DrupalJsonApiParams
+}
+
+/** Collect all imported node meta information. */
+const nodeMetaIn: NodeMetaInfo[] = [NewsStoryMeta]
```
penny-lischer, line 19: Was it intentional to have NewsStoryMeta here?

src/components/node/news_story/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+type NodeNewsStoryProps = {
+  node: NodeNewsStory
+  viewMode?: string
+}
+
+/** Full page news story. */
+export const NewsStoryFull = ({ node }: NodeNewsStoryProps) => {
+  /** Type narrowing; if we've managed to end up here with the wrong data, return. */
+  if (node.type !== 'node--news_story') return
+  const dateTime = new Date(node.created).toLocaleDateString('en-us', {
```
penny-lischer, line 19: May want to consider having a DATE_FORMAT object in a constants file that would list multiple date formats. Are there many different date formats throughout the site being used?

penny-lischer: I will refactor this after merge.

penny-lischer (approved):
LGTM

src/components/node/index.tsx
timcosgrove: Yes, this will eventually collect all node 'meta' information. We would like to be able to look up component and such by resource, i.e. `node--news_story`, which we get from the JSON:API response, points to component `NodeNewsStory` for its rendering.

src/components/node/news_story/index.tsx
timcosgrove: I am into this idea. Let's refactor.

PR #22, opened 2022-05-16
VACMS-8614 Component: paragraph--email_contact

penny-lischer:
Screen Shot 2022-05-16 at 1 53 25 PM

### DIFF
--- a/src/pages/demo/emailContacts/index.tsx
+++ b/src/pages/demo/emailContacts/index.tsx
@@ -1,0 +1,45 @@
+import { v4 as uuidv4 } from 'uuid'
+import { drupalClient } from '@/utils/drupalClient'
+import { GetStaticPropsContext, GetStaticPropsResult } from 'next'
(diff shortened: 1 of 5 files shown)


src/pages/demo/emailContacts/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+}
+
+const EmailContactsPage = ({ emailContacts }: EmailContactsPageProps) => {
+  if (!emailContacts) emailContacts = []
+
+  return (
+    <>
+      <Container className="container">
+        {emailContacts.map((paragraph) => (
+          <EmailContact key={uuidv4()} paragraph={paragraph} />
```
timcosgrove, line 20: `paragraph.id` or `node.id` is always present and is a uuid. If there aren't problems with using that as the key, we could just use that and not use the library.

timcosgrove (approved):
Looks good; one comment which doesn't necessarily need fixing, but just a question.

PR #29, opened 2022-05-19
Add unit test to news story

amponce:
- Added unit test to test that the news story mounts with the correct view mode param passed to the component

- Created mock json file for the test

Open questions: where do we want mock json to live? should it be relative to the component?


penny-lischer (approved):
LGTM

PR #30, opened 2022-05-20
VACMS package updates

amponce:
Updated packages and removed UUID in some places

- "@department-of-veterans-affairs/component-library": "^10.3.0",
- "@testing-library/jest-dom": "^5.16.4",
- "jest-axe": "^6.0.0",
- "next": "12.1.6",
- "react": "18.1.0",
- "react-dom": "18.1.0",
 ...


tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es2015",
```
amponce, line 3: https://microsoft.github.io/TypeScript-New-Handbook/reference/compiler-options/#compiler-options
According to microsoft (All modern browsers support all ES6 features, so ES6 is a good choice) 
-Additional APIs available in ES2015 (also known as ES6) - array.find , Promise , Proxy , Symbol

penny-lischer (approved):
LGTM

PR #31, opened 2022-05-20
VACMS 8616 - Component: paragraph--link_teaser

penny-lischer:


src/pages/demo/linkTeasers/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+const LinkTeaserPage = ({ linkTeasers }: LinkTeaserPageProps) => {
+  if (!linkTeasers) linkTeasers = []
+
+  return (
+    <>
+      <Container className="container">
+        {linkTeasers.map((paragraph) => (
+          <ul key={paragraph.id} className="usa-unstyled-list">
```
penny-lischer, line 19: Needed to add <ul> since this component is typically consumed by list_of_link_teasers and will not render exact without it.

src/components/paragraph/link_teaser/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+const LinkTeaser = ({ paragraph, boldTitle, sectionHeader }): JSX.Element => {
+  if (!paragraph || !isRequestValid(paragraph)) return
+
+  const isFieldSpokes = paragraph.parent_field_name === 'field_spokes'
+  const fieldLinkOption = get(paragraph.field_link.options, ['target'], '')
+
+  if (isFieldSpokes || boldTitle) {
+    return (
+      <li key={paragraph.id} className="hub-page-link-list__item">
```
penny-lischer, line 16: @timcosgrove and @amponce we may need to call the recordEvent method here to track google analytics. Need to follow up.

src/components/paragraph/link_teaser/index.test.tsx
```diff
@@ -0,0 +1,189 @@
...
+        paragraph={MOCK_PARAGRAPH}
+        boldTitle={false}
+        sectionHeader=""
+      />
+    )
+    const liEl = container.querySelector('li')
+
+    expect(liEl).not.toBeInTheDocument()
+  })
+})
```
amponce, line 189: Really nice work on the test!

src/pages/demo/linkTeasers/index.tsx
amponce: Would it be better to move the `ul` outside of the the map so it's only generating list items or is that not how this component works?

amponce (approved):

src/pages/demo/linkTeasers/index.tsx
penny-lischer: Good catch! will do

PR #33, opened 2022-05-23
Teaser News Story Pagination

amponce:
Pagination on news story listing spike. I was able to get this working using the component library 
`@department-of-veterans-affairs/component-library/Pagination`

The news story teaser component will be replaced by @penny-lischer's work on Component: node--story_listing #8652

Screen Shot 2022-05-23 at 4 53 05 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "@fortawesome/free-solid-svg-icons": "^6.1.1",
     "@fortawesome/react-fontawesome": "^0.1.18",
     "@testing-library/jest-dom": "^5.16.4",
+    "classnames": "^2.3.1",
     "clsx": "^1.1.1",
     "cross-fetch": "^3.1.5",
     "drupal-jsonapi-params": "^2.0.0",
--- a/src/components/node/news_story/index.tsx
+++ b/src/components/node/news_story/index.tsx
@@ -24,7 +24,7 @@ import { NodeNewsStory } from '@/types/node'
 /**
  * These components expect NodeNewsStory as their input.
  */
-type NodeNewsStoryProps = {
+export type NodeNewsStoryProps = {
   node: NodeNewsStory
   viewMode?: string
 }
@@ -84,13 +84,56 @@ export const NewsStoryFull = ({ node }: NodeNewsStoryProps) => {
   )
 }
 
+export const NewsStoryTeaser = ({ node }: NodeNewsStoryProps) => {
+  return (
+    <article
+      data-template="teasers/news_story_page_feature"
+      id={`featured-content-${node.id}`}
+      className="featured-story
+      usa-grid
+      usa-grid-full
+      vads-u-margin-bottom--3
+      medium-screen:vads-u-margin-bottom--4
+      vads-u-display--flex
+      vads-u-flex-direction--column
+      medium-screen:vads-u-flex-direction--row
+      vads-u-border-left--7px
+      vads-u-border-color--primary-alt-lightest"
+    >
+      <div className="usa-width-one-half medium-screen:vads-u-padding-left--2">
+        <span className="vads-u-font-weight--bold vads-u-display--block">
+          In the spotlight at
+        </span>
+        <h2 className="vads-u-font-size--md medium-screen:vads-u-font-size--lg vads-u-margin-top--0 medium-screen:vads-u-margin-bottom--0p5">
+          {node.title}
+        </h2>
+        <div className="vads-l-grid-container--full">
+          <div className="va-introtext">
+            <p className="events-show" id="office-events-description">
+              {node.field_intro_text}
+            </p>
+          </div>
+        </div>
+        <div className="usa-width-one-half vads-u-order--first medium-screen:vads-u-order--initial vads-u-margin-bottom--2 medium-screen:vads-u-margin-bottom--0">
+          <MediaImageComponent
+            image={node.field_media}
+            imageStyle={'1_1_square_medium_thumbnail'}
+          />
+        </div>
+      </div>
+    </article>
+  )
+}
+
 /** General News Story component. Allows choice of different display components by the caller. */
 export const NewsStory = ({ node, viewMode, ...props }: NodeNewsStoryProps) => {
   switch (viewMode) {
     case 'full':
       return <NewsStoryFull node={node} {...props} />
       break
-
+    case 'teaser':
+      return <NewsStoryTeaser node={node} {...props} />
+      break
     default:
       return null
   }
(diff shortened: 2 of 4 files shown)


src/pages/demo/pagination/exampleA/[page].tsx
```diff
@@ -0,0 +1,92 @@
+import { useRouter } from 'next/router'
+import { drupalClient } from '@/utils/drupalClient'
+import { GetStaticPathsResult } from 'next'
+import { DrupalJsonApiParams } from 'drupal-jsonapi-params'
+import Container from '@/components/container'
+import { NewsStory } from '@/components/node/news_story'
+import Pagination from '@department-of-veterans-affairs/component-library/Pagination'
```
Comment (amponce) [inline src/pages/demo/pagination/exampleA/[page].tsx:7]: Imported from component library  https://design.va.gov/components/pagination

src/pages/demo/pagination/exampleB/[page].tsx
```diff
@@ -0,0 +1,101 @@
...
+            current={page.current}
+            total={page.total}
+            href={(pager) =>
+              pager === 0
+                ? `/demo/pagination/exampleB/0`
+                : `/demo/pagination/exampleB/${pager}`
+            }
+            className="va-pagination-b"
+          />
+        ) : null}
```
Comment (amponce) [inline src/pages/demo/pagination/exampleB/[page].tsx:36]: using the usePagination hook

src/styles/globals.css
```diff
@@ -39,3 +39,40 @@ a {
...
+  }
+}
+.va-pagination-prev-b,
+.va-pagination-next-b {
+  padding: 0 1rem;
+}
+.va-pagination-prev-b,
+.va-pagination-next-b {
+  position: relative;
+}
```
amponce, line 78: I had to namespace everything with -b because of a style conflict with the component library

ndouglas (approved):
This looks good!  

I was unfamiliar with Next-Drupal.  That slowed me down a lot because I had to try and figure out where the magic was happening, etc.  After reading through a couple times and re-awakening dormant memories, lol, it looks like a great tool, and that and `drupal-jsonapi-params` should make developing this a lot easier.

I know we discussed some of this yesterday, but I'm not good at conversation.  **TL;DR**: I have an auditory processing disorder and basically need subtitles for real life.  Not good at translating incoming sounds into words that mean things.  So maybe this was all covered, and you're reading this like pls stop nug doug pls step away from the keyboard, in which case I'm sorry 😞  

I'm curious about why the stories shown are from Pittsburgh.  I suspect that they're from Pittsburgh because Pittsburgh was the first area added.

If you visit [here](https://prod.cms.va.gov/jsonapi/node/news_story), you get the raw JSON:API response for `news_story` with the default sort, which is AFAICT after pretty-printing and lazily scrolling through all Pittsburgh stuff.

So the problem would then become how to filter based on a specified listing page.  The obstacle I see with that is you're retrieving the information about the listing page from the `field_listing` relationship on the individual stories.  There's some bit of essential information that's missing from the setup that has to be provided in order to specify the listing page and, from there, the stories that will appear.

**Disclaimer**: our situation was different.  `content-build` has some information about the backend baked in, like menu titles and stuff.  I think this is a poor decision, as evidenced by the fact that we have to run a full content build to ensure that someone hasn't put an underscore in place of a dash or made some other typo that breaks the build.  Rather, the frontend had essentially no knowledge about listing pages or anything like that, and relied on links and queries to build that knowledge on-the-fly.  

So we would be provided with the request path, like `/nug-doug-health-care/`, and need to efficiently translate that into an adequate representation of that listing page and its contents.  Then we run into the problem where that path contains no meaningful information, JSON:API-wise.  You can't request `/jsonapi/node/news_listing/nug-doug-health-care` and get a response.

(More about this general problem [here](https://www.lullabot.com/articles/decoupled-hard-problems-routing))

How we handled this was to use subrequests like the following:

```json
[
  {
    "requestId": "translate_path_alias_into_listing_page",
    "uri": `/router/translate-path?path=${path}&_format=json`,
    "action": "view"
  },
  {
    "requestId": "get_listing_page_information",
    "uri": "/jsonapi/node/news_story_listing/{{ translate_path_alias_into_listing_page.body@$.jsonapi.individual }}?fields=...&include=...&etc",
    "action": "view",
    "waitFor": [
      "translate_path_alias_into_listing_page"
    ]
  },
  {
    "requestId": "get_listing_page_nodes",
    "uri": "/jsonapi/node/news_story/?filter[listing_page][condition][path]=field_listing.id&filter[listing_page][condition][value]={{ translate_path_alias_into_listing_page.body@$.jsonapi.individual }}&fields=...&include=...&etc",
    "action": "view",
    "waitFor": [
      "translate_path_alias_into_listing_page"
    ]
  }
]
```

So the first subrequest receives some information about the listing page based on the path provided to the frontend function.  When that subrequest is resolved, the ID of the individual resource is then passed to two subrequests that will be executed concurrently: one for the listing page information, and another for the nodes related to that listing page.

Crucially, these three subrequests occur roughly in the time of two subrequests, and both of those are executed without the overhead of individual Drupal requests; the HTTP overhead, the transit latency, Drupal's lengthy bootstrapping, etc.  So this is very fast relative to two full serialized requests.  

(It's reasonably cacheable, too; we had a Varnish server in front of the frontend server and between the frontend server and the Drupal server and requests were often satisfied in e.g. 2-3 ms.  Requests to the Drupal Subrequests module are normally POST requests, and Varnish demands an actual paid service for caching POST requests, but we just used GET requests and passed the would-be body as query string parameters.  Obviously there's length limits there, but we never encountered any.  IDK if any of this is relevant to our specific use case w.r.t. static content builds, but it absolutely would if we wanna do a dynamic frontend.)

This gets into where I'm confused about Next-Drupal; I don't see where it's possible to use subrequests to make these chains of queries more efficient.  I searched and looked at the documentation and didn't see anything helpful.  I looked in the issue queue on GH and none of the returned issues seemed relevant.  😕 

But I see this as a potential problem the project will face, and Drupal and the bolted-on nature of JSON:API makes it a bigger problem than it should be (IMHO `subrequests` and `decoupled_router` should be in Core or at least JSON:API).  IDK.

penny-lischer (commented):
@amponce were there any benefits of one over the other?
Considering that we will be leveraging other components from the component library I am leaning toward that approach unless there is a benefit to use the Next-Drupal Pagination.

amponce:
@ndouglas I think we've figured out how to render the story listings based on path alias, I have a working prototype here: https://github.com/department-of-veterans-affairs/next-build/pull/36 Thank you for the detailed recap of the previous issue. We're currently using decoupled router and next-drupal offers several different data fetching methods via resourceCollectionContext that helps resolve the endpoints. I'm still just scratching the surface of this problem so there's still a lot to figure out but we did make some good progress today.

PR #36, opened 2022-06-01
Story listing collection

amponce:
Example render collection of story listings on path alias

Screen Shot 2022-05-31 at 6 28 04 PM


src/components/node/story_listing/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+/** Export information necessary to identify the component and query it.
+ * See {@link NodeMetaInfo}
+ */
+export const Meta: NodeMetaInfo = {
+  resource: 'node--story_listing',
+  component: StoryListing,
+  params: params,
+  collection: true,
+  additional: 'node--news_story',
```
penny-lischer, line 85: Can we call this 'additionalResource' or 'additionalNode' just to be more explicit

penny-lischer (approved):
LGTM

PR #37, opened 2022-06-01
VACMS-9118 Social links component

penny-lischer:


penny-lischer:
Screen Shot 2022-06-01 at 2 14 08 PM

src/components/partials/socialLinks/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+          <a
+            className="va-js-share-link"
+            href={`https://twitter.com/intent/tweet?text=${node?.title}&url=${hostUrl}`}
+          >
+            <i className="va-c-social-icon fab fa-twitter vads-u-margin-right--0p5"></i>
+            Share on Twitter
+          </a>
+        </li>
+      </ul>
+    </div>
```
amponce, line 29: It looks good to me! Based on the code it looks like you built the facilities variant. I noticed there is a similar but slightly different version in the includes directory with some additional calendar logic. I'm not sure where it's being used but I think we might want to make a note of it and consolidate them into a single component in the future. nice work! 🚀 

```
<a
data-description="{{ fieldDescription }}"
data-end="{{ fieldDatetimeRangeTimezone.endValue }}"
data-location="{{ fieldAddress.addressLine1 }} {{ fieldAddress.locality }}, {{ fieldAddress.administrativeArea }}"
data-subject="{{ title }}"
href="{{ entityUrl.path }}"
id="add-to-calendar-link" data-start="{{fieldDatetimeRangeTimezone.value }}"
>
<i class="va-c-social-icon fas fa-calendar-check vads-u-margin-right--0p5" aria-hidden="true" role="presentation" aria-hidden="true"></i>
Add to Calendar
</a>
```

1. [/facilities/story_social_share.drupal.liquid ](https://github.com/department-of-veterans-affairs/content-build/blob/main/src/site/facilities/story_social_share.drupal.liquid )
2. [/includes/social-share.drupal.liquid](https://github.com/department-of-veterans-affairs/content-build/blob/main/src/site/includes/social-share.drupal.liquid)

amponce (approved):

src/components/partials/socialLinks/index.tsx
penny-lischer: Yes, and also here - campaign_landing_page.drupal.liquid
Per the story I based the component off of the facilities variant and it looks like it will be able to use in the other two areas.

PR #38, opened 2022-06-02
VACMS-000 news listing examples for testing locations

amponce:
Article list with filtering to test news stories from various locations

Screen Shot 2022-06-01 at 6 06 53 PM


src/components/node/story_listing/article_list.tsx
```diff
@@ -0,0 +1,88 @@
+import React, { useEffect, useState } from 'react'
+import Link from 'next/link'
+import config from '../../../../config'
+
+function isValidData(data) {
+  if (data === null) {
```
penny-lischer, line 6: These if conditions can be combined

src/pages/[...slug].tsx
```diff
@@ -56,28 +56,41 @@ export async function getStaticProps(
...
       additionalNode =
         await drupalClient.getResourceCollectionFromContext<NodeTypes>(
           addResourceToCollection,
           context,
-          { params: nodeMeta[type].additionalParams.getQueryObject() }
+          {
+            params: {
+              'filter[field_listing.drupal_internal__nid][value]' : path.entity.id, // Todo make the filter option dynamic
+              ...nodeMeta[type].additionalParams.getQueryObject(),
+            },
```
Comment (amponce) [inline src/pages/[...slug].tsx:89]: Adding the filter here based on `path.entity.id` fixes the issue, however this needs to be made more dynamic

src/pages/[...slug].tsx
```diff
@@ -56,28 +56,41 @@ export async function getStaticProps(
...
+
   if (isCollection) {
     node = await drupalClient.getResourceCollectionFromContext<NodeTypes>(
       type,
       context,
-      { params: nodeMeta[type].params.getQueryObject() }
+      {
+        params: {  
+          'filter[drupal_internal__nid][value]': path.entity.id,
+          ...nodeMeta[type].params.getQueryObject(),
```
Comment (amponce) [inline src/pages/[...slug].tsx:74]: filtering by `path.entity.id` fixes the issue but is there a better way to do this?

src/components/node/story_listing/article_list.tsx
amponce: Created a helper file

amponce: Created a helper file

penny-lischer (approved):
LGTM

PR #39, opened 2022-06-06
VACMS-9129 abstract Paragraph component

penny-lischer:
- created a Paragraph component to delegate which component to render based on the Paragraph data.
- Fixed lint errors on the news_story tests

### DIFF
--- a/src/components/paragraph/index.tsx
+++ b/src/components/paragraph/index.tsx
@@ -11,8 +11,7 @@ import { Meta as LinkTeaserMeta } from '@/components/paragraph/link_teaser'
 /** General ParagraphProps to pass paragraphs into paragraph components. */
 export interface ParagraphProps {
   paragraph: ParagraphTypes
-  boldTitle?: boolean
-  sectionHeader?: string
+  componentParams?
 }
 
 /** Each Paragraph component must export a ParagraphMetaInfo object `Meta`. This information helps next-build associate Drupal resource types with information for rendering them.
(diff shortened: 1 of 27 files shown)


src/components/paragraph/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+export function Paragraph({ paragraph, ...props }: ParagraphProps) {
+  if (!paragraph) return null
+
+  const Component = paragraphMeta[paragraph.type].component
+
+  if (!Component) return null
+
+  return <Component paragraph={paragraph} {...props} />
+}
```
amponce, line 82: This looks great! is it working as expected?

src/types/paragraph.d.ts
```diff
@@ -10,9 +10,20 @@ import {
...
+  NodeOffice,
+  NodePersonProfile,
+  NodePromoBanner,
   NodeQA,
+  NodeRegionalHealthCareServiceDes,
+  NodeStepByStep,
+  NodeStoryListing,
+  NodeSupportResourcesDetailPage,
   NodeSupportService,
 } from './node'
```
amponce, line 28: Why do we have `node` and `media` referenced in paragraph.d.ts? I haven't looked at this very closely but it seems a bit strange

src/types/paragraph.d.ts
```diff
@@ -143,6 +181,19 @@ export interface ParagraphServiceLocationAddress extends DrupalParagraph {
...
+  field_first_name: fieldFirstName
+  field_last_name: fieldLastName
+  field_email_address: fieldEmailAddress
+  field_phone_number: fieldPhoneNumber
+  field_suffix: fieldSuffix
+  field_description: fieldDescription
+  field_complete_biography_create: fieldCompleteBiographyCreate
+  field_media: fieldMedia
+  field_entity: fieldEntity
+}
```
amponce, line 195: Doesn't need to be this ticket but at some point we'll need descriptions so type doc picks everything up. Maybe we should create a task?

src/types/paragraph.d.ts
penny-lischer: Will fix. This happened when merging with main.

PR #46, opened 2022-06-23
VACMS-9185: Add tests for ImageComponent

ndouglas:
Focused on testing the Image side of things, since the Media side is actually already covered [here](https://github.com/department-of-veterans-affairs/next-build/blob/main/src/components/media/index.test.tsx).


penny-lischer (approved):
LGTM

PR #47, opened 2022-06-28
VACMS-8654 node--banner and node--promo_banner

penny-lischer:
- Includes refactoring of paragraph and node typescript files.
- Includes importing styles via defineCustomElements within _app


penny-lischer:
Screen Shot 2022-06-28 at 11 38 07 AM

PR #52, opened 2022-07-08
VACMS-9190 Layout Unit Tests

penny-lischer:


amponce (approved):
🎉 LGTM, that layout issue was a pain

PR #54, opened 2022-07-08
VACMS-9187: Shift news_story mock data into updatable fixture, etc.

ndouglas:
Closes department-of-veterans-affairs/va.gov-cms#9187.

Still gotta document this, clean this up a little, and do a little presentation, but I think this is a reasonable approach.  TL;DR: if data changes we just delete the file and a new fixture will get created.


src/components/node/news_story/index.test.tsx
```diff
@@ -2,14 +2,36 @@ import { screen, render } from '@testing-library/react'
...
+}
+
+afterAll(async () => {
+  nock.restore()
+})
 
 describe('<NewsStory> component renders', () => {
-  test('with valid data', () => {
-    render(<NewsStory node={data[0] as never} viewMode={'full'} />)
+  test('with valid data', async () => {
```
penny-lischer, line 27: @ndouglas should the other news story tests be updated as well?

penny-lischer (approved):
LGTM

PR #55, opened 2022-07-08
VACMS 8639 Component - Paragraph Rich Text Char Limit 1000

amponce:
### Component - Paragraph Rich Text Char Limit 1000
--------------------

- [x] A component exists that can render a `paragraph--rich_text_char_limit_1000` data structure
- [x] A page exists within the next-build repo that demonstrates the rendering of all `paragraph--rich_text_char_limit_1000`
- [x] That component's output is consistent with the existing output of content-build (visually, markup structure

Screen Shot 2022-07-08 at 1 14 05 PM
)


amponce:
So it looks like rich text limit is controlled by the CMS text area input so I didn't write a test the number of characters rendered.

PR #56, opened 2022-07-13
VACMS banner global context

amponce:
Set up global context for banners and footer
Pass context into layout
Determine if the user is on a slug route or the homepage

### DIFF
--- a/src/components/node/banner/index.tsx
+++ b/src/components/node/banner/index.tsx
@@ -1,7 +1,31 @@
+import { useRef, useEffect, useState } from 'react'
 import { VaBanner } from '@department-of-veterans-affairs/component-library/dist/react-bindings'
 import { isEmpty } from 'lodash'
+import { recordEvent } from '@/utils/recordEvent'
+import { NodeMetaInfo, NodeResourceType } from '@/types/node'
 
 const BannerComponent = ({ node }): JSX.Element => {
+  const [isClicked, setIsClicked] = useState(false)
+  const analyticsRef = useRef(null)
+
+  useEffect(() => {
+    function handler(event) {
+      if (analyticsRef.current?.contains(event.target)) {
+        setIsClicked(true)
+      }
+    }
+    window.addEventListener('click', handler)
+    return () => window.removeEventListener('click', handler)
+  }, [])
+
+  if (isClicked) {
+    recordEvent({
+      eventCategory: 'Banner',
+      eventAction: 'Click',
+      eventLabel: node.title,
+    })
+  }
+
   if (isEmpty(node)) return
 
   return (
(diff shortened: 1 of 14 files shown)


src/components/node/banner/index.tsx
```diff
@@ -1,7 +1,31 @@
+import { useRef, useEffect, useState } from 'react'
 import { VaBanner } from '@department-of-veterans-affairs/component-library/dist/react-bindings'
 import { isEmpty } from 'lodash'
+import { recordEvent } from '@/utils/recordEvent'
+import { NodeMetaInfo, NodeResourceType } from '@/types/node'
 
 const BannerComponent = ({ node }): JSX.Element => {
+  const [isClicked, setIsClicked] = useState(false)
```
penny-lischer, line 8: This logic doesn't need to be in the banner component

amponce: Whoops that was in there from earlier, removed

src/pages/[...slug].tsx
```diff
@@ -7,17 +7,24 @@ import {
...
 interface NodeProps {
   node: NodeTypes
   additionalNode: NodeTypes
 }
 
 /** This passes any retrieved node to the generalized Node component. */
-export default function NodePage({ node, additionalNode }: NodeTypes) {
+export default function NodePage({ node, additionalNode, props }: NodeTypes) {
   if (!node) return null
+  if (!props) return null
```
Comment (timcosgrove) [inline src/pages/[...slug].tsx:21]: Why the change from `node` to `props` here? Is the `node` check no longer relevant?

Comment (amponce) [reply src/pages/[...slug].tsx:21]: Props seems to be the convention used to access the data from globalElements, I can probably destructure props within the globalElements file but we would probably want to call it something like banners since we already have 'node' serving a different purpose in this file. I agree it's a little bit confusing

src/lib/context/getGlobalElements.ts
```diff
@@ -0,0 +1,38 @@
+import { GetServerSidePropsContext, GetStaticPropsContext } from 'next'
+import { drupalClient } from 'utils/drupalClient'
+import { LayoutProps } from 'components/layout'
+
+type GlobalElements = LayoutProps
+
+// This is a helper function to fetch global elements for layout.
+// This is going to be run for every pages on build.
+// To make this fast, you could cache the results example on Redis.
+export async function getGlobalElements(
```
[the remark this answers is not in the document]
amponce: Yeah I should add some tests today. I ended up running into a few environment issues on Friday so I left a few things undone

src/pages/[...slug].tsx
Comment (timcosgrove) [reply src/pages/[...slug].tsx:21]: That makes sense re: `props`, but is it no longer relevant if `node` is empty?

Comment (timcosgrove) [reply src/pages/[...slug].tsx:21]: I can't read.

PR #57, opened 2022-07-15
VACMS-9753 Component: node--full_width_banner_alert

penny-lischer:


src/components/node/banner_alert/index.tsx
```diff
@@ -77,6 +55,29 @@ const BannerAlert = ({ node }): JSX.Element => {
...
+        'alert-box-headline-level': '3',
+        'alert-box-background-only': 'false',
+        'alert-box-closeable': 'false',
+        'alert-box-click-label': '$2',
+      }
+    }
+    recordEvent(eventData)
+    setIsClicked(false)
+  }
+
```
amponce, line 80: Clean!

amponce (approved):
Looks great! I think we just need to run the yarn format to fix the prettier eslint issues and this should be good to go

src/components/node/banner_alert/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+      showClose={node.field_alert_dismissable}
+      headline={node.title}
+      type={node.field_alert_type}
+      visible={outputStatus}
+      windowSession={node.field_alert_dismissable == 'dismiss-session'}
+    >
+      <div ref={analyticsRef} dangerouslySetInnerHTML={{ __html: body }} />
+    </VaBanner>
+  )
+}
```
timcosgrove, line 94: We need the `meta` object exported somewhere here. Even though we're not using banner_alert in a more conventional page context, we rely on the node meta for other purposes, including letting `<Node />` know what component will render objects of type `node--full_width_banner_alert`.

It can be modeled after News Story: https://github.com/department-of-veterans-affairs/next-build/blob/main/src/components/node/news_story/index.tsx#L148

We won't need `params` for this; that should be able to be null.

src/types/taxonomy_term.d.ts
```diff
@@ -28,4 +28,4 @@ export interface TaxonomyTermLcCategories extends DrupalTaxonomyTerm {}
 
 // Allow this for consistency, even though no fields are present.
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface TaxonomyTermTopics extends DrupalTaxonomyTerm {}
+export interface TaxonomyTermTopics extends DrupalTaxonomyTerm {}
```
timcosgrove, line 31: This is causing a linting failure.

src/utils/helpers.ts
```diff
@@ -79,4 +85,4 @@ export const drupalToVaPath = (content) => {
   }
 
   return replaced
-}
+}
```
timcosgrove, line 88: Same, linting failure.

src/utils/helpers.ts
```diff
@@ -1,5 +1,6 @@
 import config from '../../config'
 import isNil from 'lodash/isNil'
+import { recordEvent } from '@/utils/recordEvent'
```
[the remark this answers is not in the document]
penny-lischer: Will remove

src/components/node/banner_alert/index.tsx
penny-lischer: Will do

PR #59, opened 2022-07-20
VACMS-8688 Component: block_content--alert

penny-lischer:


amponce (approved):
This turned out great Penny I just ran though it locally and didn't see anything that needed to change.

amponce:
I noticed that the links don't have google analytics events firing like we have on the banners, is there anything that needs to be added for tracking banner interactions?

penny-lischer:
@amponce there was no logic in the current implementation that recorded GA events so I don't think its necessary

PR #60, opened 2022-07-21
VACMS-9138 set up storybook

tjheffner:
- adds storybook to the repo with typescript support. updated the eslint settings to extend the storybook plugin and suggested overrides.
- storybook can be run in local development (`yarn storybook`) or compiled as a static site for QA/higher environment purposes (`yarn build-storybook` will output to the `/storybook-static/` directory). the compiled directory has been added to gitignore/prettierignore.
- excludes `*.stories.tsx` from the test coverage check... the stories rendering properly is enough of a test for this imo. the component code should still have tests written normally.
- `*.stories.tsx` files can live next to component files. i kept the `/stories/` directory to hold any potential markdown stories we may want (introduction, other docs, etc).
- adds or stubs example stories for the three "main" sections i found: `components`, `node`, `paragraphs`
  - I created initial stories for image and media components, button paragraph, news story node
  
  
  QA:
  pull this branch, run `yarn storybook` and visit http://localhost:6006/
  note the data used in the media and new_story stories are not fully correct at this time, but data does render


src/utils/recordEvent.js
```diff
@@ -19,6 +19,7 @@ export const recordEvent = (data) => {
  * @param {string} key the property in the event object to use when looking for
  *                     existing matches in the dataLayer
  */
+/* eslint-disable-next-line typescript-eslint/no-shadow */
```
tjheffner, line 22: storybook was having trouble compiling the news_story component without this piece... i think some rule under the hood from their eslint extension and the fact this is a .js module being imported. it doesn't like the `(event,` because we import `event` at the top of the file.

amponce: yeah, that's the analytics beacon causes a bunch of issues in tests also

amponce (approved):
Looks fantastic!

.storybook/preview-head.html
```diff
@@ -0,0 +1,13 @@
...
+  data-entry-name="style.css"
+  href="https://unpkg.com/@department-of-veterans-affairs/formation/dist/formation.min.css"
+/>
+
+<!-- CORS appears to block the fonts from loading at this link -->
+<link
+  rel="stylesheet"
+  data-entry-name="style.css"
+  href="https://prod-va-gov-assets.s3-us-gov-west-1.amazonaws.com/generated/style.css"
+/>
```
amponce, line 13: is this how we pick up styles in storybook?

tjheffner: so there's a few different ways styles can be added to storybook.

- styles can be included in the component files themselves (either imported or css-in-js) [not really doing this in hte repo that i could see]. this is the most common approach when styles live in the repo as well.
- they can be included for all stories inside of `.storybook/preview.js` [this is where i import the local global styles we do have in the repo itself]
- this `preview-head.html` serves as an extension of sorts to the preview.js file mentioned above, it modifies the head inside of the <iframe> for all component stories. because we are using these external stylesheets, i added them here.

PR #62, opened 2022-07-22
build fix

amponce:
Fixes build 

```
Page                                                      Size     First Load JS
┌ ● / (1184 ms)                                           4.2 kB          186 kB
├   /_app                                                 0 B            94.6 kB
├ ● /[...slug] (3104014 ms)                               3.77 kB         185 kB
├   ├ /central-arkansas-health-care/stories (3134 ms)
├   ├ /southeast-louisiana-health-care/stories (3079 ms)
├   ├ /salt-lake-city-health-care/stories (3056 ms)
├   ├ /salem-health-care/stories (3013 ms)
├   ├ /alexandria-health-care/stories (3004 ms)
├   ├ /philadelphia-health-care/stories (2989 ms)
├   ├ /western-colorado-health-care/stories (2984 ms)
├   └ [+1510 more paths] (avg 2042 ms)
├ ○ /404                                                  181 B          94.8 kB
├ λ /api/exit-preview                                     0 B            94.6 kB
├ λ /api/preview                                          0 B            94.6 kB
├ ● /demo/audienceTopics (791 ms)                         3.79 kB         185 kB
├ ● /demo/banner (1512 ms)                                4.51 kB         125 kB
├ ● /demo/bannerAlert (796 ms)                            5.62 kB         137 kB
├ ● /demo/blockAlert (731 ms)                             4.63 kB         125 kB
├ ● /demo/button (784 ms)                                 382 B           174 kB
├ ● /demo/emailContacts (831 ms)                          388 B           174 kB
├ ● /demo/expandableText (855 ms)                         394 B           174 kB
├ ● /demo/linkTeasers (769 ms)                            447 B           174 kB
├ ● /demo/media (970 ms)                                  3.82 kB         185 kB
├ ● /demo/pagination/exampleA/[page] (13313 ms)           3.86 kB         118 kB
├   ├ /demo/pagination/exampleA/9 (716 ms)
├   ├ /demo/pagination/exampleA/4 (705 ms)
├   ├ /demo/pagination/exampleA/11 (705 ms)
├   ├ /demo/pagination/exampleA/7 (704 ms)
├   ├ /demo/pagination/exampleA/6 (696 ms)
├   ├ /demo/pagination/exampleA/0 (689 ms)
├   ├ /demo/pagination/exampleA/3 (689 ms)
├   └ [+13 more paths] (avg 647 ms)
├ ● /demo/pagination/exampleB/[page] (12284 ms)           4.63 kB         119 kB
[...]

PR #63, opened 2022-07-25
VACMS-0000 wysiwyg as string type

tjheffner:
After pulling main I was seeing this error when running storybook in the other branch... the demo still appeared to render correctly, holler if this is breaking things / the wrong way to go about it

```
ERROR in src/components/paragraph/wysiwyg/index.tsx:25:7
TS2322: Type '{ __html: string | any[]; }' is not assignable to type '{ __html: string; }'.
  Types of property '__html' are incompatible.
    Type 'string | any[]' is not assignable to type 'string'.
      Type 'any[]' is not assignable to type 'string'.
    23 |       key={paragraph.id}
    24 |       className={className}
  > 25 |       dangerouslySetInnerHTML={createMarkup()}
       |       ^^^^^^^^^^^^^^^^^^^^^^^
    26 |     />
    27 |   )
    28 | }
```

### DIFF
--- a/src/components/paragraph/wysiwyg/index.tsx
+++ b/src/components/paragraph/wysiwyg/index.tsx
@@ -14,7 +14,7 @@ function Wysiwyg({ paragraph, className }: ParagraphProps) {
     const filteredData = filters.reduce((d, f) => d.filter(f), data)
 
     return {
-      __html: filteredData || [filteredData[0]] || '',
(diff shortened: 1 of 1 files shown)


tjheffner:
actually this probably isn't needed with the updates inside of #62

penny-lischer:
@tjheffner this will be resolved with build fix #62

PR #70, opened 2022-08-02
VACMS-10005 Refactor: button component

penny-lischer:


src/components/button/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+import validator from 'validator'
+import { isEmpty, isNull } from 'lodash'
+
+export type ButtonProps = {
+  id: string
+  label: string
+  url: string
+}
+
+function isRequestValid(button) {
```
timcosgrove, line 11: What about naming this `isRequestInvalid`, since we are returning out of the component if the function returns `true`? 

Or, keep the name and invert some of the logic:
```
!isEmpty(button.label) && !isEmpty(button.url) && validator.isURL(button.url)
```
maybe?

timcosgrove (commented):
I think it might make sense to modify the button valid function a bit; see inline.

tjheffner (approved):

amponce (commented):
Would `buttonsCollectionProps: ParagraphButton[]` work?

amponce (approved):
👍

PR #71, opened 2022-08-02
DNM: Prototype for next-drupal-query usage.

timcosgrove:
A pass at what reorganization might look like.

### DIFF
--- a/src/components/media/index.tsx
+++ b/src/components/media/index.tsx
@@ -1,14 +1,6 @@
+import { ImageProps } from '@/types/index'
 import Image from '@/components/image'
 
-// Different from the ImageProps that are imported in image/index.tsx
-export type ImageProps = {
-  url: string
-  width?: number
-  height?: number
-  alt?: string
-  title?: string
-  styles?: object //todo: do we need to type this more strongly?
-}
 
 export type MediaImageProps = {
   image: ImageProps
--- a/src/components/news_story/dataService.tsx
+++ b/src/components/news_story/dataService.tsx
@@ -5,16 +5,15 @@ import { mediaImageDataService } from '@/components/media/dataService'
 import { EntityMetaInfo } from '@/lib/delegators/entityMetaProvider'
 import { NodeNewsStory, NodeResourceType } from '@/types/node'
 import {
-  NewsStoryFull,
-  NewsStoryProps,
-  NewsStoryTeaserProps,
+  NewsStoryFull
 } from '@/components/news_story'
+import { NewsStoryPageProps, NewsStoryTeaserProps } from '@/types/index'
 
 export const newsStoryDataService = function (
   entity: NodeNewsStory,
   viewMode: string,
   headingLevel?: ComponentType | keyof JSX.IntrinsicElements
-): NewsStoryProps | NewsStoryTeaserProps {
+): NewsStoryPageProps | NewsStoryTeaserProps {
   switch (viewMode) {
     case 'teaser':
       return {
--- a/src/components/news_story/index.tsx
+++ b/src/components/news_story/index.tsx
@@ -1,32 +1,12 @@
 import { ComponentType } from 'react'
-import { ImageProps, MediaImageComponent } from '@/components/media'
-import {
-  PersonProfileTeaserProps,
-  StaffNewsProfile,
-} from '@/components/person_profile'
 import { formatDate, truncateWordsOrChar } from '@/utils/helpers'
-import { SocialLinks, SocialLinksProps } from '@/components/socialLinks'
+/* Types these components implement. */
+import { NewsStoryPageProps, NewsStoryTeaserProps } from '@/types/index'
+/* Components used within these components. */
+import { MediaImageComponent } from '@/components/media'
+import { StaffNewsProfile } from '@/components/person_profile'
+import { SocialLinks } from '@/components/socialLinks'
 import { StoryListingTeaser } from '../story_listing'
-
-export type NewsStoryProps = {
-  title: string
-  image: ImageProps
-  caption: string
-  author: PersonProfileTeaserProps | any
-  introText: string
-  bodyContent: string
-  date: string
-  socialLinks: SocialLinksProps
-  listing: string
-}
-
-export type NewsStoryTeaserProps = {
-  headingLevel?: ComponentType | keyof JSX.IntrinsicElements
-  link: string
-  title: string
-  image: ImageProps
-  introText: string
-}
 
 export const NewsStoryFull = ({
   title,
@@ -38,7 +18,7 @@ export const NewsStoryFull = ({
   date,
   socialLinks,
   listing,
-}: NewsStoryProps) => {
+}: NewsStoryPageProps) => {
   return (
     <>
       <div id="content" className="interior">
--- a/src/data/mappings/NewsStoryPageMapping.ts
+++ b/src/data/mappings/NewsStoryPageMapping.ts
@@ -1,0 +1,26 @@
+import { NewsStoryPageProps } from "@/types/index"
+import { NodeNewsStory } from "data/types/drupal"
+import { mediaImageDataService } from "@/components/media/dataService"
+import { generalEntityDataService } from "@/lib/delegators/generalEntityDataService"
+
+const NewsStoryPageMapping = function(node: NodeNewsStory): NewsStoryPageProps {
+      return {
+        type: node.type,
+        published: node.status,
+        title: node.title,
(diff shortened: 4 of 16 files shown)


src/components/media/index.tsx
```diff
@@ -1,14 +1,6 @@
+import { ImageProps } from '@/types/index'
 import Image from '@/components/image'
```
timcosgrove, line 3: Everything that we called 'display props' or 'component props' moves down to `@/types/`. These types serve two purposes:
* They define what the component should implement.
* They define what the drupal->next-build mapping should output.

src/components/news_story/dataService.tsx
```diff
@@ -5,16 +5,15 @@ import { mediaImageDataService } from '@/components/media/dataService'
 import { EntityMetaInfo } from '@/lib/delegators/entityMetaProvider'
```
timcosgrove, line 5: I did not remove this dataService file, but you can think of the `@/data/mappings` and `@/data/queries` files are serving the same function.

src/components/news_story/index.tsx
```diff
@@ -1,33 +1,13 @@
 import { ComponentType } from 'react'
-import { ImageProps, MediaImageComponent } from '@/components/media'
-import {
-  PersonProfileTeaserProps,
-  StaffNewsProfile,
-} from '@/components/person_profile'
 import { formatDate, truncateWordsOrChar } from '@/utils/helpers'
-import { SocialLinks, SocialLinksProps } from '@/components/socialLinks'
```
timcosgrove, line 8: Reorganized these. More global/helpers first; the Prop types second; any components imported for use third.

src/components/news_story/index.tsx
```diff
@@ -1,33 +1,13 @@
...
+/* Types these components implement. */
+import { NewsStoryPageProps, NewsStoryTeaserProps } from '@/types/index'
+/* Components used within these components. */
+import { MediaImageComponent } from '@/components/media'
+import { StaffNewsProfile } from '@/components/person_profile'
+import { SocialLinks } from '@/components/socialLinks'
 import { StoryListingTeaser } from '../story_listing'
```
timcosgrove, line 10: As stated earlier, these moved to be global types.

src/data/mappings/NewsStoryPageMapping.ts
```diff
@@ -0,0 +1,26 @@
+import { NewsStoryPageProps } from "@/types/index"
```
timcosgrove, line 1: This file is responsible for doing mappings from our data source to our 'global' output structure. Note that the mapping function explicitly expects a Drupal data structure as input, and it explicitly outputs the structure `NewsStoryPageProps`, which is also what the component is required to implement.

src/data/queries/NewsStoryPage.ts
```diff
@@ -0,0 +1,34 @@
+import { QueryData, QueryOpts, QueryParams } from "@next-drupal/query"
```
timcosgrove, line 1: This is based on the example at https://shadcn.notion.site/next-drupal-query-bdbe85efca244b33b7b253ef87938826#:~:text=Bash-,Create,-your%20first%20query. A couple differences:
* we are defining `NewsStoryPageProps` elsewhere, since it's essentially a contract for what `NewsStoryPage` data should look like
* We also defined the mapping above.

src/data/mappings/NewsStoryPageMapping.ts
```diff
@@ -0,0 +1,26 @@
...
+import { NodeNewsStory } from "data/types/drupal"
+import { mediaImageDataService } from "@/components/media/dataService"
+import { generalEntityDataService } from "@/lib/delegators/generalEntityDataService"
+
+const NewsStoryPageMapping = function(node: NodeNewsStory): NewsStoryPageProps {
+      return {
+        type: node.type,
+        published: node.status,
+        title: node.title,
+        image: mediaImageDataService(node.field_media),
```
timcosgrove, line 11: These are still old-style. They would need to be replaced by a call directly to a Mapping function.

src/data/types/drupal.d.ts
```diff
@@ -0,0 +1,418 @@
+import { DrupalFile, DrupalMedia, DrupalNode } from "next-drupal"
```
timcosgrove, line 1: This is a types file for Drupal-specific structures. These are intended to verify that the data we receive from Drupal contains the information we expect.

This might want to be broken up into multiple files. It's already hard to navigate, and we're going to have many more types here.

src/pages/[...slug].tsx
```diff
@@ -1,110 +1,75 @@
-import {
```
Comment (timcosgrove) [inline src/pages/[...slug].tsx:1]: This follows the example at https://shadcn.notion.site/next-drupal-query-bdbe85efca244b33b7b253ef87938826, with some modifications for this proposed structure.

src/pages/[...slug].tsx
```diff
@@ -1,110 +1,75 @@
...
-import Layout from '@/components/layout'
-import { generalEntityDataService } from '@/lib/delegators/generalEntityDataService'
+import * as React from "react"
+import { GetStaticPathsResult } from "next"
 
-interface EntityProps {
-  entityProps: any
+import { drupalClient } from "@/utils/drupalClient"
+import { queries } from "@/data/queries"
+import { NewsStoryPageProps } from "../types"
```
Comment (timcosgrove) [inline src/pages/[...slug].tsx:6]: We load this here to type the resource that gets returned from our `getStaticProps` function. In practice this would be a union type.

src/pages/[...slug].tsx
```diff
@@ -1,110 +1,75 @@
...
-export default function Page({ entityProps, props }) {
-  const Component = entityMeta[entityProps.type].component
+export default function ResourcePage({ resource }: ResourcePageProps) {
+  if (!resource) return null
 
   return (
-    <Layout props={props}>
-      <Component {...entityProps} />
+    <Layout>
+      {resource.type === "node--news_story" && <NewsStoryFull {...resource} />}
```
Comment (timcosgrove) [inline src/pages/[...slug].tsx:21]: I have not yet created an equivalent of our `<Node />` component that does delegation, but we will want something similar to that which maps a resource type to a component for page rendering.

src/pages/[...slug].tsx
```diff
@@ -1,110 +1,75 @@
...
-/** @todo This cannot handle non-node urls yet. */
-export async function getStaticProps(
-  context: GetStaticPropsContext
-): Promise<GetStaticPropsResult<EntityProps>> {
-  const params = new DrupalJsonApiParams()
+export async function getStaticProps(context) {
   const path = await drupalClient.translatePathFromContext(context)
 
-  if (!path || !resourceTypes.includes(path.jsonapi.resourceName)) {
+  if (!path) {
```
Comment (timcosgrove) [inline src/pages/[...slug].tsx:39]: I like our version better here.

src/pages/[...slug].tsx.bak
```diff
@@ -0,0 +1,110 @@
+import {
```
Comment (timcosgrove) [inline src/pages/[...slug].tsx.bak:1]: This is just moved aside, ignore it.

src/types/index.d.ts
```diff
@@ -3,3 +3,43 @@ export global {
     gtag: any
   }
 }
+export interface SocialLinksProps {
```
timcosgrove, line 6: (also above): Everything that we called 'display props' or 'component props' moves down to `@/types/`. These types serve two purposes:

* They define what the component should implement.
* They define what the drupal->next-build mapping should output.

It would probably be sensible to break these up into multiple files.

src/data/mappings/NewsStoryPageMapping.ts
tjheffner: that makes sense to me, especially w/r/t paragraphs

src/data/queries/NewsStoryPage.ts
```diff
@@ -0,0 +1,34 @@
...
+
+// Implement the data loader.
+export const data: QueryData<DataOpts, NewsStoryPageProps> = async (
+  opts
+): Promise<NewsStoryPageProps> => {
+  const node = await drupalClient.getResource<NodeNewsStory>("node--news_story", opts?.id, {
+    params: params().getQueryObject(),
+  })
+
+  return NewsStoryPageMapping(node)
```
tjheffner, line 33: this extra layer of abstraction does keep the query file pretty clean... i can dig it

src/pages/[...slug].tsx
Comment (tjheffner) [reply src/pages/[...slug].tsx:39]: would we still want to have additional type checks in any spot? i.e. at the start of the mapping functions?

timcosgrove:
Closing in favor of an upcoming PR.

PR #77, opened 2022-08-03
VACMS-10009 Refactor: expandable_text component

penny-lischer:
## Description
Refactor to allow for DRUPAL API data to get processed/transformed through data_service

Relates to #10009


src/components/expandable_text/dataService.tsx
```diff
@@ -14,12 +14,13 @@ export const expandableTextDataService = function (
   viewMode: string
 ) {
   if (!entity || !isRequestValid(entity)) return
+
   switch (viewMode) {
     default:
       return {
```
amponce, line 20: do we still need viewmode?

amponce (approved):
Lgtm

src/components/expandable_text/dataService.tsx
penny-lischer: Removed

PR #90, opened 2022-08-15
Modifications to help with export.

timcosgrove:
## Description
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/10265


next.config.js
```diff
@@ -9,11 +9,7 @@ const nextConfig = {
       's3-us-gov-west-1.amazonaws.com',
       'va.gov',
     ],
-  },
-  experimental: {
-    images: {
-      unoptimized: true,
-    },
+    loader: 'custom',
```
timcosgrove, line 12: @amponce I'd like your review in particular to see if what I've done to image handling in this PR matches with the original intention for how we're handling images in here. We can't rely on Next/Vercel image optimization for export. Export fails if it's in place, because at export time there's not a Next server running. The way we had set it up previously was triggering that complaint.

src/lib/utils/drupalClient.ts
```diff
@@ -16,8 +17,14 @@ export const fetcher = async (input: RequestInfo, init?: RequestInit) => {
     ...init,
   }
 
-  return crossFetch(input, {
-    ...options,
+  // Wrap fetching in p-retry for resilience.
```
timcosgrove, line 20: Did this because if a single API call failed for whatever reason, the entire export would fail. This just retries a few times.

next.config.js
amponce: I'm going to pull this repo down and take a look at this right now

amponce: So I think we can remove the unoptimized prop from the Image component and the failing test can be updated to

```
 test('throws an error if no URL is passed in', () => {
    // eslint-disable-next-line
    expect(() => render(<Image src={url} alt=''/>)).toThrowError(
      /url is not defined/i
    )
  })
  
  ```

Comment (amponce) [reply next.config.js:22]: I just did a trial run and the export didn't work with the images, I'll see if there's another way to do it

File: src/templates/components/emailContact/index.test.tsx [hunk 3]
```diff
@@ -1,5 +1,6 @@
 import { render, screen } from '@testing-library/react'
 import { fireEvent, getByRole } from '@testing-library/dom'
+jest.mock('@/lib/utils/recordEvent')
```
timcosgrove, line 3: This and below is based on examples from https://github.com/aelbore/esbuild-jest/issues/26

I would appreciate a look to see if I am still in fact testing for the same things.

ndouglas (approved):
Yeah, I think this looks right.

src/templates/components/emailContact/index.test.tsx
amponce: Screen Shot 2022-08-23 at 10 41 53 PM

Looks like the output matches the expected links-list-header:  `'t%24st.vet%3Dran%40va.gov',`
```
  event: 'nav-linkslist',
      'links-list-header': 't%24st.vet%3Dran%40va.gov',
      'links-list-section-header': 'Need more help?',
    })
```

amponce (commented):
It's looking good, I don't know what else remains on this one. Did you get .swc and jest working or is that one going to take some time?

amponce (approved):
👍 The changes look good unless we get blocked with swc issues

next.config.js
```diff
@@ -2,18 +2,24 @@
  * @type {import('next').NextConfig}
  */
 const nextConfig = {
+  experimental: {
+    swcPlugins: [
+      [
+        'next-superjson-plugin',
+        {
```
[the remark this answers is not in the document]
tjheffner: yeah, the plugin config slurps it in automatically

PR #91, opened 2022-08-15
VACMS-10260 queries refactor: linkTeaser

penny-lischer:
## Description

Relates to #10260

## Screenshots
Screen Shot 2022-08-15 at 4 13 17 PM


src/lib/utils/drupalClient.ts
```diff
@@ -2,19 +2,59 @@ import { DrupalClient } from 'next-drupal'
 import crossFetch from 'cross-fetch'
 import { SocksProxyAgent } from 'socks-proxy-agent'
 import { addCAs } from 'syswide-cas'
+import * as fs from 'fs'
```
timcosgrove, line 5: I'd like us to remove changes to this file in favor of what we'll merge in later.

timcosgrove (changes requested):

src/types/dataTypes/drupal/field_type.d.ts
```diff
@@ -23,7 +23,7 @@ export interface FieldFormattedTextWithSummary extends FieldFormattedText {
 export interface FieldLink {
   uri: string
   title: string
-  options: [string[]]
+  options: any[]
```
timcosgrove, line 26: What is prompting this change?

penny-lischer: If the JSON data returns an empty array i.e.options: [] the query tests fail when definition is options: [string[]]. It seems that the data coming back from the response is not of type [string[]].

timcosgrove (approved):

timcosgrove:
Reworked into #94 and merged. (Rework was due to my own issues doing the merge.)

PR #99, opened 2022-08-26
VACMS-Media-Image-Rebase

amponce:
## Description

Relates to #98 . (or closes?)

Acceptance Criteria

- [x] Any data mapping logic & params moved to queries 
- [x] Refactor and rename the component per the description
- [x] Tests are written for the data mapping/formatter piece of the query.
- [x] The data service file is removed

## Testing done
Unit Testing

### Media Image Component
 PASS  src/templates/common/media/index.test.tsx
  Media Image component renders with valid data
    ✓ <MediaImage> renders (128 ms)
    ✓ MediaImage renders with large image (19 ms)
    ✓ MediaImage renders when image style is defined (23 ms)
    ✓ MediaImage does not render with null data (3 ms)

Test Suites: 1 passed, 1 total
Tests:       4 passed, 4 total
Snapshots:   0 total
Time:        2.166 s

### Image Component
 PASS  src/templates/common/image/index.test.jsx
  Image Component
    ✓ throws an error if URL is empty (12 ms)
    ✓ throws an error if no height/width properties are passed in (101 ms)
    ✓ triggers an accessibility error if no alt text is passed in (138 ms)
    ✓ renders if sufficient properties are provided (76 ms)

Test Suites: 1 passed, 1 total
Tests:       4 passed, 4 total
Snapshots:   0 total
Time:        4.055 s, estimated 10 s
Ran all test suites matching /src\/templates\/common\/image/i.
✨  Done in 5.62s.
## Screenshots

Screen Shot 2022-08-26 at 9 35 24 AM
Screen Shot 2022-08-26 at 9 37 29 AM
Screen Shot 2022-08-26 at 9 36 37 AM
Screen Shot 2022-08-26 at 9 37 01 AM


src/data/queries/mediaImage.ts
```diff
@@ -0,0 +1,46 @@
...
+export const data: QueryData<DataOpts, MediaImage[]> = async (): Promise<
+  MediaImage[]
+> => {
+  const entities = await drupalClient.getResourceCollection<MediaImage[]>(
+    'media--image',
+    {
+      params: params().getQueryObject(),
+    }
+  )
+  return entities
```
tjheffner, line 29: are there fields with multiple images that require this to be an array vs single media object? a little unclear to me

amponce: This was confusing to me as well yes there are both types and I had some issues mapping though the data in the query service so I added the logic to mediaImage component directly. Let me know if you see a better path

src/pages/demo/media/index.tsx
```diff
@@ -1,52 +1,51 @@
...
-      }
-    )
-  const mediaProps = generalEntityDataService(media)
-
+export async function getStaticProps() {
+  const mediaImageCollection = await queries.getData(MediaResourceType.Image)
   return {
     props: {
-      mediaProps,
+      MediaCollectionProps: JSON.parse(JSON.stringify(mediaImageCollection)),
```
tjheffner, line 48: we probably don't need to do this now that superjson is included

src/templates/common/media/index.test.tsx
```diff
@@ -1,41 +1,68 @@
...
+  },
+
+  alt: 'Smiling man in glasses.',
+  title: '',
+  width: 1299,
+  height: 1512,
+  url: '/sites/default/files/2020-08/Raab.jpg',
+}
+
+function MediaImageComponent({ imageStyle }: MediaImageType) {
```
tjheffner, line 30: why is this new component in the test vs using the component itself?

src/templates/common/media/index.tsx
```diff
@@ -1,44 +1,52 @@
...
+      setShowImages(true)
+    }
+  }, [optimizedData, props])
+
+  const image = optimizedData[imageProps?.id]
+
+  if (!image) {
+    return null
   }
+  console.log('Image', image)
```
tjheffner, line 34: console.log hanging out

src/templates/layouts/staffProfile/dataService.tsx
```diff
@@ -5,7 +5,7 @@ import {
 } from '@/types/dataTypes/drupal/paragraph'
 import { StaffProfile, StaffProfileProps } from './index'
 import { DrupalJsonApiParams } from 'drupal-jsonapi-params'
-import { mediaImageDataService } from '@/templates/common/media/dataService'
+import { generalEntityDataService } from '@/data/delegators/generalEntityDataService'
```
tjheffner, line 8: can this be removed entirely? doesn't look like it's used in the file from what i can see

package.json
```diff
@@ -57,6 +57,7 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-gtm-module": "^2.0.11",
+    "react-test-renderer": "^18.2.0",
```
tjheffner, line 60: where does this get used?

tjheffner (changes requested):

src/templates/common/media/index.test.tsx
timcosgrove: +1

amponce: fixed, I had that in temporarily and then forgot all about it.

src/data/queries/mediaImage.ts
amponce: Found a better way to handle this

src/data/queries/tests/__snapshots__/newsStory.test.tsx.snap
```diff
@@ -121,18 +121,18 @@ Object {
...
+      "alt": undefined,
+      "height": undefined,
+      "id": "5aa01b90-7156-45b1-85fb-95f82465409c",
+      "link": null,
+      "title": undefined,
+      "type": "media--image",
+      "url": undefined,
+      "width": undefined,
+    },
+  ],
```
amponce, line 135: These null values are a result of the news story mock data not having the expanded field_media relationship

src/templates/components/personProfile/dataService.tsx
```diff
@@ -26,7 +25,7 @@ export const personProfileDataService = function (
     default:
       return {
         title: entity.title,
-        image: mediaImageDataService(entity.field_media, 'full_content_width'),
+        image: queries.formatData('media--image', entity.field_media),
```
tjheffner, line 28: just noting this is a stopgap until personProfile has a query

src/templates/layouts/staffProfile/dataService.tsx
```diff
@@ -27,10 +26,7 @@ export const transformStaffProfileData = function (
       return {
         id: entity.field_staff_profile.id,
         name: name,
-        thumbnail: mediaImageDataService(
-          entity.field_staff_profile.field_media,
-          '1_1_square_medium_thumbnail'
-        ),
+        thumbnail: entity.field_staff_profile.field_media,
```
tjheffner, line 29: same as above, just noting this is a stopgap until staffProfile has a query

src/data/queries/mediaImage.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+export const formatter: QueryFormatter<MediaImage, MediaImageType[]> = (
+  entities: MediaImage
+) => {
+  if (!entities) return null
+
+  const castArray = (val) => (Array.isArray(val) ? val : [val])
```
tjheffner, line 38: this formatter is the only part that jumps out at me still, feels a little weird to need to cast to array and then potentially null/undefined values for all of the `.image?` checks... but i haven't run it locally yet. if this is working i dont want it to be a blocker

tjheffner (approved):

src/data/queries/mediaImage.ts
amponce: Yeah lets talk about this more during the next build sync. The slug route pages return data as an object vs data being returned as an collection array on no slug rout pages. Working with the new story listings is where the data structure is a little bit odd. 

```
{
0: {
id: 3
  link : {
      href: https://
      meta: {
        linkParams : {
         width: 100
         height: 100
        }
      }
    }
  }
}
```

PR #102, opened 2022-09-06
VACMS-10466 nuke demo pages

tjheffner:
## Description

Relates to [#10466 
](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/10466)

## Testing done
ran content-build many many times with these files present and with them missing.

ran local dev, story listing pages and news story pages still render as expected. the default homepage will show a 404 until we have a real homepage present, though

## Screenshots
n/a

## QA steps
run `yarn export` and notice its faster without the demo pages
run `NODE_OPTIONS='--max-http-header-size=50000' yarn build --nosymlink --run-next-build` and notice the next-build portion finishes much faster


timcosgrove (approved):
💥

PR #143, opened 2023-08-14
use string.match() instead of array.includes() to check hostname

tjheffner:
## Description
should fix the problem when building locally using the https://content-build-medc0xjkxm4jmpzxl3tfbcs7qcddsivh.ci.cms.va.gov endpoint. if the codeQL check is okay with it 🤞 

## QA steps

[ ] - does the site resolve correctly locally when using the content build endpoint?
[ ] - when using the local ddev cms endpoint?
[ ] - on tugboat environments?

### DIFF
--- a/src/lib/utils/drupalClient.ts
+++ b/src/lib/utils/drupalClient.ts
@@ -7,11 +7,9 @@ export const baseUrl = process.env.NEXT_PUBLIC_DRUPAL_BASE_URL || 'cms.va.gov'
 const url = new URL(baseUrl)
 const host = url.host
 
-const allowedHosts = ['cms.va.gov']
-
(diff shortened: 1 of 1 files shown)


src/lib/utils/drupalClient.ts
```diff
@@ -39,4 +37,5 @@ export const fetcher = async (input: RequestInfo, init?: RequestInit) => {
 
 export const drupalClient = new DrupalClient(baseUrl, {
   fetcher,
+  useDefaultResourceTypeEntry: true,
```
tjheffner, line 40: setting this to true seems to increase build speed locally by `278548 ms`.

Without: `├ ● /[[...slug]] (1435969 ms)`
With: `├ ● /[[...slug]] (1157421 ms)`

timcosgrove (approved):
r+

PR #149, opened 2023-08-23
adds necessary style sheet for sign on widget styling

jtmst:
## Description
Adds a style sheet that was missing in order for the sign on widget to style correctly.  There will be a seperate vets-website PR to address the crisis modal fix

Relates to #14802. (or closes?)
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/14802
## Testing done
local visual testing
## Screenshots
image

## QA steps
Verify that sign on widget loaded from the main nav renders as expected

### Definition of Done

- [ ] Documentation has been updated, if applicable.
- [ ] Tests have been added if necessary.
- [ ] Automated tests have passed.
- [ ] Code Quality Tests have passed.
- [ ] Acceptance Criteria in related issue are met.
- [ ] Manual Code Review Approved.

### Is this PR blocked by another PR?

- [ ] `DO NOT MERGE`

### DIFF
--- a/src/templates/globals/header/index.tsx
+++ b/src/templates/globals/header/index.tsx
@@ -2,7 +2,7 @@ import { TopNav } from './TopNav'
 
 export const Header = () => {
   return (
-    <header className="header merger" role="banner">
+    <header className="header" role="banner">
       {/*  Mobile Layout */}
       <div
         data-widget-type="header"
--- a/src/pages/_document.tsx
+++ b/src/pages/_document.tsx
@@ -33,7 +33,7 @@ const Document = () => {
           src="https://prod-va-gov-assets.s3-us-gov-west-1.amazonaws.com/generated/static-pages.entry.js"
         />
       </Head>
-      <body>
+      <body className="merger">
         <noscript>
           <iframe
             src={`https://www.googletagmanager.com/ns.html?id=${GTM_ID}`}


tjheffner:
I do see the sign in widget rendering on tugboat, but it feels a little off from va.gov still. Are other stylesheets missing that would effect this?

Note the difference in the margin on the right side of the open widget container
Screenshot 2023-08-23 at 10 24 43 AM
Screenshot 2023-08-23 at 10 24 48 AM

jtmst:
> I do see the sign in widget rendering on tugboat, but it feels a little off from va.gov still. Are other stylesheets missing that would effect this?
> 
> Note the difference in the margin on the right side of the open widget container Screenshot 2023-08-23 at 10 24 43 AM Screenshot 2023-08-23 at 10 24 48 AM

I see the difference but it also looks like the styles are the same.  The css unit used for this is 'ch' which is derived from the font, and we are still having CORS issues preventing the correct fonts from loading in.  The margin issue should fix itself once the fonts are corrected

jtmst:
looking at it again though, im seeing that the sign in button is mis-colored on hover, let me check that out before we merge

tjheffner:
> I see the difference but it also looks like the styles are the same. The css unit used for this is 'ch' which is derived from the font, and we are still having CORS issues preventing the correct fonts from loading in. The margin issue should fix itself once the fonts are corrected

Copy that. I have noticed some issues related to `em` units being used as well (not in this PR, but with some font stuff in general). We'll have to fix the CORS stuff sooner than later (also not in this PR)

src/templates/globals/header/index.tsx
```diff
@@ -2,7 +2,7 @@ import { TopNav } from './TopNav'
 
 export const Header = () => {
   return (
-    <header className="header" role="banner">
+    <header className="header merger" role="banner">
```
tjheffner, line 5: this fixes the header styles, but will you move this onto the `<body>` tag in `_document.tsx` instead? That's where I'm seeing it on va.gov, other non-header components may also need that class in their scope

jtmst: good catch, fixed.

tjheffner (approved):

PR #155, opened 2023-08-31
VACMS-14824: Pagination for story listing pages

ryguyk:
## Description

Relates to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/14824

Adds pagination to story-listing pages by statically generating additional pages according to the total number of stories and page size. For example:
- Listing page `some-health-care-system/stories` contains 25 stories.
- Page size is 10 stories.
- `some-health-care-system/stories` is kept as first page, and `some-health-care-system/stories/page-2` and `some-health-care-system/stories/page-3` are added.

## QA steps
Note: this will likely fail locally. This does not mean something with this PR is the issue. Rather, this PR introduces the need to hit Drupal with additional requests, and Drupal is falling down as a result sometimes.
1. On Tugboat:
    - [x] Validate that build succeeds
    - [x] Validate the existence of some paginated listing pages
       - [x] `eastern-oklahoma-health-care/stories/` should exist
       - [x] `eastern-oklahoma-health-care/stories/page-2` should exist
       - [x] `eastern-oklahoma-health-care/stories/page-7` should exist
       - [x] `eastern-oklahoma-health-care/stories/page-8` should NOT exist

2. If run locally, you can slice the data results to deal with a smaller subset (see below):
    - [x] Validate that build succeeds
    - [x] Validate that the `out` directory contains some paginated listing pages
       - [x] `eastern-oklahoma-health-care/` should exist
       - [x] `eastern-oklahoma-health-care/stories/page-2` should exist
       - [x] `eastern-oklahoma-health-care/stories/page-7` should exist
       - [x] `eastern-oklahoma-health-care/stories/page-8` should NOT exist
```
/* src/lib/utils/staticPaths.ts */
export async function getStaticPathsByResourceType(
  resourceType,
  context: GetStaticPathsContext
): ReturnType<typeof drupalClient.getStaticPathsFromContext> {
  return (await drupalClient.getStaticPathsFromContext([resourceType], context)).slice(0, 5)
}
```


src/lib/utils/staticPaths.ts
```diff
@@ -0,0 +1,28 @@
...
+import { getAllPagedListingPaths } from '@/lib/utils/listingPages'
+import RESOURCE_TYPES from '@/lib/constants/resourceTypes'
+
+export async function getStaticPathsByResourceType(
+  resourceType,
+  context: GetStaticPathsContext
+): ReturnType<typeof drupalClient.getStaticPathsFromContext> {
+  return (
+    await drupalClient.getStaticPathsFromContext([resourceType], context)
+  ).slice(0, 10)
```
tjheffner, line 12: i think we want to lose this slice

src/pages/[[...slug]].tsx
```diff
@@ -26,49 +30,68 @@ export default function ResourcePage({ resource, globalElements }) {
...
     return {
       notFound: true,
     }
   }
 
-  const resource = await queries.getData(type, {
+  const resource = await queries.getData(resourceType, {
     context,
-    id: path.entity.uuid,
+    id: pathInfo?.entity?.uuid,
```
Comment (tjheffner) [inline src/pages/[[...slug]].tsx:90]: why are these ? added here? if there's not pathinfo, we return up above. if we get to this point, it should have all the data we need, no?

src/templates/layouts/storyListing/index.tsx
```diff
@@ -44,6 +47,23 @@ export function StoryListing({
...
           </div>
+
+          {totalPages > 1 && (
+            <VaPagination
+              page={currentPage}
+              pages={totalPages}
+              maxPageListLength={3}
+              onPageSelect={(page) => {
+                const newPage =
+                  page.detail.page > 1 ? `page-${page.detail.page}` : ''
```
tjheffner, line 58: ah, this was a missing piece for me on how to drop the `page-1`. nice.

src/lib/utils/slug.ts
```diff
@@ -0,0 +1,54 @@
+/**
```
tjheffner, line 1: i'd almost argue for this to be included in the `utils/listingPages` file instead (i realize it is used in `getStaticProps` instead of `getStaticPaths`, but it feels more like a listingPage utility than a slug utility to me personally)

tjheffner (changes requested):
Overall, this is very slick and I appreciate you digging in on this. I only have a few minor nits and questions. 

The cache files failing in tugboat is a weird one. I am going to dig into that.

src/data/queries/storyListing.ts
```diff
@@ -36,13 +43,19 @@ export const data: QueryData<DataOpts, StoryListingData> = async (opts) => {
...
+
+  // Fetch list of stories related to this listing. `deserialize: false` for jsonapi pagination
+  const stories = await drupalClient.getResourceCollection<TJsonaModel>(
     'node--news_story',
     {
       params: queries
         .getParams('node--news_story--teaser')
         .addFilter('field_listing.id', entity.id)
+        .addSort('-changed')
+        .addPageLimit(10)
```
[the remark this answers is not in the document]
ryguyk: Absolutely. This is precisely why this is moved to a constant. I had this in place and lost it when I started over on some things. I forgot to put it back. Great catch!

src/lib/constants/resourceTypes.ts
```diff
@@ -0,0 +1,5 @@
+export default {
+  STORY_LISTING: 'node--story_listing',
+  STORY: 'node--news_story',
+  QA: 'node--q_a',
+} as const
```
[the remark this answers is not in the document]
ryguyk: I suppose we _could_ use `typeof slug === undefined` but you're absolutely right that's a bit unnatural and verbose. This was a complete brain fart on my part caused by checking `typeof slug === 'string'` below.

To the question about when it's undefined, yeah, it's for the homepage, I believe. I didn't initially have this check and then caught it in manual testing when it came through as undefined for the homepage, IIRC.

src/lib/utils/staticPaths.ts
ryguyk: Definitely. I forgot to remove this.

src/lib/utils/slug.ts
```diff
@@ -0,0 +1,54 @@
...
+ * ]
+ * output: [
+ *  'minneapolis-health-care/stories/some-story-title`,
+ *  undefined
+ * ]
+ */
+export function getPathAndPageNumberFromSlug(
+  slug: string | string[]
+): [string, number] {
+  if (typeof slug === undefined || slug?.length === 0) {
```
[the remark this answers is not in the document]
ryguyk: You're absolutely right. It's a bit unnatural and verbose - and broken. It would need to be `typeof slug === 'undefined'`. This was a complete brain fart on my part caused by checking `typeof slug === 'string'` below.

To the question about when it's undefined, yeah, it's for the homepage, I believe. I didn't initially have this check and then caught it in manual testing when it came through as undefined for the homepage, IIRC.

src/pages/[[...slug]].tsx
Comment (ryguyk) [reply src/pages/[[...slug]].tsx:90]: I tend to get overdefensive on some of this prop drilling, so this is probably partially out of habit. After all, there's nothing inherently _wrong_ with having the extra checks. It ultimately boils down to how we are trapping expected errors, etc. 

But I agree it is partially unnecessary here. We do know that `pathInfo` will be defined, so we don't need the first `?`. If we have `pathInfo.entity`, it will never result in a `ReferenceError`. I think that means it's likely _almost_ impossible for `entity` to not be defined, because we know what `pathInfo` _should_ look like, but I think we might be remiss to make that assumption. The goal here was to ensure that `undefined` would be passed rather than raising an uncaught error. As I said, I don't think it's "wrong" as it is, but I really appreciate your thorough review and attention to detail here, and I agree that this isn't the most pure way of handling this. My vote would be for removing the first `?` and keeping the second `?`. 

Note: This will have the result of possibly passing `undefined` to the `getData` function. We might need to look at that function to make sure it's handling bad data properly.

package.json
```diff
@@ -47,6 +47,7 @@
     "debug": "^4.3.4",
     "drupal-jsonapi-params": "^2.0.0",
     "jest-axe": "^6.0.0",
+    "jsona": "^1.11.0",
```
ryguyk, line 50: @tjheffner Am I recalling that this was something you were wrestling with previously? I got some weirdness around this, but I thought this was already in `package.json`. Do you recall?

ryguyk: Actually, I'm remembering now that my branch was based on yours, so that's probably why it's showing up here.

src/lib/constants/resourceTypes.ts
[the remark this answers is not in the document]
ryguyk: Your thought here resonates. It felt natural to have something in `[[...slug]].tsx`, and moving it out left it feeling like we removed something that specifically applied to that file, namely, a data point that indicates the types of resources being built by that template.

But I think, for a couple reasons, that _this_ constant belongs where it now is, in `constants/resourceTypes.ts`:
1. This object is required in multiple places now. It's currently used in three files (`[[...slug]].tsx` plus two others).
2. This object is a mapping for _all_ possible resource types. We'll need something like this throughout the application, likely. It's a _different_ data object than what was previously in `[[...slug]].tsx`. It occurred to me at some point that we _might_ want to still have something in `[[...slug]].tsx` as well that indicates all the resource types that are built by that template, but since my changes ultimately split up the fetches of those resource types into individual calls, it wasn't strictly necessary (at least yet). To paint the picture more clearly, it might be useful (at some point) to have something like this in the `[[...slug]].tsx` file:
```
// [[...slug]].tsx
import RESOURCE_TYPES from '@/lib/constants/resourceTypes'
const TEMPLATE_RESOURCE_TYPES = [
  RESOURCE_TYPES.STORY_LISTING,
  RESOURCE_TYPES.NEWS.STORY,
  RESOURCE_TYPES.QA
]
```

That might be even more clear if we think that at some point we might have another file that builds a specific resource type. In fact, we toyed with this even for this PR, initially. We had the `[facility]/stories/[[...page]]` path. If that file were going to build, say, all of our story-listing pages, then we would have the following:
```
// [[...slug]].tsx
import RESOURCE_TYPES from '@/lib/constants/resourceTypes'
const TEMPLATE_RESOURCE_TYPES = [
  RESOURCE_TYPES.NEWS.STORY,
  RESOURCE_TYPES.QA
]

```
```
//[facility]/stories/[[...page]].tsx
import RESOURCE_TYPES from '@/lib/constants/resourceTypes'
const TEMPLATE_RESOURCE_TYPES = [
  RESOURCE_TYPES.STORY_LISTING,
]
```

package.json
tjheffner: . It was a sub-dependency in our installed modules already, but because we have to deserialize paginated responses manually in `data/queries/storyListing` (`TJsonaModel`) I needed to explicitly add it as a dependency to satisfy eslint

src/pages/[[...slug]].tsx
Comment (tjheffner) [reply src/pages/[[...slug]].tsx:90]: Why would we ever want to pass `undefined` though? We're inside getStaticProps, we'll need the uuid every time. If it's a paginated page, we still want the uuid of the base listing page and then the correct page #

src/lib/utils/slug.ts
ryguyk: I had this exact same dilemma. I initially wrote it with verbiage specific to listing pages, in fact. But the real catch here is that this function isn't _only_ for listing pages. You can see that it is called for _all_ pages. It is primarily meant to simplify the `getStaticProps` function by removing a ton of conditional logic out of it. I didn't want to bloat that function with a lot of checks for various things. It seemed a _lot_ cleaner to me to simply have a single line of code that would standardize things for all possible slugs (`const [path, pageNumber] = getPathAndPageNumberFromSlug(slug)`).

It might help clarify my point by considering that this function ultimately does this:
```
if (isListingPage(slug)) {
  return getListingPagePathAndPageNumberFromSlug(slug)
} else {
  return [getOtherPagePathFromSlug(slug), undefined]
}
```

It's not written in this fashion exactly, but that's one way to consider what it's doing. If it were written in this way, we could then have the `getListingPagePathAndPageNumberFromSlug` function defined in `utils/listingPages.ts`. 

I considered this approach, but ultimately didn't write the function in that way because I went the route of writing it a bit more general. I didn't want to think specifically of a certain resource type (e.g. story listings). Instead, it made sense to just check for any path that has `page-{n}` in the last slug segment. Since I went this route, I didn't need anything specifically in `utils/listingPages.ts` because, again, the function isn't specific to listing pages. It's called for _every_ page. After all, every page comes in as a slug and needs to be converted to a path.

src/lib/constants/resourceTypes.ts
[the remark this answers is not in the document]
tjheffner: Oh, I certainly agree with you that it belongs here now. It just feels slightly less clear to me with it removed from slug, but it does clear things up overall for re-use... and I'll get used to it 😁

src/lib/utils/slug.ts
tjheffner: it does run on every path, but it's specifically finding page numbers for the benefit of listing pages. if it isn't a listing page, it returns the path and keeps it moving. in my mind, that logic really feels like a listingPage utility (whether story listing, event listing, etc etc), but i don't think it's a blocker here, just a matter of semantics

ryguyk: While I still think the function I wrote would belong outside `utils/listingPages.ts`, this conversation made me reconsider the way I was structuring the logic in the first place. My initial goal of abstracting away the conditional logic around listing pages in `getStaticProps` ultimately felt confusing. The function `getPathAndPageNumberFromSlug` was, indeed, called for _every_ slug, but the page number was ultimately not applicable for non-listing pages. So, I refactored this and introduced some conditional logic based on whether a page is a listing page. I think this ultimately makes the `getStaticProps` function more clear, if slightly more verbose.

ryguyk: And, to your initial point @tjheffner, this allows the move of the page-number logic to `utils/listingPages.ts` where it belongs. It's now just in a function that _is_ specific to listing pages.

src/pages/[[...slug]].tsx
Comment (ryguyk) [reply src/pages/[[...slug]].tsx]: It's not that we would _want_ to pass `undefined`. It's that we want to handle the possibility that it _is_ `undefined` ("it" is actually `pathInfo.entity`, to be clear).

### `pathInfo.entity?.uuid`
In this case, if `pathInfo.entity` is `undefined`, we set `id: undefined`. The call to `queries.getData` _should_ handle this possibility and return `undefined` as the resource.
### `pathInfo.entity.uuid`
In this case, if `pathInfo.entity` is undefined, an error is thrown and things break.

src/lib/utils/slug.ts
tjheffner: i love the refactor, thank you

tjheffner:
Overall this looks really good and tugboat didn't report errors building.... but I am noticing some pages don't appear to render correctly:

https://pr155-ldbhehn1t70aydz5grjdijkm9rdelo07.tugboat.vfs.va.gov/eastern-oklahoma-health-care/stories/ paginates all the from 0 - 7 correctly, but none of the stories until near the bottom of page 2 actually resolve correctly for me. 

https://pr155-ldbhehn1t70aydz5grjdijkm9rdelo07.tugboat.vfs.va.gov/eastern-oklahoma-health-care/stories/annual-dinner-for-veterans-still-going-strong-after-five-years/ is the first individual story that resolves, (3 from the bottom on page 2, every story newer than that 404s). 

https://pr155-ldbhehn1t70aydz5grjdijkm9rdelo07.tugboat.vfs.va.gov/butler-health-care/stories/ the first story that resolves correctly is Clearing the Air About Smoking  (second from the bottom on page 1)

tjheffner (approved):
The "missing" pages above is not an issue introduced by this pagination approach, which is wonderful. The main tugboat preview is also missing these pages. va.gov is not, and the paginated lists match that.

I'm going to merge and we will follow up on these missing pages.

PR #165, opened 2023-09-12
adds mission comment to footer

jtmst:
## Description
Relates to #[14875](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/14875)

## Testing done
Local visual testing

## Screenshots
Screen Shot 2023-09-12 at 10 09 55 AM

## QA steps

```[tasklist]
- [ ] Verify comment is in footer
```

### DIFF
--- a/src/templates/globals/utils/FooterComment.tsx
+++ b/src/templates/globals/utils/FooterComment.tsx
@@ -1,0 +1,29 @@
+import { useEffect } from 'react'
+
+const FooterComment = () => {
+  const commentText = `\n“To care for him who shall have borne the battle and for his widow, and his orphan.”\n- Abraham Lincoln\n`
+
+  useEffect(() => {
+    const comment = document.createComment(commentText)
+
+    const footer = document.querySelector('footer')
+    if (footer) {
+      const firstChild = footer.firstChild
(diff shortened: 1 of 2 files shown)


tjheffner (commented):
I like this approach, but I have two things.

1. i'd either include this directly in the `footer/index.tsx` file (a separte const above the exported footer const should be fine)

or 

2. make the hook generic by accepting a comment string & an element to be passed to the `document.querySelector`. 

if you go with option 2, the hook could be re-used for the development info comment in the head. it could either live in utils or in src/hooks, depending on how it's structured

src/templates/globals/utils/FooterComment.tsx
```diff
@@ -0,0 +1,29 @@
+import { useEffect } from 'react'
+
+const FooterComment = () => {
```
timcosgrove, line 3: Is it possible to have a general `Comment` component and pass in the comment as a child prop?

jtmst: good call as I'll be using this kind of comment in the head as well, will refactor

timcosgrove:
We will not be doing this.

PR #178, opened 2023-09-22
move some stuff in lib around

tjheffner:
## Description
No ticket, but cleans up the `src/lib/` directory a little bit, so everything isn't just shoved into `src/lib/utils`.

I did not remove analytics, it is a little more intertwined than I thought. We should revisit the implementation in a future ticket.

## QA steps

```[tasklist]
- [ ] `yarn dev`, `yarn build`, `yarn export` all work as expected
```


jtmst (approved):
Definitely prefer this organization, nice!

PR #179, opened 2023-09-25
VACMS-15065 use swc instead of babel

tjheffner:
## Description
Closes #[15065](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15065).

- removes the `.babelrc` file
- yarn deps associated with that babel plugin updated/removed
- also comments out the `/resource/` node type (QA) for now. we only need to focus on story + story listing, don't want to go to prod with `/resource/` paths enabled before they are ready.

## Testing done
`yarn build`
`yarn export`

No notice about using babel over SWC is thrown when running those commands, they still complete as expected.

## Screenshots
Before:
```
yarn build
- info Linting and checking validity of types...
- info Creating an optimized production build...
- info Disabled SWC as replacement for Babel because of custom Babel configuration ".babelrc" https://nextjs.org/docs/messages/swc-disabled
- info Using external babel configuration from /var/lib/tugboat/.babelrc
- info Compiled successfully
...
```

After:
```
yarn build
- info Linting and checking validity of types
- info Creating an optimized production build
- info Compiled successfully
...
```

## QA steps

```[tasklist]
```


jtmst (approved):
lgtm

PR #180, opened 2023-09-25
VACMS-15239 can build & run cypress on tugboat

tjheffner:
## Description
Closes to #[15239](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15239).

Currently always marks Cypress tests as passing, after they run. Additional pass/fail handling, surfacing screenshots, running the full sitemap scan in the ASG workflow, etc. will be needed once the underlying issue causing all tests to fail always is sorted out. That will be a different PR.

## Testing Done
- lots of commits and rebuilding tugboat environment

## Screenshots
Screenshot 2023-09-28 at 8 46 11 AM
Screenshot 2023-09-28 at 8 49 13 AM
Screenshot 2023-09-28 at 10 00 22 AM

## QA steps
```[tasklist]
```


KitLykos (approved):
My eyes are new but nothing jumped out at me.

.github/workflows/ci.yml
```diff
@@ -17,21 +17,21 @@ jobs:
...
-
-      - name: Cache yarn dependencies
-        uses: actions/cache@v3
-        id: yarn-cache
-        with:
-          path: ${{steps.yarn-cache-dir-path.outputs.dir }}
-          key: yarn-cache-${{ hashFiles('**/yarn.lock', '.yarnrc.yml') }}
-          restore-keys: |
-            yarn-cache-
+      # - name: Get yarn cache directory path
```
timcosgrove, line 20: Is disabling this necessary for this PR?

.github/workflows/cypress.yml.disabled
```diff
@@ -6,134 +6,42 @@ on:
   pull_request:
 
 env:
-  INSTANCE_TYPE: m5.large
   NODE_EXTRA_CA_CERTS: /etc/ssl/certs/ca-certificates.crt
   OPENSSL_CONF: /etc/ssl/openssl.cnf
+  CYPRESS_INSTALL_BINARY: 0
+  CYPRESS_CACHE_FOLDER: /home/runner/runner1/_work/next-build/next-build/.yarn/cypress/cypress.zip
```
timcosgrove, line 12: `runner1` is not guaranteed; there are usually 3 runners per ASG instance. This would need to be relative or otherwise variable.

.github/workflows/cypress.yml.disabled
```diff
@@ -6,134 +6,42 @@ on:
...
-              {"Key": "VAECID", "Value": "AWG20180517003"},
-              {"Key": "environment", "Value": "utility"}
-            ]
-
   build:
     name: Build static
-    runs-on: ${{ needs.start-runner.outputs.label }}
-    needs: start-runner
-    timeout-minutes: 5
+    runs-on: [self-hosted, asg]
```
timcosgrove, line 18: 🎉

.github/workflows/cypress.yml.disabled
```diff
@@ -6,134 +6,42 @@ on:
...
-        id: check-node-env
-        run: echo $NODE_ENV
+         corepack enable
+         corepack hydrate ./.yarn/releases/corepack.tgz
+         corepack prepare yarn@3.6.1 --activate
+
+      # todo: figure out where/how to cache this
+      # necessary to do this way because cypress postinstall command ignores ssl/certs settings
+      - name: Download Cypress Binary
+        run: curl https://download.cypress.io/desktop/13.2.0?platform=linux&arch=x64 > /home/runner/runner1/_work/next-build/next-build/.yarn/cypress/cypress.zip
```
timcosgrove, line 35: As long as we're doing this, the destination could be `$CYPRESS_CACHE_FOLDER` I think. But, see above note re: `runner1` not being a given.

.github/workflows/cypress.yml.disabled
```diff
@@ -6,134 +6,42 @@ on:
...
+      
+      - name: Install Dependencies
+        run: HUSKY=0 yarn install --immutable
 
       - name: Generate static site
-        run: yarn export
+        run: NODE_EXTRA_CA_CERTS=/etc/ssl/certs/ca-certificates.crt yarn export
```
timcosgrove, line 41: If `NODE_EXTRA_CA_CERTS` is set to this above, this shouldn't be necessary here. If it is necessary, something is not quite right (I think).

package.json
```diff
@@ -92,7 +92,7 @@
     "@types/uuid": "^8.3.4",
     "@typescript-eslint/eslint-plugin": "^5.19.0",
     "@typescript-eslint/parser": "^5.0.0",
-    "cypress": "13",
+    "cypress": "13.2.0",
```
timcosgrove, line 95: is it normal for us to specify specific versions of Cypress rather than semver versions?

timcosgrove (commented):
questions

package.json
tjheffner: no, this should be semver'd. i'll adjust

.github/workflows/ci.yml
tjheffner: In the course of re-triggering workflows on this PR a bunch, I was noticing the steps after the install step would sometimes repeatedly fail if the install step wasn't run. I'll try reverting this and see if it plays nicely again

.github/workflows/cypress.yml.disabled
tjheffner: yeah good callout. this workflow is still disabled for the time being but i'll add a comment

PR #181, opened 2023-09-25
VACMS-15062: Handle Lovell paths

ryguyk:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15062

## Testing done/QA Steps
- [ ] Ensure Lovell bifurcation works
  - Here's an example of a story that is assigned to Lovell Federal: https://content-build-medc0xjkxm4jmpzxl3tfbcs7qcddsivh.ci.cms.va.gov/lovell-federal-health-care-va/stories/nmrtc-great-lakes-blood-donor-center-meets-mission
    - Ensure it has TWO pages built for it:
      - https://pr181-qc3cy62gmybmm8mfl4i3qkpe0tpnfgnx.tugboat.vfs.va.gov/lovell-federal-health-care-va/stories/nmrtc-great-lakes-blood-donor-center-meets-mission/
      - https://pr181-qc3cy62gmybmm8mfl4i3qkpe0tpnfgnx.tugboat.vfs.va.gov/lovell-federal-health-care-tricare/stories/nmrtc-great-lakes-blood-donor-center-meets-mission/

- [ ] Ensure Lovell Federal has no listing page, but TRICARE and VA do have listing pages.
  - This should NOT exist: https://pr181-qc3cy62gmybmm8mfl4i3qkpe0tpnfgnx.tugboat.vfs.va.gov/lovell-federal-health-care/stories/
  - This should exist: https://pr181-qc3cy62gmybmm8mfl4i3qkpe0tpnfgnx.tugboat.vfs.va.gov/lovell-federal-health-care-tricare/stories/
  - This should exist: https://pr181-qc3cy62gmybmm8mfl4i3qkpe0tpnfgnx.tugboat.vfs.va.gov/lovell-federal-health-care-va/stories/

- [ ] Ensure listing-page pagination is not broken
  - Eastern Oklahoma should have SEVEN pages:
    - This should exist: https://pr181-qc3cy62gmybmm8mfl4i3qkpe0tpnfgnx.tugboat.vfs.va.gov/eastern-oklahoma-health-care/stories
    - This should exist: https://pr181-qc3cy62gmybmm8mfl4i3qkpe0tpnfgnx.tugboat.vfs.va.gov/eastern-oklahoma-health-care/stories/page-7
    - This should NOT exist: https://pr181-qc3cy62gmybmm8mfl4i3qkpe0tpnfgnx.tugboat.vfs.va.gov/eastern-oklahoma-health-care/stories/page-8


src/lib/constants/resourceTypes.ts
```diff
@@ -1,5 +1,8 @@
-export default {
+export const RESOURCE_TYPES = {
   STORY_LISTING: 'node--story_listing',
   STORY: 'node--news_story',
   QA: 'node--q_a',
 } as const
+
+export type ResourceTypeType =
```
tjheffner, line 7: `ResourceTypeType` and `ListingResourceTypeType` both feel... funky to me. I realize the constant is already SNAKE_CASE_TYPE, but is it too confusing / collide-y to have the types simply be `ResourceType` and `ListingResourceType` ?

src/lib/drupal/staticPaths.ts
```diff
@@ -1,80 +1,60 @@
...
+import { RESOURCE_TYPES, ResourceTypeType } from '@/lib/constants/resourceTypes'
+import { queries } from '@/data/queries'
+import { StaticPathResourceType } from '@/types/index'
+import {
+  bifurcateLovellFederalPathResources,
+  removeLovellFederalPathResources,
+} from '@/lib/utils/lovell'
+
+/**
+ * Returns a static-path resource collection that is modified per business logic.
```
tjheffner, line 16: love this comment, love this structure

src/lib/drupal/staticPaths.ts
```diff
@@ -1,80 +1,60 @@
...
+ * E.g.
+ * - Resources are ADDED for subsequent listing pages (e.g. `stories/page-2`)
+ * - Resources are ADDED for Lovell bifurcation (Federal pages become VA *and* TRICARE)
+ * - Lovell Federal listing pages are REMOVED
+ */
+async function modifyStaticPathResourcesByResourceType(
+  resourceType: ResourceTypeType,
+  resources: StaticPathResourceType[]
+): Promise<StaticPathResourceType[]> {
+  if (resourceType === RESOURCE_TYPES.STORY) {
```
tjheffner, line 28: assuming in the future most content types would get added here, except for the listing types which are handled below?

tjheffner (commented):
This is wonderful stuff. The static-path-resources query is slick, that's a nice refactor.

I looked at the links in your QA steps, everything seems to be working as far as pages generated, though as noted in slack the lovell story listings don't actually have stories connected to them yet (likely because the stories are assigned a level up, to the federal list)

I have a couple questions, and a desire to see tests (but the tests can absolutely be a follow-up / handled by multiple people -- It doesn't have to all be on you here in this PR). This work really handles _a lot_ of the Lovell question marks.

jtmst (approved):
Looks very clean to me, nice work!

src/lib/constants/resourceTypes.ts
ryguyk: I had this thought. I'm open to it. The make-everything-make-perfect-sense part of me tells me it should be `...TypeType`, as that is technically "correct", but I can certainly understand why that feels a bit funky.

src/lib/drupal/staticPaths.ts
ryguyk: Probably, though it's really a matter of what business cases we need to handle (both known cases that still need to be implemented, and unknown future cases). It might be the case that something needs to be run on _everything_, for example, in which case we wouldn't need to explicitly have a condition for every resource type. In that case, we'd have a call outside any condition. It's hard to say, but this implementation handles our current business logic.

src/lib/constants/resourceTypes.ts
ryguyk: It makes me think that a good rule of thumb might be to not name things `...Type` when working in TypeScript. So we could rename `RESOURCE_TYPES` to `RESOURCE_CATEGORIES` or something along those lines. But that would be a _lot_ of changes at this point.

I don't love `ResourceTypeType` but I think I might prefer it still to `ResourceType`, because I think, in general, it'd be very common to need both `ResourceType` _and_ `ResourceTypeType` (or whatever we'd name it). In other words, I do think this _might_ create a collision at some point, or at least theoretically could when considering the general problem space.

But, again, I'm open to change if people want that. I'm not going to die on this hill. I might not even take a nap on it.

tjheffner (approved):

PR #183, opened 2023-09-29
VACMS-15313 sidenav update

tjheffner:
## Description
Closes #[15313](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15313).

- Removes the custom facilityMenu component that was in place in favor of the vets-website widget.
- Dynamically adds sidenav data to window object
- Refactor the headerFooter data to follow same approach. Promo blocks in the megamenu appear if they are correctly referenced by the menu item in the CMS. There is a ticket here: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15491 that should connect the missing ones.

## Screenshots
Using the widget
Screenshot 2023-09-29 at 11 17 33 AM

## QA steps
```[tasklist]
- [ ] Automated tests have passed
- [ ] Tugboat environment was generated without errors
- [ ] Side navigation renders on story listing pages
- [ ] Header & Footer render as expected on all pages
```

PR #184, opened 2023-10-03
initial stab at building vets-web as part of our tugboat

tjheffner:
## Description
Relates to #[#14977](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/14977).

We were sourcing the main entry point from the correct place before, but then the bundle looked for local asset paths anyways. So now I am providing those local asset paths on tugboat.

This should unblock #183 to be tested on tugboat, because the asset paths for the side nav widget will able to properly load.

## Testing done
Checked the generated tugboat environment and saw that assets from vets-website were present and available to the build:
Screenshot 2023-10-03 at 4 48 11 PM

Compare with the main preview tugboat environment, which has a ton of additional errors in the browser console:
Screenshot 2023-10-03 at 4 48 30 PM

## Screenshots
https://pr184-6ojykrqnalsqywbph3phj1v1lvlufaqq.tugboat.vfs.va.gov/butler-health-care/stories/

## QA steps
```[tasklist]
- [ ] Header megamenu should work as expected on tugboat
```

### DIFF
--- a/.tugboat/config.yml
+++ b/.tugboat/config.yml
@@ -33,19 +33,19 @@ services:
         # Install Node.js from the new repository
         - sudo apt-get install -y nodejs
 
-        # Install nvm
-        # - ./scripts/install-nvm.sh
-        # - nvm use 14.15.0 
-
         # Install yarn
         - corepack enable
+
+        # Install nvm
+        - chmod -R 775 "./scripts/"
+        - ./scripts/install-nvm.sh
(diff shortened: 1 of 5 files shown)


tjheffner:
Noting the CI test failure is related to the yarn cache being found, so an install isn't run... then later steps complain that an install hasn't been run.

I disable the check for the cache in #183, it can be safely ignored here.

KitLykos (approved):

PR #190, opened 2023-10-06
clean up header queries

tjheffner:
## Description
Relates to #183. @ryguyk pointed out how to use addInclude properly, so we don't need to make additional queries for the promo blocks. They correctly come along for the ride with the menu data, which simplifies some of the logic when constructing the data into shape. Nice! 

## Testing done
- visit any story or story listing page, check the megamenu and footer and see that promo blocks are present on sections that have them in the cms (top-level `About VA`, or the `Disability`, `Education` hubs underneath the first item)

## QA steps
```[tasklist]
```

### DIFF
--- a/src/data/queries/headerFooter.ts
+++ b/src/data/queries/headerFooter.ts
@@ -7,14 +7,13 @@ import {
 import { queries } from '.'
 import { drupalClient } from '@/lib/drupal/drupalClient'
 import { Menu, HeaderMegaMenu } from '@/types/dataTypes/drupal/menu'
-import { HeaderFooterData, MegaMenuPromoColumn } from '@/types/index'
+import { HeaderFooterData } from '@/types/index'
 import { buildHeaderFooterData } from '@/lib/utils/headerFooter'
 
 export type RawHeaderFooterData = {
   footerColumns: Menu
   footerBottomRail: Menu
   headerMegaMenu: HeaderMegaMenu
-  promoBlocks: MegaMenuPromoColumn[]
 }
 
 // Define the query params for fetching footer menu data.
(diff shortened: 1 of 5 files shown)


src/data/queries/headerFooter.ts
```diff
@@ -7,82 +7,68 @@ import {
...
+      'field_promo_reference.field_promo_link',
+    ])
 }
 
 // Define the option types for the data loader.
 type DataOpts = QueryOpts<{
   params?: object
   megaMenuParams?: object
-  promoBlockParams?: object
 }>
```
ryguyk, line 40: This should not be necessary. Our `data` function does not actually require any input `opts`. `params` and `megaMenuParams` are defined within this file and can be accessed directly as we now have it below. We should be able to remove this type definition (as well as its usage below; more commentary there).

src/data/queries/headerFooter.ts
```diff
@@ -7,82 +7,68 @@ import {
...
 
 export const data: QueryData<DataOpts, RawHeaderFooterData> = async (opts) => {
```
ryguyk, line 42: This function does not actually require any input. I think we can change to:
```
export const data: QueryData<null, RawHeaderFooterData> = async () => {
```

src/lib/utils/header.ts
```diff
@@ -105,41 +106,29 @@ function makeColumns(hostUrl, linkData, arrayDepth, promo): MegaMenuLinkObject {
     }
 
     // Generate promo block column if present
-    if (promo !== undefined || null) {
-      columns[columnNames[i]] = promo
+    if (promo && promo.id) {
+      columns[columnNames[i]] = queries.formatData(
+        'block_content--promo',
+        promo
+      )
```
ryguyk, line 113: I find myself wondering if we still need a separate query file for block promos. Will we ever need to query this separately? Or should the formatting logic for promos be moved into this file?

ryguyk (changes requested):
One change request and one question for discussion.

src/data/queries/headerFooter.ts
tjheffner: good callout, will nuke

src/lib/utils/header.ts
tjheffner: I am not 100% sure yet. These promo blocks are also included on the main benefits hubs pages in addition to the header megamenu. I think the safe route is to keep the formatting logic separate from this header file for now. 

The query may be redundant but the formatting will be used in both places

ryguyk: Sounds fair!

ryguyk (approved):

PR #193, opened 2023-10-11
15056 breadcrumb component

jtmst:
## Description
Relates to #[15056](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15056). (or closes?)

Changes:
- Adds breadcrumb component to next-builld, mimicking the functionality in content-build but using the VA design system breadcrumbs web-component.
- Modifies queries to include breadcrumbs in resource objects where needed
- Adds unit tests for breadcrumb component
- Adds unit tests for breadcrumb utility functions
- Adds storyboard component
- Adds the component to newsStory and storyListing layouts

## Testing done
Local visual testing

## Screenshots
image

## QA steps
- verify that breadcrumbs render on appropriate pages
- verify styles match prod

### DIFF
--- a/src/templates/common/breadcrumbs/index.tsx
+++ b/src/templates/common/breadcrumbs/index.tsx
@@ -4,11 +4,10 @@ import {
   deriveLcBreadcrumbs,
   filterInvalidCrumbs,
 } from '@/templates/globals/util/breadcrumbUtils'
-import { BreadcrumbItem } from '@/types/index'
+import { BreadcrumbItem } from '@/types/dataTypes/drupal/field_type'
 
 interface BreadcrumbProps {
   breadcrumbs?: BreadcrumbItem[]
-  breadcrumbsOverride?: BreadcrumbItem[]
   breadcrumbTitle?: string
   disableAnalytics?: boolean
   deriveBreadcrumbsFromUrl?: boolean
@@ -21,7 +20,6 @@ interface BreadcrumbProps {
 }
 const Breadcrumbs = ({
   breadcrumbs,
-  breadcrumbsOverride,
   entityPath,
   breadcrumbTitle,
   disableAnalytics,
@@ -32,27 +30,24 @@ const Breadcrumbs = ({
   hideHomeBreadcrumb,
   customCrumbHomeText,
 }: BreadcrumbProps) => {
-  const breadcrumbData: BreadcrumbItem[] = breadcrumbs
-  let crumbs = breadcrumbsOverride || breadcrumbData
-
-  if (!crumbs) return null
+  if (!breadcrumbs) return null
 
   if (hideHomeBreadcrumb) {
     if (customCrumbHomeText) {
-      crumbs = crumbs.map((crumb) => {
+      breadcrumbs = breadcrumbs.map((crumb) => {
         if (crumb.title === 'Home') {
           return { ...crumb, title: customCrumbHomeText }
         }
         return crumb
       })
     } else {
-      crumbs = crumbs.filter((crumb) => crumb.title !== 'Home')
+      breadcrumbs = breadcrumbs.filter((crumb) => crumb.title !== 'Home')
     }
   }
 
   if (deriveBreadcrumbsFromUrl) {
-    crumbs = deriveLastBreadcrumbFromPath(
-      breadcrumbData,
+    breadcrumbs = deriveLastBreadcrumbFromPath(
+      breadcrumbs,
       breadcrumbTitle,
       entityPath,
       replaceLastItem
@@ -60,15 +55,15 @@ const Breadcrumbs = ({
   }
 
   if (constructLcBreadcrumbs) {
-    crumbs = deriveLcBreadcrumbs(
-      breadcrumbData,
+    breadcrumbs = deriveLcBreadcrumbs(
+      breadcrumbs,
       breadcrumbTitle,
       entityPath,
       titleInclude
     )
   }
 
-  const breadcrumbList = transformBreadcrumbs(crumbs)
+  const breadcrumbList = transformBreadcrumbs(breadcrumbs)
   const filteredCrumbs = filterInvalidCrumbs(breadcrumbList)
 
   return (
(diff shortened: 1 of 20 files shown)


src/templates/layouts/newsStory/index.tsx
```diff
@@ -20,9 +21,16 @@ export const NewsStory = ({
   listing,
   lovellVariant,
   lovellSwitchPath,
+  entityPath,
+  breadcrumbs,
 }: LovellExpandedResourceTypeType<NewsStoryType>) => {
   return (
     <>
+      <Breadcrumbs
```
tjheffner, line 29: I think we should include breadcrumbs at the `[[...slug]].tsx` level instead of on every page layout. It will keep things more consistent in the end, imo. 

```
      <Breadcrumbs
        breadcrumbs={resource.breadcrumbs}
        entityPath={resource.entityPath}
        hideHomeBreadcrumb={true}
      />
```

src/types/index.d.ts
```diff
@@ -328,3 +330,21 @@ export type MegaMenuPromoColumn = {
   link: MegaMenuLink
   description: string
 }
+
+/**
+ * Types for breadcrumb data
+ */
+export interface Breadcrumb {
```
tjheffner, line 337: is this the type coming from json:api? if so i'd prefer it inside of `src/types/dataTypes/drupal/field_type.d.ts`

src/types/index.d.ts
```diff
@@ -91,6 +91,7 @@ export type NewsStoryType = PublishedEntity & {
   entityId: number
   entityPath: string
   administration: Administration
+  breadcrumbs: any
```
tjheffner, line 94: I think we know what types these will be (maybe `BreadcrumbItem[]` ?), per the testing snapshot:

```
  "breadcrumbs": Array [
    Object {
      "options": Array [],
      "title": "Home",
      "uri": "https://cms-7m0jwgfd3t0pt3txlgzvixowibrfagsq.ci.cms.va.gov/",
    },
  ],
  ```

File: src/templates/globals/wrapper/index.tsx [hunk 4]
```diff
@@ -65,6 +65,16 @@
     }
   }, [bannerData, showBanners, headerFooterData])
 
+
+
+  function replacer(key, value) {
```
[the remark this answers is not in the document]
tjheffner: this doesn't appear to be used anywhere. also not sure why it's in this file

tjheffner (changes requested):

src/templates/common/breadcrumbs/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+  disableAnalytics,
+  deriveBreadcrumbsFromUrl,
+  replaceLastItem,
+  constructLcBreadcrumbs,
+  titleInclude,
+  hideHomeBreadcrumb,
+  customCrumbHomeText,
+}: BreadcrumbProps) => {
+  const breadcrumbData: BreadcrumbItem[] = breadcrumbs
+  let crumbs = breadcrumbsOverride || breadcrumbData
```
ryguyk, line 36: **Question**
Why do we need `breadcrumbData`? Couldn't we have `let crumbs = breadcrumbsOverride || breadcrumbs` ?

ryguyk: **Question**
More fundamentally, what's the purpose of the override prop in the first place? It seems to me like any desire to override could be accomplished by simply passing the override data in the `breadcrumbs` prop. It doesn't seem like there is any sort of cascading going on between the override and the base prop. Couldn't the caller of this component just pass the override data as _the_ breadcrumb data?

src/templates/layouts/newsStory/index.tsx
jtmst: i considered this, but different templates will need to pass in different flags.  For example, some templates hide the home breadcrumb and others dont.

src/templates/globals/util/breadcrumbUtils.tsx
```diff
@@ -0,0 +1,65 @@
...
+  if (replaceLastItem) {
+    breadcrumbs.splice(breadcrumbs.length - 1, 1, last)
+  } else {
+    breadcrumbs.push(last)
+  }
+
+  return breadcrumbs
+}
+
+export function deriveLcBreadcrumbs(
```
ryguyk, line 24: What is "Lc"? I found myself wondering this a few times when looking through. I wonder if it could be added in a comment somewhere. Or maybe I'm just really blanking on something very obvious 😅

jtmst: you know, i just assumed it was 'learning center' or something else I hadn't worked with yet and breezed past that,  I'm actually not sure.

src/templates/common/breadcrumbs/index.tsx
jtmst: yeah thats a good catch, I was parroting content-build there and didnt consider that we dont need that prop at all.  Will adjust

src/types/index.d.ts
jtmst: missed that on my cleanup, fixed

src/types/index.d.ts
jtmst: Good call, moved

src/templates/layouts/newsStory/index.tsx
tjheffner: I hear you, but I think that logic can still be handled at the slug level.

also coinciding with the changes in the skip link PR, I believe we will want breadcrumbs placed outside of the `<main>` container, like va.gov is doing:
Screenshot 2023-10-12 at 11 11 02 AM

src/templates/globals/util/breadcrumbUtils.tsx
jtmst: for the record it is learning centers, here is the PR it originates from: https://github.com/department-of-veterans-affairs/vets-website/pull/14687

jtmst: Added a comment.  When we add said content types we can re-evaluate if this logic is actually necessary.

jtmst:
New PR created here to circumvent merge conflicts: https://github.com/department-of-veterans-affairs/next-build/pull/216

PR #194, opened 2023-10-12
skip link added

jtmst:
## Description
Relates to #[15512](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15512). (or closes?)

Adds a skip link to all pages, only visible on focus and links to the first h1 or the content div

Changes:
Adds skip link to the wrapper component
Adds a div to the story listing template with a `#content` id, as it didnt have one previously

## Testing done
Local visual testing

## Screenshots
image

## QA steps
- Navigate to a a story listing or news story page
- tab to first element above footer
- verify enter/clicking on the 'skip to content' link sets focus and scrolls to content as you would expect

### DIFF
--- a/src/templates/globals/util/handleSkipLink.js
+++ b/src/templates/globals/util/handleSkipLink.js
@@ -1,0 +1,23 @@
+export const handleSkipLink = (e) => {
+  e.preventDefault()
+  const contentDiv = document.getElementById('content')
+  const firstH1 = contentDiv && contentDiv.querySelector('h1')
+  if (firstH1) {
+    firstH1.setAttribute('tabindex', '-1')
+    firstH1.focus()
+
+    const h1Position = firstH1.getBoundingClientRect().top + window.scrollY
+    window.scrollTo({
+      top: h1Position,
+    })
+
+    // Cleanup
+    firstH1.addEventListener(
+      'blur',
+      () => {
+        firstH1.removeAttribute('tabindex')
+      },
+      { once: true }
+    )
+  }
+}
--- a/src/templates/globals/wrapper/index.tsx
+++ b/src/templates/globals/wrapper/index.tsx
@@ -14,6 +14,7 @@ import { NodeBanner } from '@/types/dataTypes/drupal/node'
 import { BannerDisplayType, BannerTypeMapping } from '@/data/queries/banners'
 import { Header } from '../header'
 import { Footer } from '../footer/index'
+import { handleSkipLink } from '../util/handleSkipLink'
 
 // Allows additions to window object without overwriting global type
 interface customWindow extends Window {
@@ -67,6 +68,9 @@ export function Wrapper({
 
   return (
     <>
+      <a href="#content" onClick={handleSkipLink} className="show-on-focus">
+        Skip to content
+      </a>
       <Header />
       {showBanners ? banners : null}
       <main>{children}</main>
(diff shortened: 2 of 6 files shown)


src/templates/layouts/storyListing/index.tsx
```diff
@@ -51,44 +51,46 @@ export function StoryListing({
...
 
   return (
-    <div key={id} className="usa-grid usa-grid-full">
-      {/* Widget coming from vets-website */}
-      <nav
-        data-template="navigation/facility_sidebar_nav"
-        aria-label="secondary"
-        data-widget-type="side-nav"
-      ></nav>
+    <div id="content" className="interior">
```
tjheffner, line 54: can we put this div inside of `[[...slug]]` so it only needs added in the one place? i do see it present on the newsStory layout already

src/templates/globals/util/handleSkipLink.js
```diff
@@ -0,0 +1,23 @@
+export const handleSkipLink = (e) => {
```
tjheffner, line 1: i think i'd prefer this inside `@lib/utils/` instead of `templates/globals/util` (mostly because I want to nuke this globals directory (see: #191)), but open to reasons why it should stay here

tjheffner (changes requested):
Skip link & focus functionality appears to work as expected for story listing + news story pages, but I do have some comments around structure / location

src/templates/layouts/storyListing/index.tsx
tjheffner: we may want to move `<main>` out of `wrapper.tsx` into `[[...slug]].tsx` also but that doesn't have to be done in this PR

ryguyk, line 54: It feels like we can move this to a more common location so it's not needed in every template.

jtmst: agreed, moved main and content divs to slug

src/templates/globals/util/handleSkipLink.js
jtmst: moved to utils

src/templates/layouts/storyListing/index.tsx
jtmst: Agreed, moved it to the slug component

tjheffner (dismissed):
This feels much cleaner, thank you!

src/templates/globals/wrapper/index.tsx
```diff
@@ -14,6 +14,7 @@ import { NodeBanner } from '@/types/dataTypes/drupal/node'
 import { BannerDisplayType, BannerTypeMapping } from '@/data/queries/banners'
 import { Header } from '../header'
 import { Footer } from '../footer/index'
+import { handleSkipLink } from '../../../lib/utils/handleSkipLink'
```
ryguyk, line 17: **CONSIDER**
Now that this is moved to the general `utils` directory, I think we could consider an absolute path using `@`. I don't think it's necessary (or advised) for the `Header` and `Footer` imports above it, but it seems to make sense here.

ryguyk (approved):
Approved with one item for further consideration.

src/templates/globals/wrapper/index.tsx
jtmst: Good call, fixed

tjheffner (approved):

PR #208, opened 2023-10-20
assorted repo housekeeping things (cypress, env files, etc)

tjheffner:
## Description
- Updates the env file vars provided in-repo
- Updates cypress to use env vars for testing
- Updates the fallback base endpoint to use local cms instead of a nonexistent endpoint (`cms.va.gov` ??)
- Extracts the env handling to a script for DRY across the different config files
- Fixes one a11y error related to banners and aria roles. The a11y tests are disabled because there are some separate errors related to aria roles (the header has `role="banner"` and so do the shadow DOM for other banners (PACT Act, etc). Not sure why this isn't reported on the existing site.
- Adds some new e2e tests for news story + story listing pages

## Testing done
- `yarn test`
- `yarn export` then `yarn export:serve`. in a different terminal window `yarn test:cypress`

## QA steps

```[tasklist]
```


cypress/e2e/newsStory.cy.js
```diff
@@ -0,0 +1,16 @@
+/// <reference types="cypress" />
+
+context('News Story', () => {
+  it('News Story page renders with navigation back to parent story list', () => {
+    cy.visit('/butler-health-care/stories/its-flu-shot-time/')
+
+    cy.get('#news-stories-listing-link').click()
+    cy.location('pathname').should('equal', '/butler-health-care/stories/')
+  })
```
ryguyk, line 9: Just noting here that we'll want specific tests for Lovell pages, as this link is modified in the bifurcation process.

tsconfig.json
```diff
@@ -38,8 +38,9 @@
     "*.{ts|tsx|js|jsx}",
     "scripts/*",
     "test/*",
-    "cypress/*",
-    ".storybook/*"
+    "cypress/**/*",
+    ".storybook/*",
+    "*.ts"
```
ryguyk, line 43: Do we need this `"*.ts"`? Isn't it captured in line 38? (`"*.{ts|tsx|js|jsx}"`)

ryguyk (approved):
Approved with limited context/understanding around some of the pieces. I did add one note and one question.

tsconfig.json
tjheffner: no we don't & you're right it is, leftover weirdness. will remove

cypress/e2e/newsStory.cy.js
tjheffner: yes definitely. I'll stub some of those out soon

jtmst (approved):

PR #216, opened 2023-10-25
15056 new breadcrumbs

jtmst:
## Description
Relates to #[15056](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15056).

Continuation of this PR https://github.com/department-of-veterans-affairs/next-build/pull/193

Changes:
- Adds breadcrumb component to next-builld, mimicking the functionality in content-build but using the VA design system breadcrumbs web-component.
- Modifies queries to include breadcrumbs in resource objects where needed
- Adds unit tests for breadcrumb component
- Adds unit tests for breadcrumb utility functions
- Adds storyboard component
- Adds the component slug file

## Testing done
Local visual testing

## Screenshots

Screen Shot 2023-10-25 at 2 31 34 PM
Screen Shot 2023-10-25 at 2 32 00 PM

## QA steps
- verify that breadcrumbs render on appropriate pages
- verify styles match prod

### DIFF
--- a/src/data/queries/newsStory.ts
+++ b/src/data/queries/newsStory.ts
@@ -61,7 +61,7 @@ export const formatter: QueryFormatter<NodeNewsStory, NewsStoryType> = (
     introText: entity.field_intro_text,
     bodyContent: entity.field_full_story,
     date: entity.created,
-    breadcrumbs: entity.breadcrumbs || null,
+    breadcrumbs: entity.breadcrumbs,
     socialLinks: {
       path: entity.path.alias,
       title: entity.title,
--- a/src/data/queries/newsStory.ts
+++ b/src/data/queries/newsStory.ts
@@ -61,6 +61,7 @@ export const formatter: QueryFormatter<NodeNewsStory, NewsStoryType> = (
     introText: entity.field_intro_text,
     bodyContent: entity.field_full_story,
     date: entity.created,
+    breadcrumbs: entity.breadcrumbs || null,
     socialLinks: {
       path: entity.path.alias,
       title: entity.title,
--- a/src/lib/utils/breadcrumbUtils.tsx
+++ b/src/lib/utils/breadcrumbUtils.tsx
@@ -26,7 +26,7 @@ export function deriveLcBreadcrumbs(
   breadcrumbs: BreadcrumbItem[],
   breadcrumbTitle: string,
   currentPath: string,
-  titleInclude?: boolean
+  lcBreadcrumbsTitleInclude?: boolean
 ): BreadcrumbItem[] {
   const filteredCrumbs: BreadcrumbItem[] = breadcrumbs.filter(
     (crumb) => crumb.uri !== '/resources'
@@ -38,7 +38,7 @@ export function deriveLcBreadcrumbs(
     options: [],
   })
 
-  if (titleInclude) {
+  if (lcBreadcrumbsTitleInclude) {
     filteredCrumbs.push({
       uri: currentPath,
       title: breadcrumbTitle,
@@ -50,17 +50,20 @@ export function deriveLcBreadcrumbs(
 }
 
 export function transformBreadcrumbs(breadcrumbs) {
-  const transformedCrumbs = breadcrumbs.map((crumb) => ({
-    href: crumb.uri,
-    label: crumb.title,
-    options: crumb.options,
-  }))
+  const transformedCrumbs = breadcrumbs.map((crumb) => {
+    const pathname = new URL(crumb.uri, 'http://va.gov').pathname
+    return {
+      href: pathname,
+      label: crumb.title,
+      options: crumb.options,
+    }
+  })
 
   return transformedCrumbs
 }
 
 export function filterInvalidCrumbs(crumbs) {
   return crumbs.filter((crumb, index, arr) => {
-    return crumb.href !== 'internal:#' || index === arr.length - 1
+    return crumb.href !== '' || index === arr.length - 1
   })
 }
--- a/src/lib/utils/breadcrumbs.ts
+++ b/src/lib/utils/breadcrumbs.ts
@@ -1,4 +1,7 @@
-import { BreadcrumbItem } from '@/types/dataTypes/drupal/field_type'
+import {
+  BreadcrumbItem,
+  BreadCrumbLink,
+} from '@/types/dataTypes/drupal/field_type'
 
 export function deriveLastBreadcrumbFromPath(
   breadcrumbs: BreadcrumbItem[],
@@ -53,7 +56,7 @@ function getRelativePathFromURL(uri) {
   return new URL(uri, 'http://va.gov').pathname
 }
 
-export function transformBreadcrumbs(breadcrumbs) {
+export function transformBreadcrumbs(breadcrumbs: BreadcrumbItem[]) {
   return breadcrumbs.map((crumb) => ({
     href: getRelativePathFromURL(crumb.uri),
     label: crumb.title,
@@ -62,7 +65,9 @@ export function transformBreadcrumbs(breadcrumbs) {
 }
 
 // Filters out breadcrumbs with no relavant href, excluding the last which would represent the current page
-export function filterInvalidCrumbs(crumbs) {
+export function filterInvalidCrumbs(
+  crumbs: BreadCrumbLink[]
+): BreadCrumbLink[] {
   return crumbs.filter((crumb, index, arr) => {
     return crumb.href !== '' || index === arr.length - 1
   })
--- a/src/lib/utils/breadcrumbs.ts
+++ b/src/lib/utils/breadcrumbs.ts
@@ -49,19 +49,19 @@ export function deriveLcBreadcrumbs(
   return filteredCrumbs
 }
 
-export function transformBreadcrumbs(breadcrumbs) {
-  const transformedCrumbs = breadcrumbs.map((crumb) => {
-    const pathname = new URL(crumb.uri, 'http://va.gov').pathname
-    return {
-      href: pathname,
-      label: crumb.title,
-      options: crumb.options,
-    }
-  })
-
-  return transformedCrumbs
+function getRelativePathFromURL(uri) {
+  return new URL(uri, 'http://va.gov').pathname
 }
 
+export function transformBreadcrumbs(breadcrumbs) {
+  return breadcrumbs.map((crumb) => ({
+    href: getRelativePathFromURL(crumb.uri),
+    label: crumb.title,
+    options: crumb.options,
+  }))
+}
+
+// Filters out breadcrumbs with no relavant href, excluding the last which would represent the current page
 export function filterInvalidCrumbs(crumbs) {
   return crumbs.filter((crumb, index, arr) => {
     return crumb.href !== '' || index === arr.length - 1
--- a/src/templates/common/breadcrumbs/index.tsx
+++ b/src/templates/common/breadcrumbs/index.tsx
@@ -1,0 +1,94 @@
+import {
+  transformBreadcrumbs,
+  deriveLastBreadcrumbFromPath,
+  deriveLcBreadcrumbs,
+  filterInvalidCrumbs,
+} from '@/templates/globals/util/breadcrumbUtils'
+import { BreadcrumbItem } from '@/types/dataTypes/drupal/field_type'
+
+interface BreadcrumbProps {
+  breadcrumbs?: BreadcrumbItem[]
+  breadcrumbTitle?: string
+  disableAnalytics?: boolean
+  deriveBreadcrumbsFromUrl?: boolean
+  replaceLastItem?: boolean
+  constructLcBreadcrumbs?: boolean
+  titleInclude?: boolean
+  hideHomeBreadcrumb?: boolean
+  customCrumbHomeText?: string
+  entityPath?: string
+}
+const Breadcrumbs = ({
+  breadcrumbs,
+  entityPath,
+  breadcrumbTitle,
+  disableAnalytics,
+  deriveBreadcrumbsFromUrl,
+  replaceLastItem,
+  constructLcBreadcrumbs,
+  titleInclude,
+  hideHomeBreadcrumb,
+  customCrumbHomeText,
+}: BreadcrumbProps) => {
+  if (!breadcrumbs) return null
+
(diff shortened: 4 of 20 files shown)


src/templates/globals/util/breadcrumbUtils.tsx
```diff
@@ -0,0 +1,66 @@
+import { BreadcrumbItem } from '@/types/dataTypes/drupal/field_type'
+
+export function deriveLastBreadcrumbFromPath(
```
tjheffner, line 3: i would move this file & its tests to `@lib/utils/` instead of `@templates/globals/util` because they aren't actually returning templates, only modifying data

src/templates/common/breadcrumbs/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+      entityPath,
+      titleInclude
+    )
+  }
+
+  const breadcrumbList = transformBreadcrumbs(breadcrumbs)
+  const filteredCrumbs = filterInvalidCrumbs(breadcrumbList)
+
+  return (
+    <div>
```
tjheffner, line 70: do we need this wrapper div?

src/types/index.d.ts
```diff
@@ -91,6 +92,7 @@ export type NewsStoryType = PublishedEntity & {
   entityId: number
   entityPath: string
   administration: Administration
+  breadcrumbs?: BreadcrumbItem[]
```
tjheffner, line 95: the cms pr was merged, so it is providing breadcrumbs now on all envs, i think we should remove the `?`'s

https://content-build-medc0xjkxm4jmpzxl3tfbcs7qcddsivh.ci.cms.va.gov/jsonapi/node/news_story

src/data/queries/newsStory.ts
```diff
@@ -61,6 +61,7 @@ export const formatter: QueryFormatter<NodeNewsStory, NewsStoryType> = (
     introText: entity.field_intro_text,
     bodyContent: entity.field_full_story,
     date: entity.created,
+    breadcrumbs: entity.breadcrumbs || null,
```
tjheffner, line 64: let's nuke these `null`'s now that cms is providing breadcrumbs

tjheffner (changes requested):

tjheffner:
One other thing I noticed, the breadcrumbs link to the content-build endpoint instead of this environment's built pages. we may need to bring over some of the DOM/data rewriting functions from content-build to properly link things. If it turns out to be particularly tricky, we can work on that in a follow-up PR if needed.

src/templates/globals/util/breadcrumbUtils.tsx
ryguyk: I think I raised a similar issue on the previous PR, though I can't find it now. I think there's an argument to having utility modules that are specific to a thing residing with that thing. I recently implemented this with `src/drupal/lovell/utils.ts` as a collection of utility functions that are specific to Lovell and that do not fall under the other files (i.e. `constants.ts`, `staticPaths.ts`, etc.). 

But I also think this situation is slightly different. I think there's a strong argument for having the functions currently in this file reside in `@/lib/utils/breadcrumbs.ts` because they are not necessarily specific to this template. Breadcrumb utils feel more general than a single template. If there _were_ a utility function that were somehow only for a single template (something like `formatBreadcrumbForMyVerySpecialUseCase`) then we could perhaps have a `utils` directory within that template directory, but, honestly, in that case I'd probably keep that function in the component file.

In any case, I think we should settle on `utils` rather than `util` since that is what is used elsewhere already.

src/templates/common/breadcrumbs/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+}
+const Breadcrumbs = ({
+  breadcrumbs,
+  entityPath,
+  breadcrumbTitle,
+  disableAnalytics,
+  deriveBreadcrumbsFromUrl,
+  replaceLastItem,
+  constructLcBreadcrumbs,
+  titleInclude,
```
ryguyk, line 29: I feel like this prop could maybe be more clear if it included `lc` somehow (e.g. `lcBreadcrumbsTitleInclude`). As I understand it, this prop is only useful when `constructLcBreadcrumbs` is `true`, but it _feels_ like it's named in a way that might make me (as a consumer) feel like I can set it in all cases.

src/templates/common/breadcrumbs/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+  hideHomeBreadcrumb,
+  customCrumbHomeText,
```
ryguyk, line 31: This relationship is a bit interesting to me. 
### Scenario 1 (default, makes sense)
```
default
```
Home > Path > To > Page

### Scenario 2 (makes sense)
```
hideHomeBreadcrumb: true, 
```
Path > To > Page

### Scenario 3 (kind of makes sense, depending how you look at it)
```
hideHomeBreadcrumb: true, 
customCrumbHomeText: 'MyCustonHomeCrumb',
```
MyCustomHomeCrumb > Path > To > Page

### Scenario 4 (feels very odd to me given current naming/structure)
```
customCrumbHomeText: 'MyCustonHomeCrumb',
```
Home > Path > To > Page

ryguyk (commented):
Just left a couple comments. Haven't gotten all the way through this yet, but this looks very good overall from what I saw.

src/templates/globals/util/breadcrumbUtils.tsx
tjheffner: I would really like them placed in `@lib/utils` solely because I think there's a future where we do a lot of the business-logic-breadcrumb processing inside of the queries files, instead of at the breadcrumb component level.

I'd love that future breadcrumb component to be as dumb as possible, just taking a list of breadcrumbs and displaying them. It should be up to the query to determine what format those breadcrumbs should be delivered in (hidden or custom home crumb, whatever)

src/templates/common/breadcrumbs/index.tsx
jtmst: I considered this a bit as well, and maybe wrongly left it as is.   Looking again, I think its more logical that `if customCrumbHomeText -> set hideHomeBreadcrumb to true` so that the consumer doesnt need to be aware of that relationship at all.

ryguyk: I think this largely makes sense, though I might put it slightly differently. To me, `hideHomeBreadcrumb` should be the prop with higher authority. It should default to `false`, but, if it's set to `true`, there should be no home breadcrumb shown, regardless of whether a custom one is passed in. If it's not set (or explicitly set to `false`), its contents should then fall to whether a custom value is passed in. If so, use the custom value. If not, "Home".

### Default
```
default
```
Home > Path > To > Page

### Hide home
```
hideHomeBreadcrumb: true, 
customCrumbHomeText: 'MyCustonHomeCrumb',  // this is ignored
```
Path > To > Page

### Replace home
```
customCrumbHomeText: 'MyCustonHomeCrumb',
```
MyCustonHomeCrumb > Path > To > Page

src/types/dataTypes/drupal/field_type.d.ts
```diff
@@ -64,3 +64,21 @@ export interface FieldAdministration {
...
+/**
+ * Types for breadcrumb data
+ */
+export interface Breadcrumb {
+  url: {
+    path: string
+    routed: boolean
+  }
+  text: string
+}
```
ryguyk, line 77: Is this type being used?

src/templates/common/breadcrumbs/index.tsx
jtmst: agreed, updated

src/templates/common/breadcrumbs/index.tsx
jtmst: fixed with new if/else logic

src/types/dataTypes/drupal/field_type.d.ts
jtmst: missed this in cleanup, fixed

src/templates/common/breadcrumbs/index.tsx
jtmst: I'm thinking that if customCrumbText is passed in at all, then the user wants to see it displayed.  I can see a case for either, but will keep the updated if/else logic for now and can update later if found confusing

src/templates/common/breadcrumbs/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+      })}
+    </va-breadcrumbs>
+  )
+}
+
+Breadcrumbs.defaultProps = {
+  label: 'Breadcrumbs',
+  uswds: true,
+  string: '',
+}
```
ryguyk, line 88: **SHOULD**

Is this needed? It looks like it is a remnant of copying something as a starting point. Can it be removed?

FWIW, I think it's more idiomatic to define default prop values in the function signature (when destructuring `props`).

src/lib/utils/breadcrumbUtils.tsx
```diff
@@ -0,0 +1,69 @@
+import { BreadcrumbItem } from '@/types/dataTypes/drupal/field_type'
```
ryguyk, line 1: **SHOULD**
Can we rename this to just breadcrumbs.ts?
- Don't need `.tsx`
- Don't need `Utils` (it's redundant given the directory structure)

src/lib/utils/breadcrumbUtils.tsx
```diff
@@ -0,0 +1,69 @@
...
+      options: [],
+    })
+  }
+
+  return filteredCrumbs
+}
+
+export function transformBreadcrumbs(breadcrumbs) {
+  const transformedCrumbs = breadcrumbs.map((crumb) => {
+    const pathname = new URL(crumb.uri, 'http://va.gov').pathname
```
ryguyk, line 54: **CONSIDER**
I wonder if this might be useful as an abstraction of its own. `getRelativePathFromUrl`. We'll likely need this again elsewhere, and, while the code is relatively short as it is, it is a bit esoteric (need to know parameters as well as property to grab). It would maybe be useful to have `@/lib/utils/url` with functions like this.

src/templates/common/breadcrumbs/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+  if (constructLcBreadcrumbs) {
+    breadcrumbs = deriveLcBreadcrumbs(
+      breadcrumbs,
+      breadcrumbTitle,
+      entityPath,
+      lcBreadcrumbsTitleInclude
+    )
+  }
+
+  const breadcrumbList = transformBreadcrumbs(breadcrumbs)
```
ryguyk, line 64: **CONSIDER**
I wonder if this step belongs in the `formatter`.

Also, this step changes the shape of things. Do we want a type definition for a `BreadcrumbLink` or something to that effect?

src/lib/utils/breadcrumbUtils.tsx
```diff
@@ -0,0 +1,69 @@
...
+  })
+
+  return transformedCrumbs
+}
+
+export function filterInvalidCrumbs(crumbs) {
+  return crumbs.filter((crumb, index, arr) => {
+    return crumb.href !== '' || index === arr.length - 1
+  })
+}
```
ryguyk, line 69: **CONSIDER**
A quick comment about what makes a breadcrumb invalid might be useful here.

ryguyk (changes requested):
I marked a few things for consideration. None of them are strictly necessary individually, but I'm marking this as "Request changes" because they should collectively lead to a change or two, I think.

src/templates/common/breadcrumbs/index.tsx
jtmst: Yeah this was leftover from previous iteration, removed

src/lib/utils/breadcrumbUtils.tsx
jtmst: Agreed, fixed

src/lib/utils/breadcrumbUtils.tsx
jtmst: That makes sense, broke it out to its own function

src/lib/utils/breadcrumbUtils.tsx
jtmst: Good call, added

src/templates/common/breadcrumbs/index.tsx
jtmst: Agreed on the type definition, but I think that having that logic happen in one place here instead of every templates formatter is more DRY.  Happy to reconsider if we feel strongly about moving it though

src/lib/utils/breadcrumbs.ts
```diff
@@ -0,0 +1,69 @@
...
+      options: [],
+    })
+  }
+
+  return filteredCrumbs
+}
+
+function getRelativePathFromURL(uri) {
+  return new URL(uri, 'http://va.gov').pathname
+}
```
ryguyk, line 54: This looks great, though I think it will be most useful if it's pulled out (and exported from) a separate file. Maybe `@/lib/utils/url` or something like that.

ryguyk (approved):
Just one comment, but approving. It isn't strictly required.

src/templates/common/breadcrumbs/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+  if (constructLcBreadcrumbs) {
+    breadcrumbs = deriveLcBreadcrumbs(
+      breadcrumbs,
+      breadcrumbTitle,
+      entityPath,
+      lcBreadcrumbsTitleInclude
+    )
+  }
+
+  const breadcrumbList: BreadcrumbItem[] = transformBreadcrumbs(breadcrumbs)
```
ryguyk, line 67: I may have approved too early. I think this `BreadcrumbItems[]` might be in error. `transformBreadcrumbs` looks to return `BreadcrumbLink[]`, though that function doesn't have type definitions.

ryguyk (changes requested):

src/templates/common/breadcrumbs/index.tsx
jtmst: My mistake. <3 typescript

ryguyk (approved):

timcosgrove (approved):
r+

jtmst:
closed to get around change request block

PR #226, opened 2023-10-31
Fix sidenavs not redering for lovell facility pages

jtmst:
## Description
Relates to #[14745](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/14745). (or closes?)

## Testing done
Local visual testing

## Screenshots
Screen Shot 2023-10-31 at 1 54 59 PM
Screen Shot 2023-10-31 at 1 55 27 PM

## QA steps
Navigate to lovell pages (/lovell-federal-health-care-tricare/stories/)
Verify that sidenav is rendering
Verify that links are correct with no duplicates
Verify that non-lovell page sidenavs are unchanged (/butler-health-care/stories/)


src/lib/drupal/lovell/utils.ts
```diff
@@ -133,3 +133,34 @@ export function isLovellBifurcatedResource(
     isLovellFederalResource(resource as LovellBifurcatedFormattedResource)
   )
 }
+
+export function getLovellVariantOfMenu(menu, variant: LovellVariant) {
```
tjheffner, line 137: If it's possible to type `menu` I think that would be helpful for understanding at a glance. It looks like this is recursing through `DrupalMenuLinkContent[]` or a `MenuItem`, is that right?

src/lib/drupal/facilitySideNav.ts
```diff
@@ -8,25 +8,25 @@ const normalizeMenuItem = (item: MenuItem): SideNavItem => {
...
   // Bail early if no tree is provided
   if (!menu.tree || menu.tree.length === 0) return null
 
-  const links = []
+  let links = []
   menu.tree.forEach((item) => {
     return links.push(normalizeMenuItem(item))
   })
-
+  links = links.filter((link) => link.links.length > 0)
```
tjheffner, line 29: I'm curious what items this is filtering out for Lovell compared to existing menus (that didn't seem to need this filtering)

jtmst: the lovell data comes with 2 extra top layer menu objects that are needed for some other purposes, but just muddy up the data for side navs.  This just filters those out

tjheffner: thanks for clarifying, that makes sense. if you could add a short comment to that effect for our future selves, something like `// This filters out two lovell-specific menu items that aren't used in side navs.`

src/lib/drupal/lovell/utils.ts
```diff
@@ -133,3 +133,34 @@ export function isLovellBifurcatedResource(
...
+  }
+
+  // Filter objects
+  const newData = {}
+  for (const key in menu) {
+    const filteredData = getLovellVariantOfMenu(menu[key], variant)
+    if (filteredData !== null) {
+      newData[key] = filteredData
+    }
+  }
```
ryguyk, line 163: I'm not sure about this. Let's sync up and discuss some things.

src/lib/drupal/facilitySideNav.ts
```diff
@@ -8,25 +8,25 @@ const normalizeMenuItem = (item: MenuItem): SideNavItem => {
...
     item.items.forEach((i) => nestedItems.push(normalizeMenuItem(i)))
   }
-
   return {
     description: item.description,
     expanded: item.expanded,
     label: item.title,
     links: nestedItems,
     url: { path: item.url },
+    fieldMenuSection: item.field_menu_section || null,
```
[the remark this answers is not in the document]
tjheffner: completely agree here. `lovellSection` feels appropriate, and clearer for future debuggers

src/lib/drupal/lovell/utils.ts
ryguyk: I understand what's going on here, but it feels a bit weird to me. It took me a bit to wrap my head around why there's "Filter arrays" and "Filter objects". I think that kind of muddies the situation a bit. It doesn't seem like we'd ever need to worry about objects separately, as the recursion should always be on the `links` property. 

The following code might not be perfect, as it depends on some of the other restructuring that we discussed, but it seems to me that this function should look something like this (general idea):

```
  return menu
    .filter(
      (menuItem) =>
        menuItem.fieldMenuSection === variant ||
        menuItem.fieldMenuSection === 'both'
    )
    .map((menuItem) => ({
      ...menuItem,
      links: getLovellVariantOfMenu(menuItem.links, variant),
    }))
```

src/lib/drupal/facilitySideNav.ts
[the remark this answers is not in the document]
jtmst: Makes sense to me, fixed

src/lib/drupal/facilitySideNav.ts
jtmst: Yep, moved that bit to ride along with the util functions, and added this comment for clairty

src/lib/drupal/lovell/utils.ts
jtmst: Good call, fixed

src/lib/drupal/lovell/utils.ts
jtmst: Updated this logic for clarify, thanks for the feedback

src/lib/drupal/lovell/utils.ts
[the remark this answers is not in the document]
jtmst: Going to leave this as is to be addressed in future issue per this discussion: https://dsva.slack.com/archives/C01SR56755H/p1698932592500079

tjheffner (approved):
this feels cleaner, thank you

ryguyk (approved):
Looks good. We'll likely continue to tweak some of this, per our ongoing conversations, but this is a good place to leave this. Nice work!

PR #233, opened 2023-11-06
404 page added

jtmst:
## Description
Relates to #[14711](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/14711)

- Adds the 404 page component
- Adds the common and popular component
- Modifies the getGlobalElements call to allow for only headerData calls
- Adds unit tests for 404 page

## Testing done
Local visual testing

## Screenshots
image
image

## QA steps
Navigate to any non-existant page
Verify that 404 page matches prod


src/pages/404.tsx
```diff
@@ -0,0 +1,71 @@
...
+          </div>
+        </div>
+
+        <CommonAndPopular />
+      </div>
+    </Wrapper>
+  )
+}
+
+export async function getStaticProps() {
```
tjheffner, line 50: I think I would restructure this to match `[[...slug]].tsx` a little closer, unless there's a specific reason that the 404 page needs handled in this way.

`const { headerFooterData } = await getGlobalElements()` feels a bit cleaner than the let & try catch block. slug has the whole getStaticProps function in a try catch

tjheffner (changes requested):

tjheffner (approved):

PR #243, opened 2023-11-16
Vacms 15949 cleanup

ryguyk:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15949

This is a refactor. No functionality is changed. 

Our TypeScript types needed some cleanup and normalization. This PR goes a long way toward accomplishing that. The cleanup uncovered a few silent issues with some types that are also resolved by this PR, but the bulk of the work boils down to implementing this model:

- TypeScript types for data can be thought of as _before_ and _after_ formatting. Types for _before_ formatting are in `types/dataTypes/drupal`. Types for _after_ formatting are in `types/dataTypes/formatted`.
- Pre-formatted types are types specific to Drupal. If we added another data source or replaced Drupal altogether, types for the new data source could augment or replace these types. Those new types would get formatted into the same `formatted` types.
- In this way, our application logic is decoupled from the specifics of the data source. Our business logic can be written to use the normalized/formatted data properties.

This model was already largely implemented, but it wasn't clean or explicit. Additionally, there were a number of places (and still might be a few) where "formatted" data types contained properties that were pre-formatted types. This sort of mixing is ripe for confusion at least, and, more than likely, bugs.

Another thing that is noteworthy is that this PR removes `index.d.ts`. That file held the bulk of the formatted type definitions, but naming it with the `.d.ts` extension was not entirely correct. That extension is generally reserved for defining types for _external_ libraries. Using it here led to some silent type mismatches. The result of the move, as mentioned above, uncovered these previously silent type problems, so this PR addresses those as well.

## Testing done/QA steps
- Ensure existing unit tests still pass
- Ensure build still succeeds.
- A lot of manual QA. This changes a ton of files, and it needs to be tested accordingly.


src/lib/drupal/drupalClient.ts
```diff
@@ -62,10 +62,10 @@ export const drupalClient = new DrupalClient(baseUrl, {
...
   useDefaultResourceTypeEntry: true,
   throwJsonApiErrors: false,
-  auth: {
-    clientId: process.env.DRUPAL_CLIENT_ID,
-    clientSecret: process.env.DRUPAL_CLIENT_SECRET,
-  },
+  // auth: {
+  //   clientId: process.env.DRUPAL_CLIENT_ID,
+  //   clientSecret: process.env.DRUPAL_CLIENT_SECRET,
+  // },
```
tjheffner, line 68: noting we do need these values present for preview functionality to work. but maybe setting it here for everything is the wrong approach. instead we should probably include it during the specific calls for preview routes (eg lines 38-44 in `src/data/queries/newsStory.ts`)

src/lib/drupal/staticPaths.ts
```diff
@@ -55,6 +55,14 @@ export async function getStaticPathsByResourceType(
   )
 
   // Convert the resources to static paths
-  const paths = drupalClient.buildStaticPathsFromResources(modifiedResources)
+  //  TODO: This should probably not use a Drupal-specific utility to work on post-formatted data.
+  //  This is currently using a type assertion to make it possible since we know that the necessary
+  //  props will be present, though they probably shouldn't be present because the formatted data
+  //  prop names are a bit specific to drupal (e.g. `pid` on Path).
```
tjheffner, line 61: i'm not 100% following this this early, `modifiedResources as Parameters<
      typeof drupalClient.buildStaticPathsFromResources
    >[0]` does look a bit funky

src/lib/drupal/drupalClient.ts
ryguyk: I commented this out just to get things to work. Forgot to remove it before pushing up.

src/lib/drupal/staticPaths.ts
ryguyk: This ideally wouldn't be here, as noted in the code comment. More or less, we're just telling the Typescript compiler that the thing is shaped like the object expected by the function to which the thing is being passed.

ryguyk: I'm changing this to implement the conversion without next-drupal. This will go away.

PR #252, opened 2023-11-30
VACMS-15668 cache layer

tjheffner:
## Description
Closes #[15668](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15668).

Adds a cache layer (redis) to our drupalClient for production builds. This means `yarn build` and `yarn export` will both use the cache, `yarn dev` will not.

## Testing done
I ran `time yarn export` on the main branch and against this branch and compared the differences
local no cache: `yarn export  2371.92s user 36.34s system 294% cpu 13:37.13 total`
locally caching header + footer menus: `yarn export  753.89s user 24.78s system 578% cpu 2:14.64 total`

Nearly 11 minutes faster with just those 3 responses cached!

## QA steps
`brew install redis` and then `redis-server` to start redis locally

in a separate terminal window, `yarn dev` and visit any page

in a separate terminal window, `redis-cli --scan --pattern '*'` should return nothing at this point. no cache utilized.

stop `yarn dev` and run `yarn export`. It should complete much faster than it has. When it completes, run `redis-cli --scan --pattern '*'` again and see:
```
"menu:footer-bottom-rail"
"menu:header-megamenu"
"menu:va-gov-footer"
```
You can also type `redis-cli` with no arguments to enter the redis interface itself. `KEYS *` will surface equivalent information (but will be much slower as we add more items to the cache)

```[tasklist]
```


READMEs/caching.md
```diff
@@ -0,0 +1,6 @@
+mac:
+`brew install redis` then `redis-server`
+
+this starts redis on your machine at 127.0.0.1:6379
+
+in a separate window you can now `yarn dev`
```
timcosgrove, line 6: Can we incorporate redis-server into our yarn commands, or does it need to be running prior or persistently across yarn commands?

timcosgrove (commented):
Single question on running redis-server locally, plus, still curious about how this performs on Tugboat.

READMEs/caching.md
ryguyk: Will this work in Linux/WSL?

tjheffner: this readme definitely needs some additional words, i forgot about it. 

I believe redis-server needs to be running prior, it can be a background service, though. I don't believe we want it persistently, only per-build (?)


It should work in Linux/WSL as long as homebrew is available (I believe it is?). Though maybe this is a good point towards running it in a docker container locally

src/lib/utils/redisCache.ts
```diff
@@ -0,0 +1,46 @@
...
+  return redisClient
+}
+
+getCache()
+  .then((connection) => {
+    redisClient = connection
+  })
+  .catch((err) => {
+    console.error({ err }, 'Failed to connect to Redis')
+  })
```
ryguyk, line 32: This feels weird to me. I'm guessing this is being called here so as to jumpstart the redis connection rather than wait until the first attempt to set or get, but it feels like that should be explicitly done by the thing that actually uses it (in this case, the DrupalClient object). This sort of top-level code feels like an anti-pattern.

I think I'm envisioning something like this, though I might be missing something:
```
/*redisCache.ts*/
export async function createRedisCache(url): DataCache {
  const redisClient = createClient({ url })
  redisClient.on('error', (err) => console.error(`Redis Error: ${err}`))
  redisClient.on('connect', () => console.info('Redis connected'))
  redisClient.on('reconnecting', () => console.info('Redis reconnecting'))
  redisClient.on('ready', () => {
    isReady = true
    console.info('Redis ready!')
  })
  
  await redisClient.connect()
  
  return {
    async set(key, value) {
      return await redisClient.set(key, value)
    },
    async get(key) {
      return await redisClient.get(key)
    },
  }
}
```
```
/*drupalClient.ts*/
export const drupalClient = new DrupalClient(baseUrl, {
  fetcher,
  useDefaultResourceTypeEntry: true,
  throwJsonApiErrors: false,
  auth: {
    clientId: process.env.DRUPAL_CLIENT_ID,
    clientSecret: process.env.DRUPAL_CLIENT_SECRET,
  },
  cache: createRedisCache(process.env.REDIS_URL),
  previewSecret: process.env.DRUPAL_PREVIEW_SECRET,
  forceIframeSameSiteCookie: process.env.NODE_ENV === 'development',
})
```

ryguyk (commented):
A few thoughts.

src/lib/utils/redisCache.ts
ryguyk: Though I'm now thinking about the implications of this being async.

ryguyk: Though I'm now realizing there are some implications of this being async.

src/lib/drupal/queryResources.ts
```diff
@@ -0,0 +1,15 @@
...
+  const menu = await drupalClient.getMenu(name, {
+    params: params().getQueryObject(),
+
+    // Cache resource during build, not dev.
+    withCache: process.env.NEXT_PHASE === PHASE_PRODUCTION_BUILD,
+    cacheKey: `menu:${name}`,
+  })
+
+  return menu
+}
```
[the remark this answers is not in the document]
tjheffner: >  I'm not sure that abstraction is necessary, though, because I now realize that there are really only two or three that would exist

This is correct, we really just need them for getMenu and getResource, if we want to abstract it. we can add cachekeys individually to the calls themselves, if we don't want the abstraction.

I do think nextjs's app router does allow caching at a slightly higher level than our fetches wrapped into next-drupal and the cache it can make use of, but this seemed like an appropriate spot for the time being

src/lib/utils/redisCache.ts
```diff
@@ -0,0 +1,44 @@
+/* eslint-disable no-console */
```
ryguyk, line 1: This file looks better, but I still have some questions. Async stuff can be quite confusing, so perhaps I just need to wake up further. See individual questions/comments below, and then some of my code brainstorming at the bottom.

src/lib/utils/redisCache.ts
```diff
@@ -0,0 +1,44 @@
...
+    redisClient.on('error', (err) => console.error(`Redis Error: ${err}`))
+    redisClient.on('connect', () => console.info('Redis connected'))
+    redisClient.on('reconnecting', () => console.info('Redis reconnecting'))
+    redisClient.on('ready', () => {
+      isReady = true
+      console.info('Redis ready!')
+    })
+  }
+
+  await redisClient.connect()
```
ryguyk, line 20: I find I generally have to think about `Promise`/`async`/`await` stuff a million times to chew on it, but I think we can likely do away with the `await` here, as well as the `async` in the function signature. `connect` will return a `Promise`, which, to me, in this case, is more clear than awaiting something within the function itself. I think we can just return the result of the `.connect()` call, which is effectively saying that we're returning something that has attempted to make a connection and will eventually be a connection. Then, we can `await` that when we need to call a method on it in the `redisCache` function below.

This might just be my flavor of reasoning about this, so take it with a grain of salt.

src/lib/utils/redisCache.ts
```diff
@@ -0,0 +1,44 @@
...
+    .then((connection) => {
+      redis = connection
+    })
+    .catch((err) => {
+      console.error({ err }, 'Failed to connect to Redis')
+    })
+
+  return {
+    async set(key, value) {
+      return await (await client).set(key, value)
```
ryguyk, line 38: Do we need to `await` the `.set` call? Since this is `await`ed in the next-drupal code, I think we can just return the promise rather than wait for a resolved promise. It ultimately doesn't matter, but it might make things a bit cleaner here than having a double `await`.

src/lib/utils/redisCache.ts
```diff
@@ -0,0 +1,44 @@
...
+    },
+    async get(key) {
+      return await (await client).get(key)
+    },
+  }
+}
```
ryguyk, line 44: I think this makes sense to me.
```
/* eslint-disable no-console */
import { DataCache } from 'next-drupal'
import { createClient, RedisClientType } from 'redis'

export function createRedisClient(url): Promise<RedisClientType> {
  return createClient({ url })
    .on('error', (err) => console.error(`Redis Error: ${err}`))
    .on('connect', () => console.info('Redis connected'))
    .on('reconnecting', () => console.info('Redis reconnecting'))
    .on('ready', () => console.info('Redis ready!'))
    .connect()
}

export function redisCache(client: Promise<RedisClientType>): DataCache {
  return {
    async set(key, value) {
      return (await client).set(key, value)
    },
    async get(key) {
      return (await client).get(key)
    },
  }
}

```

README.md
```diff
@@ -71,6 +71,18 @@ NEXT_IMAGE_DOMAIN=https://va-gov-cms.ddev.site
 
 Now you can run `yarn dev` and data will be coming from your local CMS environment instead.
 
+## Generating the static site
+
+To generate the static pages for https://va.gov, run `yarn export`. This command will generate static pages for all paths that next-build is aware of.
+
+To use the redis cache during your static build, run the following command before running `yarn export`:
```
timcosgrove, line 78: Would it make sense to just always incorporate redis into export and build commands?

READMEs/caching.md
```diff
@@ -0,0 +1,56 @@
+# Caching
+
+Next-build interacts with several different caching layers during its operations. Some of these layers it can control and others it lives with.
+
+One of the caching layers that next-build does control is [drupalClient's](/src/lib/drupal/drupalClient.ts) cache. For this layer, we use [redis](https://redis.io/).
+
+To use the redis cache during your local static builds, start redis locally using the following:
+
+```sh
```
timcosgrove, line 9: Could this be a yarn command?

READMEs/caching.md
```diff
@@ -0,0 +1,56 @@
...
+
+You can also interact with redis directly inside the container using:
+
+```
+docker exec -it next-redis redis-cli
+```
+
+Once redis has started, running `yarn export` will generate all static pages and cache certain repeated requests.
+
+We intentionally do not use this cache when running `yarn dev` (TODO: or `yarn build:preview`) so that every page
```
timcosgrove, line 23: Noted, thanks for the clarity here.

timcosgrove (commented):
Some comments/questions

README.md
tjheffner: i'm chewing on this still. I think it might be better to surface enable it via env var or command line through the options handling mechanism

timcosgrove: I guess put another way...leaving aside ISR scenarios for the moment, and understanding that we don't want redis running during `yarn dev`, could we sandwich `yarn export/build`'s internal commands with

1. stand up redis
2. run the actual export/build stuff
3. tear down redis

And have a flag to disable redis for when we actually want that for a build.

I'd be super happy to have that be a follow-on ticket also. This is just 'what if...' kind of thinking.

src/lib/utils/redisCache.ts
```diff
@@ -0,0 +1,33 @@
...
+    .on('reconnecting', () => console.info('Redis reconnecting'))
+    .on('ready', () => console.info('Redis ready!'))
+    .connect()
+}
+
+export function redisCache(
+  client: Promise<RedisClientType<RedisModules, RedisFunctions, RedisScripts>>
+): DataCache {
+  return {
+    async set(key, value) {
```
timcosgrove, line 26: Could we set a default TTL here and a mechanism for override via flag/option? 

Very much could be follow-on work.

tjheffner: yes, we could set a default TTL. if we are only caching at build-time and doing a teardown of redis after, it may not be necessary, though.

it would be something like:
```ts
    async set(key, value) {
      return (await client).get(key, { 
        EX: option.REDIS_EXPIRE_TIME 
      })
    },
```

README.md
tjheffner: yeah, i think that sounds reasonable. would be a good follow-up

ryguyk (approved):

PR #264, opened 2023-12-04
Add events query/formatter/layout

jtmst:
## Description
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15908

Adds Events to next-build.

Changes:
- Adds events query and formatter
- formatter and query tests and mock data
- Event layout
- Event utl functions for date calculations
- Unit tests
- Google maps component
- Adjusts social links component
- Adds to storybook

## Testing done
Local testing, unit tests

## Screenshots
image

## QA steps
Verify that event nodes (ie http://localhost:3000/central-iowa-health-care/events/52265/) render as expected
Verify that events data is properly brought in


src/templates/layouts/event/util.tsx
```diff
@@ -0,0 +1,72 @@
+import _ from 'lodash'
```
tjheffner, line 1: i would move all of these date time utils to our existing `lib/utils/date.ts` file.

src/templates/common/socialLinks/index.tsx
```diff
@@ -1,28 +1,117 @@
...
-export const SocialLinks = ({ path, title }: SocialLinksProps): JSX.Element => {
+export const SocialLinks = ({
+  path,
+  title,
+  isNews = false,
+  description,
+  address,
+  dateObject,
+}: SocialLinksProps): JSX.Element => {
+  if (isNews) {
```
tjheffner, line 42: this feels a little funky to me. seems like `address`, `description`, `dateObject` are actually the optional values here (versus adding a boolean flag), and the additional item renders if they are present. Something like:

```
{description && address && dateObject && (
    <li className="vads-u-margin-bottom--2p5">
          <a
            data-description={description}
            data-end={dateObject.endValue}
            data-location={`${address.addressLine1} ${address.locality}, ${address.administrativeArea}`}
            data-start={dateObject.value}
            data-subject={title}
            href={path}
            id="add-to-calendar-link"
          >
            <i
              className="va-c-social-icon fas fa-calendar-check vads-u-margin-right--0p5"
              aria-hidden="true"
              role="presentation"
            ></i>
            Add to Calendar
          </a>
        </li>)}
 ```

File: src/templates/layouts/event/index.tsx [hunk 3]
```diff
@@ -0,0 +1,267 @@
+import React, { useState } from 'react'
+import { NodeEvent } from '@/types/dataTypes/drupal/node'
```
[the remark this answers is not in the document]
tjheffner: this appears to be correctly reported as unused, using the formatted version instead

src/templates/layouts/event/index.tsx
```diff
@@ -0,0 +1,268 @@
...
+}: FormattedEvent) => {
+  const [showRecurringEvents, setShowRecurringEvents] = useState(false)
+
+  const toggleRecurringEvents = () => {
+    setShowRecurringEvents((prevState) => !prevState)
+  }
+  const formattedDates = formatDateObject(datetimeRange)
+
+  const mostRecentDate = deriveMostRecentDate(formattedDates)
+  const formattedTimestamp = deriveFormattedTimestamp(mostRecentDate)
```
tjheffner, line 42: I wonder if these should be handled inside the query formatter itself. fine here for now, just a thought

tjheffner (changes requested):

src/templates/common/socialLinks/index.tsx
jtmst: I went this route as it felt easier to deal with the different styling and classes used for the news vs non-news version.  Do you think it would be better to make them just be different components all together?

tjheffner: looking at content-build, there are at least 4 different social share templates :sob:

I think let's make them different components for now, and see if we can streamline that logic in the future for next-build

src/lib/constants/pageSizes.ts
```diff
@@ -5,5 +5,6 @@ import {
 
 export const PAGE_SIZES = {
   [RESOURCE_TYPES.STORY_LISTING]: 10,
+  [RESOURCE_TYPES.EVENT]: 10,
```
tjheffner, line 8: this will be for EVENT_LISTING instead of EVENT

src/templates/layouts/event/event.stories.tsx
```diff
@@ -0,0 +1,926 @@
...
+
+export default {
+  title: 'Layouts/Event',
+  component: Event,
+} as ComponentMeta<typeof Event>
+
+const Template: ComponentStory<typeof Event> = (args) => <Event {...args} />
+
+export const EventExample = Template.bind({})
+EventExample.args = {
```
tjheffner, line 13: this feels like it's probably too many args, very likely doesn't need all of this data in order to render (like the nested breadcrumbs on facilityLocation, etc). not a blocker, we can clean it up later... i think most stories need some cleanup

src/types/dataTypes/drupal/node.ts
```diff
@@ -267,6 +270,136 @@ export interface NodePersonProfile extends DrupalNode {
...
+  timezone: string
+}
+
+export interface urlOfOnlineEvent {
+  uri: string
+  title: string
+  options: unknown
+}
+
+export interface fieldEventListing {
```
tjheffner, line 343: this type should probably be a NodeEventListing, similar to how NodeNewsStory & NodeStoryListing. If it's extending DrupalNode, a lot of those values won't need to be repeated here (`type`, `id`, etc)

src/types/dataTypes/drupal/node.ts
```diff
@@ -267,6 +270,136 @@ export interface NodePersonProfile extends DrupalNode {
...
+    drupal_internal__target_id: number
+  }
+  node_type: {
+    type: string
+    id: string
+    resourceIdObjMeta: unknown
+  }
+}
+
+export interface FacilityLocation extends DrupalNode {
```
tjheffner, line 377: similarly, I think there may be some duplicated types here. We probably want to be using `NodeHealthCareLocalFacility` (line 130ish). it appears to come through the included json response on events as: `node--health_care_local_facility`

tjheffner (changes requested):

tjheffner (approved):
wonderful work. there are a few small follow-up tasks, but those can be handled in a separate ticket & pr. this does a lot to get all of the event-related wiring in place and will be easy to iterate on.

the main follow up tasks i see:
- event breadcrumbs are different from va.gov (home is hidden for us)
- the "see more times" for recurring events shows all immediately, instead of initially trimming to 5
- there appears to be some spacing differences at the bottom, but i like ours better.
- some of the event field types still don't feel DRY to me. we can potentially re-use some things from `paragraph.ts` or `field_types.d.ts`
Screenshot 2023-12-12 at 1 54 53 PM
Screenshot 2023-12-12 at 1 55 19 PM

PR #269, opened 2023-12-05
Vacms 16105 metatags

ryguyk:
## Description
Relates to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/16105

Custom meta tags were made available via JSON:API in recent work. This PR places those custom meta tags, along with other global meta tags, on the page.

## Testing done
Manual; see QA steps

## QA steps
- Visit a number of different page types and ensure meta tags are in the page markup
  - [ ] News story
  - [ ] Lovell news story
  - [ ] Story listing
  - [ ] Lovell story listing

### DIFF
--- a/src/data/queries/newsStory.ts
+++ b/src/data/queries/newsStory.ts
@@ -1,14 +1,9 @@
-import {
-  QueryData,
-  QueryFormatter,
-  QueryOpts,
-  QueryParams,
-} from 'next-drupal-query'
+import { QueryData, QueryFormatter, QueryParams } from 'next-drupal-query'
 import { drupalClient } from '@/lib/drupal/drupalClient'
 import { queries } from '.'
 import { NodeNewsStory } from '@/types/dataTypes/drupal/node'
 import { NewsStory } from '@/types/dataTypes/formatted/newsStory'
-import { GetServerSidePropsContext } from 'next'
+import { ExpandedStaticPropsContext } from '@/lib/drupal/staticProps'
 
 // Define the query params for fetching node--news_story.
 export const params: QueryParams<null> = () => {
@@ -24,10 +19,10 @@ export const params: QueryParams<null> = () => {
 }
 
 // Define the option types for the data loader.
-export type NewsStoryDataOpts = QueryOpts<{
+export type NewsStoryDataOpts = {
   id: string
-  context?: GetServerSidePropsContext
-}>
+  context?: ExpandedStaticPropsContext
+}
 
 // Implement the data loader.
 export const data: QueryData<NewsStoryDataOpts, NodeNewsStory> = async (
@@ -64,6 +59,7 @@ export const formatter: QueryFormatter<NodeNewsStory, NewsStory> = (
     type: entity.type,
     published: entity.status,
     title: entity.title,
+    metatags: entity.metatag,
     image: queries.formatData('media--image', {
       entity: entity.field_media,
       cropType: '2_1_large',
--- a/src/lib/utils/environment.ts
+++ b/src/lib/utils/environment.ts
@@ -1,0 +1,20 @@
+import { hosts } from '@/lib/constants/environment'
+
+export const generateAbsoluteUrlFromBuildType = (
+  relativeUrl: string,
+  buildType: string
+) => {
+  const host = hosts[buildType]
+  if (!host) {
+    return relativeUrl
+  }
+  return relativeUrl.charAt(0) === '/'
+    ? `${host.origin}${relativeUrl}`
+    : `${host.origin}/${relativeUrl}`
+}
+
+export const generateAbsoluteUrlFromEnv = (relativeUrl: string) =>
+  generateAbsoluteUrlFromBuildType(
+    relativeUrl,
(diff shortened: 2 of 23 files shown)


src/data/queries/newsStory.ts
```diff
@@ -24,10 +19,10 @@ export const params: QueryParams<null> = () => {
 }
 
 // Define the option types for the data loader.
-export type NewsStoryDataOpts = QueryOpts<{
+export type NewsStoryDataOpts = {
   id: string
-  context?: GetServerSidePropsContext
```
ryguyk, line 29: The `QueryOpts` type definition provided by next-drupal doesn't seem correct to me.

```
export declare type QueryOpts<Options> = Options & {
    context?: GetStaticPathsContext | GetServerSidePropsContext;
};
```
It seems like it should be `context?: GetStaticPropsContext | GetServerSidePropsContext`. After all, we are querying data from within `getStaticProps` and, as such, passing in a `GetStaticPropsContext` property. So, our query functions need to accept a context that is of type `GetStaticPropsContext`.

We have a lot of instances of `QueryOpts<{some type definition}>`, but, for now, I simply removed this from stories and story listings and explicitly added the context property with the type we want (`ExpandedStaticPropsContext`, which is a decorated form of `GetStaticPropsContext`).  We'll need to address the other instances later. Perhaps we need to ask about/contribute to next-drupal re: `QueryOpts` type.

src/lib/utils/environment.ts
```diff
@@ -0,0 +1,20 @@
...
+  }
+  return relativeUrl.charAt(0) === '/'
+    ? `${host.origin}${relativeUrl}`
+    : `${host.origin}/${relativeUrl}`
+}
+
+export const generateAbsoluteUrlFromEnv = (relativeUrl: string) =>
+  generateAbsoluteUrlFromBuildType(
+    relativeUrl,
+    process.env.NEXT_PUBLIC_BUILD_TYPE
```
tjheffner, line 19: i think we may want to use `APP_ENV` over `NEXT_PUBLIC_BUILD_TYPE`

src/templates/globals/meta/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+      <Head>
+        {noIndex && <meta name="robots" content="noindex" />}
+        <meta charSet="utf-8" />
+        <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+        <meta name="HandheldFriendly" content="True" />
+        <meta name="MobileOptimized" content="320" />
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <link
+          rel="canonical"
+          href={generateAbsoluteUrlFromEnv(canonicalLink)}
```
tjheffner, line 114: I'm not sure that `generateAbsoluteUrlFromEnv()` is working as intended

Screenshot 2023-12-04 at 8 29 55 PM
Both `og:url` and the `<link rel="canonical" href="...">` seem to be relative only, which returns if the build type is not valid (currently `NEXT_PUBLIC_BUILD_TYPE`)

ryguyk: This relates to [the discussion in Slack](https://dsva.slack.com/archives/C01SR56755H/p1701743739547139?thread_ts=1701742110.519999&cid=C01SR56755H). 

This will not work right now. We need to do some work to nail down what different ENV vars mean, and how we are going to use them. 

This was an initial pass, but I'm now thinking we should not be using `NEXT_PUBLIC_BUILD_TYPE` or `NEXT_PHASE` to generate absolute URLs. A dev vs. prod build says very little about _where_ the build will ultimately live. We can do a production build locally, and we should still be able to generate local URLs.

src/lib/utils/environment.ts
ryguyk: See [my other comment](https://github.com/department-of-veterans-affairs/next-build/pull/269#discussion_r1415880770). I'm not sure either of these is ultimately what we want. I think this is a bit deeper. The rhetorical question I would ask is, what is the value of `APP_ENV` on Tugboat?

tjheffner: it's `tugboat`, so naively I thought something like:

```
  TUGBOAT: 'tugboat',

  [environments.TUGBOAT]: {
    hostname: process.env.TUGBOAT_DEFAULT_SERVICE_URL_HOST,
    origin: process.env.TUGBOAT_DEFAULT_SERVICE_URL,
  },
```

src/templates/globals/meta/index.tsx
tjheffner: yes, completely agree

I think ultimately we will need some separate options handling from the .env files, something like: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15322 (analogous to the cli options available to content build https://github.com/department-of-veterans-affairs/content-build/blob/main/src/site/stages/build/options.js )

src/lib/utils/environment.ts
ryguyk: This is clarifying. I need to chew on things still, but this helps paint the picture of your vision.

tjheffner (approved):

PR #295, opened 2023-12-18
Run playwright in a container

olivereri:
## Description
Relates to #[TICKET_NUMBER](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/TICKET_NUMBER). (or closes?)

## QA steps

```[tasklist]
```


tjheffner (approved):

.github/workflows/playwright.yml
```diff
@@ -5,9 +5,16 @@ jobs:
...
-      NODE_EXTRA_CA_CERTS: /etc/ssl/certs/ca-certificates.crt
+    container:
+      image: public.ecr.aws/docker/library/node:18.19.0-bullseye
+      # Certs added for the self hosted runner
+      env:
+        NODE_EXTRA_CA_CERTS: /etc/ssl/certs/ca-certificates.crt
+      ports:
+        - 80
+      volumes:
+        - my_docker_volume:/home/runner
```
tjheffner, line 16: where does this volume dir come from?

.github/workflows/playwright.yml
```diff
@@ -5,9 +5,16 @@ jobs:
   test:
     timeout-minutes: 15
     runs-on: [self-hosted, asg]
-    # Certs added for the self hosted runner
-    env:
-      NODE_EXTRA_CA_CERTS: /etc/ssl/certs/ca-certificates.crt
+    container:
+      image: public.ecr.aws/docker/library/node:18.19.0-bullseye
```
tjheffner, line 9: just for my clarity is this a public aws image or something from the va?

olivereri: These are all from AWS's public repository. We could create our own and store them in Private ECR repos too. Image pulls are pretty fast from here since I think they pull across AWS's network backbone and not the TIC.

.github/workflows/playwright.yml
olivereri: I actually pulled this from another example. I think this can be safely removed.

.github/workflows/playwright.yml
tjheffner: if we can pull from here & still install our own system deps (checks passed, looks like we can), i'm fine with this. we are still using the self hosted runner underneath it all

PR #305, opened 2023-12-26
VACMS-10355 component generator

tjheffner:
## Description
Closes #[10355](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/10355).

- Adds `plop` as a dependency
- Adds handlebars templates for generating things
- Adds 3 generators. One for new basic react components, one for a new data query, and one for a new content type (basically combines the first two generators + a playwright test file)
- Removed the `dataType` folder in `src/types/`. (`src/types/dataTypes/drupal/node.ts => src/types/drupal/node.ts`) This is where the bulk of the changed files come from, sorry 🙈 

## Testing done
`yarn plop`, tried all 3 generators. tried different cases of text input (`VBA Facility`, `vba facility`, `vba_facility`, etc.) and ensured all files generated as expected.

Running `yarn storybook` should show a component that is newly generated without error.

## Screenshots
Screenshot 2023-12-26 at 1 45 41 PM
Screenshot 2023-12-26 at 1 46 12 PM
Screenshot 2023-12-26 at 1 46 17 PM
Screenshot 2023-12-26 at 1 46 24 PM

## QA steps
```[tasklist]
- [ ] Automated tests have passed
- [ ] Tugboat environment was generated without errors
- [ ] Locally, all 3 component generators work without errors. You may need to update types manually after generating.
```

PR #306, opened 2023-12-27
Base field formatter

tjheffner:
## Description
adds a helper function for formatting fields that will present on just about every content type. cleans up associated types.
also reorganizes utils for queries slightly. refactors the fetch for entity / preview mode to a consistent handler.

## Testing done
`yarn test:types`, verified every page type still loaded as expected (story / story list / event)
verified that pages still render normally and when viewing from drupal preview

## Screenshots
n/a

## QA steps

```[tasklist]
```

Not blocked by #305, but the generator file should be updated to use this helper


src/lib/drupal/drupalClient.ts
```diff
@@ -16,3 +17,16 @@ export const drupalClient = new DrupalClient(baseUrl, {
...
+  const menu = await drupalClient.getMenu(name, {
+    params: params().getQueryObject(),
+
+    // Cache resource during build, not dev.
+    // withCache: process.env.NEXT_PHASE === PHASE_PRODUCTION_BUILD,
+    // cacheKey: `menu:${name}`,
+  })
+
+  return menu
+}
```
ryguyk, line 32: This move is interesting to me. Why `drupalClient`? It seems like the job of this file is to create an instance of `DrupalClient` which can then be used elsewhere. The previous structure made more sense to me, at least naively. `getMenu` feels like a drupal utility that belongs in `src/lib/drupal/[menu|query|etc].ts` OR in `src/data/queries/menu.ts`. Maybe I'm thinking of this incorrectly, though.

ryguyk (approved):
Just one comment. Curious about the file reorganization in the one place. Approving and leaving that resolution up to you. Looks like some really good refactoring overall.

src/lib/drupal/drupalClient.ts
tjheffner: no, i think you are thinking of it correctly and i was maybe a little overzealous moving stuff here. I was putting it in drupalClient because it's essentially a wrapper around `drupalClient.getMenu`, but it does make sense to leave `drupalClient.ts` as solely creating the DrupalClient instance, and `getMenu` is a utility that belongs elsewhere.

ryguyk (approved):

PR #308, opened 2023-12-29
Redis fixin' + some small cleanup

tjheffner:
## Description
Relates to #283, where I disabled redis because it was causing CMS tugboat preview servers to fail, as no redis service is available there (nor does it need to be, all preview pages should be fetched fresh). This re-enables redis using an env var. I pass it to the cli via `yarn export` only, when static pages are generated.

I also updated the generator template + terminal output to better reflect the slight refactor to queries in #306 .

## Testing done
Verified redis did not run during `yarn dev`, verified it does run & store menu keys during `yarn export`

## Screenshots
n/a

## QA steps
```[tasklist]
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "build:preview": "yarn build:env-loader && node ./scripts/yarn/build:preview.js",
     "build:sitemap": "yarn build:env-loader & node ./scripts/yarn/build:sitemap.js",
     "start": "yarn build:env-loader && node ./scripts/yarn/start.js",
-    "export": "yarn build:env-loader && node ./scripts/yarn/export.js --USE_REDIS=true",
+    "export": "yarn build:env-loader && node ./scripts/yarn/export.js --USE_REDIS",
     "export:serve": "http-server out -p 8001",
     "redis": "docker run --name next-redis -p 6379:6379 -d redis",
     "redis:cli": "docker exec -it next-redis redis-cli",
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "build:preview": "yarn build:env-loader && node ./scripts/yarn/build:preview.js",
     "build:sitemap": "yarn build:env-loader & node ./scripts/yarn/build:sitemap.js",
     "start": "yarn build:env-loader && node ./scripts/yarn/start.js",
-    "export": "yarn build:env-loader && node ./scripts/yarn/export.js",
+    "export": "yarn build:env-loader && node ./scripts/yarn/export.js --USE_REDIS=true",
     "export:serve": "http-server out -p 8001",
     "redis": "docker run --name next-redis -p 6379:6379 -d redis",
     "redis:cli": "docker exec -it next-redis redis-cli",
--- a/packages/env-loader/src/cli-options.ts
+++ b/packages/env-loader/src/cli-options.ts
@@ -56,6 +56,8 @@ const configureAdditionalHelpText = (
 
 /**
  * Parses CLI options from the command line into an object.
+ *
+ * See https://www.npmjs.com/package/commander#common-option-types-boolean-and-value
  */
 export const getCliOptions = (scriptName: string): EnvVars => {
   const program = new Command()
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
     "build:preview": "yarn build:env-loader && node ./scripts/yarn/build:preview.js",
     "build:sitemap": "yarn build:env-loader & node ./scripts/yarn/build:sitemap.js",
     "start": "yarn build:env-loader && node ./scripts/yarn/start.js",
-    "export": "yarn build:env-loader && node ./scripts/yarn/export.js",
+    "export": "yarn build:env-loader && node ./scripts/yarn/export.js --USE_REDIS=true",
```
ryguyk, line 16: This makes sense. I think I like it.

Notably, it's inconsistent with the way we're doing some other things like the following example:
```
"build": "yarn build:env-loader && node ./scripts/yarn/build.js",
```
```
const { processEnv } = require('env-loader')

processEnv('SSG=true next build')
```

I think I like this better, though, and am likely going to push up a PR to bring the above example in line with this approach because I think it's useful to be able to set these commands via CLI. I have, for example, needed to comment out the check for SSG on `yarn dev` because I wanted to test some things. I think running `yarn dev --SSG=false` or `yarn dev --no-SSG` would be nice.  See my other comment on the syntax question I just alluded to.

packages/env-loader/src/cli-options.ts
```diff
@@ -78,6 +78,7 @@ export const getCliOptions = (scriptName: string): EnvVars => {
     .option('--DRUPAL_CLIENT_SECRET <secret>', 'Drupal client secret')
     .option('--DRUPAL_PREVIEW_SECRET <secret>', 'Drupal preview secret')
     .option('--DRUPAL_SITE_ID <id>', 'Drupal site ID')
+    .option('--USE_REDIS <boolean>', 'Enable redis')
```
ryguyk, line 81: Do we want `--USE_REDIS=true` (`--USE_REDIS=false`) or `--USE_REDIS` (`--no-USE_REDIS`)? One actually uses boolean values whereas the other uses string values. I don't love the `no-USE_REDIS` syntax because it mixes kabob and snake case, but that's a product of us using our UPPER_CASE snake formatting directly on the CLI rather than having a conversion. I don't think the conversion is necessary, ultimately, but this case would feel prone to accidently using `--no-USE-REDIS`, perhaps.

src/lib/drupal/drupalClient.ts
```diff
@@ -13,6 +13,9 @@ export const drupalClient = new DrupalClient(baseUrl, {
     clientId: process.env.DRUPAL_CLIENT_ID,
     clientSecret: process.env.DRUPAL_CLIENT_SECRET,
   },
-  // cache: redisCache(createRedisClient(process.env.REDIS_URL)),
+  // Generally use cache for `yarn export` as it generates all pages
+  cache: process.env.USE_REDIS
```
ryguyk, line 17: Depending on what we decide with boolean vs. string, this might need to change. `process.env.USE_REDIS` will evaluate to true in the case of strings even if `USE_REDIS=false`. We'd need to check `process.env.USE_REDIS === 'true'`, which I don't particularly love, but it is how things are being done with `SSG` currently (which I think I want to change, as mentioned).

ryguyk (commented):
This looks good. I'm not yet approving because of open questions I raised for discussion, but this is very close.

packages/env-loader/src/cli-options.ts
tjheffner: I think I lean towards booleans here, but would we even need `--no-USE_REDIS`? i would expect `--USE_REDIS` = true, and the absence of that flag = false (same for `--SSG`, if it were bool)

ryguyk: Yeah, fair point. I didn't mean to suggest that we'd actually _need_ it, but rather that the utility provides that syntax OOB for boolean values. It probably would only be necessary for a value that is overriding an otherwise true value (from .env file, likely). That is n/a here.

ryguyk (approved):

PR #323, opened 2024-01-05
Some tweaks to CLI configuration

ryguyk:
## Description
- Configures CLI to pass arguments through to the underlying utility.
- Updates CLI help text to explain how to pass args through to underlying utility.
- Displays help text of underlying utility (in addition to general env help) if `-h` is passed to the yarn script.

These changes allowed the removal of a few yarn scripts:
- `lint:staged:test-modified-files`: no longer need this because we can pass the necessary commands through to `yarn test` now and we couldn't before. (Note: before the creation of this now-deleted file, we were simply running `jest`, but this yarn script was created so as to be able to run the env-loader before running jest. We can now accomplish both by running `yarn test` and passing arguments through to jest.)
- `test:watch`: can now run `yarn test -- --watch`
- `test:update-snapshots`: can now run `yarn test -- -u` (Note: We could keep `test:watch` and `test:update-snapshots` if we wanted that layer of abstraction, but I'm not sure it's necessary.)

## Testing done
Manual testing of yarn scripts

## QA steps
- [ ] Run yarn scripts and ensure args can be passed through to underlying utility (e.g. `jest`, `next`)
- [x] Run yarn scripts with `-h` and see both general env help _and_ underlying-utility help (e.g. `yarn test -h`)


packages/env-loader/src/index.ts
```diff
@@ -8,53 +8,36 @@ export type EnvVars = {
...
-  // CMS FEATURE FLAGS
-  const drupalBaseUrlProp = 'NEXT_PUBLIC_DRUPAL_BASE_URL'
-  const drupalBaseUrl =
-    cliOptions[drupalBaseUrlProp] || envVars[drupalBaseUrlProp]
-  const cmsFeatureFlags = await getCmsFeatureFlags(drupalBaseUrl)
-
-  // Return all options with proper cascading:
-  // 1. CLI options have highest precedence
-  // 2. CMS feature flags next
-  // 3. ENV file vars last
```
tjheffner, line 36: I think this comment (or something to that effect) is valuable to keep even if it is technically self-documented in the code

packages/env-loader/src/index.ts
```diff
@@ -8,53 +8,36 @@ export type EnvVars = {
...
-  // For now, don't fetch CMS feature flags for tests.
-  // Will fail CI.
+  // CMS FEATURE FLAGS
+  let cmsFeatureFlags
   if (process.env.APP_ENV === 'test') {
-    return {
-      ...envVars,
-      ...cliOptions,
-    }
+    // For now, don't fetch CMS feature flags for tests. Will fail CI.
```
tjheffner, line 21: not a blocker for this, just curious what failures you were seeing? is it related to fetching an env behind the VA network through GHA (gh-provided vs self-hosted runners), or something else?

tjheffner (approved):
pulled and tested locally, works great. thank you!

I re-ordered some of the `test:` commands in package.json and updated `test:watch` to use the passthrough syntax.

PR #335, opened 2024-01-11
VACMS-16638: Static-Data-File Generation (KISS)

ryguyk:
## Description
Relates to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/16638

This PR implements a framework for including arbitrary static json files in the output of next-build, and making them available at an accessible URL. This framework largely mimics a feature that currently exists in content-build. Since its inception there, it has colloquially been referred to as KISS (Keep It Simply Static).

This PR also configures the generation of one specific static file (vamc-ehr). It's worth noting that the output of this (in the current state of this PR) is a file that has data shaped as though it was produced by GraphQL. This is due to the fact that the current file being generated from content-build outputs data in this format and vets-website code currently expects it to be shaped this way. This PR introduces some commentary about how we might be able to move forward from that by moving some of the pre-processing currently done in vets-website to some additional formatting on the next-build (and, while it exists, content-build) side.

## Testing done/QA steps
- Run `yarn export`
- [ ] Ensure `data/cms/vamc-ehr.json` is included in the output
- [ ] Ensure all other pages are included in the output per usual


src/data/queries/vamcEhr.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export const formatter: QueryFormatter<DrupalVamcEhr[], VamcEhrGraphQLMimic> = (
+  entities: DrupalVamcEhr[]
+) => {
+  // For now, return data formatted as it is in content-build (mimic GraphQL output).
+  // In future, we should move the formatting from the preProcess in vets-website
+  // into this formatter and, while it exists, into postProcess in content-build.
+  // This change will require a coordinated effort so as to not break things with regard
+  // to what vets-website is expecting and what is present in the generated file.
```
tjheffner, line 45: completely agree with this callout, but i am kind of thinking the "future" state should not happen until w/e page type(s) that use this data have been ported to next-build. avoids the most potential drift and eases the coordinated effort back to 2 repos.

src/pages/data/cms/[filename].tsx
```diff
@@ -0,0 +1,81 @@
...
+    const directoryPath = path.dirname(filePath)
+    if (!fs.existsSync(directoryPath)) {
+      fs.mkdirSync(directoryPath, {
+        recursive: true,
+      })
+    }
+    fs.writeFileSync(filePath, JSON.stringify(data))
+  }
+  return {
+    notFound: true,
```
Comment (tjheffner) [inline src/pages/data/cms/[filename].tsx:79]: Noting that this 404 only triggers if someone visits `SITE_URL/data/cms/[filename]`. The json file path `public/data/cms/[filename].json` resolves on the web at `SITE_URL/data/cms/[filename].json`.

tjheffner (approved):
LGTM. I appreciate all of the added comments, made it easy to follow. Existing pages seem to render without issue. Nice work.

PR #336, opened 2024-01-12
VACMS-16476: production tag workflow

timcosgrove:
## Description
Relates to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/16476

## Testing done
This was tested by running the workflow against the Next Build repo to generate tags. There were a number of both unsuccessful and successful runs, all of which can be seen here: https://github.com/department-of-veterans-affairs/next-build/actions/workflows/production-tag.yml

The resulting tags are here: https://github.com/department-of-veterans-affairs/next-build/tags
Every tag after the first one was created by this workflow.

Both success and failure are being reported into this Slack channel: https://dsva.slack.com/archives/C06DSBT7CBW

## QA steps
There is not a good way to test this without enabling the workflow for Pull Request events; this is how the testing/development was done in the first place. I would recommend reviewing the workflow runs to see how the workflow played out. 

There are also explanatory comments on the file itself.

```[tasklist]
- [x] Automated tests have passed
- [x] Tugboat environment was generated without errors
```


.github/workflows/production-tag.yml
```diff
@@ -0,0 +1,131 @@
...
+    steps:
+      - name: Cancel workflow due to DSVA schedule
+        if: ${{ github.event_name == 'schedule' && env.DSVA_SCHEDULE_ENABLED != 'true' }}
+        uses: andymckay/cancel-action@b9280e3f8986d7a8e91c7462efc0fa318010c8b1 # v0.3
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@e1e17a757e536f70e52b5a12b2e8d1d1c60e04ef # v2.0.0
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
olivereri, line 35:
```suggestion
role-to-assume: {{ vars.AWS_ASSUME_ROLE }}
```

You can remove those lines. OIDC is already configured for next-build. There is a trust setup between Github and AWS that uses and IAM role to grant permissions.

.github/workflows/production-tag.yml
```diff
@@ -0,0 +1,128 @@
+name: Create Production Tag
+
```
timcosgrove, line 2: This based on the Content Build Daily production release (tag), with some (IMHO) improvements and changes. 

https://github.com/department-of-veterans-affairs/content-build/actions/runs/7506884417/workflow

.github/workflows/production-tag.yml
```diff
@@ -0,0 +1,128 @@
...
+on:
```
timcosgrove, line 3: This intention here is that 2 things will trigger a new production tag:
* manual tagging (this is a backup)
* a completed successful run of the CI workflow on the main branch

This is to get away from daily timed tagging and enable more continual deployment with tests in place to ensure application integrity.

.github/workflows/production-tag.yml
```diff
@@ -0,0 +1,128 @@
...
+env:
+  SLACK_CHANNEL: C06DSBT7CBW #status-next-build
+  DSVA_SCHEDULE_ENABLED: true
+
+jobs:
+  create-production-tag:
+    name: Create Production Tag
+    runs-on: ubuntu-latest
+    # Run the workflow unless it was triggered by CI and that failed
+    if: ${{ !(github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'failure') }}
```
timcosgrove, line 23: We cannot prevent a failed but completed CI run from triggering the workflow, so we have this here to prevent the tagging from running if it was a failure.

.github/workflows/production-tag.yml
```diff
@@ -0,0 +1,128 @@
...
+
+      - name: Get current ref
+        id: get-current-ref
+        run: echo REF=$(git rev-parse HEAD) >> $GITHUB_OUTPUT
+
+      - name: Get latest tag
+        id: get-latest-tag
+        run: echo LATEST_TAG_VERSION=$(git fetch --all --tags > /dev/null && git tag -l | sort -V --reverse | head -n 1) >> $GITHUB_OUTPUT
+
+      - name: Get next tag version
```
timcosgrove, line 65: This was reworked some from the Content Build version, which assumes the numbering scheme will always be `v0.0.123` or some other number on the end. The rework expects a semver-type version (3 numbers separated by `.` but otherwise makes no assumptions about the incoming tag structure. 

It will output the same tag structure with the point version incremented by 1.

olivereri (approved):

PR #355, opened 2024-01-22
Broken link check workflow

timcosgrove:
## Description
Relates to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/16824

This sets up the broken link script to run on GHA. It is heavily parallelized, currently running on 64 instances. Once all the instances are complete, the workflow takes the resulting reports, combines them, generates a CSV and a Markdown report from the JSON, and attaches them all to the workflow. It then informs Slack that it is complete, with links to the CSV and Markdown docs.

There is nothing to do with the JSON file currently. Down the line if further work is done to link the broken link report to its individual pieces of content, or to create more of an application/dashboard for managing broken links, the JSON file will be useful.

## Testing done
Many many runs of the workflow, and examination of the resulting files.  Most recent run against the current state of these files: https://github.com/department-of-veterans-affairs/next-build/actions/runs/7876528021

One thing that will be an ongoing process is that the checker returns a number of false positives - that is, links it reports as broken when they are fine. This can happen for a variety of reasons, but the two most common seem to be:

* Requesting a link times out
* The remote server disallows automated traffic like ours

We currently have a 'skip' list of domains and patterns which we've made a decision to ignore. This list can grow as we find more domains that are consistently false positives. However, this should be done with caution, as adding a domain to the skip list means we will never find legitimate broken links on that domain.

Ideally, we would figure out how to get past server rejection so that we can check these domains legitimately.


playwright/utils/getSitemapLocations.js
```diff
@@ -43,11 +44,23 @@ async function getSitemapLocations(baseUrl) {
 
 // VA.gov sitemaps have a lot of urls in them. Helper function for things that
 // may want to parallelize checking that list (broken links, a11y, etc.)
-function splitPagesIntoSegments(pages, count) {
-  const segmentSize = Math.ceil(pages.length / count)
-  return new Array(count).fill().map((_, index) => {
-    return pages.slice(index * segmentSize, (index + 1) * segmentSize)
+function splitPagesIntoBatches(pages, batchCount) {
```
timcosgrove, line 47: These two functions could possible be dried up; they're somewhat similar. The first is a rework of the existing batching of pages and is used within a single instance. The second function `getPagesSlice` is used to provide a given review instance its set of files. The output of the second function ultimately gets passed to the first to be further broken up into batches.

scripts/broken-links-slack-payload.mjs
```diff
@@ -0,0 +1,72 @@
+/* eslint-disable no-console */
```
timcosgrove, line 1: This is an adaptation of the script used for content release broken links. It's greatly simplified because we decided to abandon sending actual broken link information directly to Slack.

scripts/broken-links-slack-payload.mjs
```diff
@@ -0,0 +1,72 @@
...
+import fs from 'fs'
+import core from '@actions/core'
+import { program } from 'commander'
+
+const DEFAULT_INPUT_FILE = 'broken-links-report.json'
+const DEFAULT_OUTPUT_FILE = 'broken-links-slack-payload.json'
+const GROUP_TO_NOTIFY = 'subteam^S010U41C30V|cms-helpdesk' // '<!subteam^S010U41C30V|cms-helpdesk>'
```
timcosgrove, line 8: This should be changed back to the commented version once we discuss this with CMS Helpdesk. The current version does not @ them.

scripts/broken-links-slack-payload.mjs
```diff
@@ -0,0 +1,72 @@
...
+const RUN_URL = `${process.env.GITHUB_SERVER_URL}/${process.env.GITHUB_REPOSITORY}/actions/runs/${process.env.GITHUB_RUN_ID}`
+
+// Gather any command line options or provide defaults
+program
```
timcosgrove, line 12: I used commander to process command-line flags. I realize this isn't consistent with other scripts. Can convert this to use env vars if we feel strongly.

scripts/broken-links-slack-payload.mjs
```diff
@@ -0,0 +1,72 @@
...
+  .option('-o, --output-file <output-file>', 'The output location of the Slack payload file', DEFAULT_OUTPUT_FILE)
+  .option('-m, --markdown-url <markdown-url>', 'The URL for the markdown report artifact')
+  .option('-c, --csv-url <csv-url>', 'The URL for the CSV report artifact')
+  .option('-d, --debug', 'Output debug information')
+  .parse(process.argv)
+
+const options = program.opts()
+
+// Wrap up regular and debug logging and only allow if the debug flag is set.
+const say = (level, ...args) => {
```
timcosgrove, line 24: Wrapper around console.log. The intention was to allow console.logs that only show when the debug flag is on. It ends up not very used.

scripts/check-broken-links.mjs
```diff
@@ -3,17 +3,60 @@ import { LinkChecker } from 'linkinator'
...
+  splitPagesIntoBatches,
+  getPagesSlice,
 } from '../playwright/utils/getSitemapLocations.js'
 import fs from 'fs'
 
 const OPTIONS = {
   sitemapUrl: process.env.SITE_URL || 'http://www.va.gov/',
-  batchSize: process.env.BATCH_SIZE || 32,
+  // totalInstances is the number of instances to split work among.
+  totalInstances: process.env.TOTAL_INSTANCES || 1,
```
timcosgrove, line 14: The default total instances and instance number is 1 each. This means that if the vars aren't set, the broken link script will act as before, checking the entire script.

This mechanism can also be (ab)used to do a slice during testing. 64 instances  = ~330 pages per slice. If you set 64 total instances and whatever number between 1 and 64 for the instance, you'll get a slice of the full scan.

scripts/check-broken-links.mjs
```diff
@@ -3,17 +3,60 @@ import { LinkChecker } from 'linkinator'
...
+  // instanceNumber is the specifc instance currently running the check.
+  // In this and totalInstances case, the defaults will check the entire set.
+  instanceNumber: process.env.INSTANCE_NUMBER || 1,
+  // batchSize is the number of pararllel link check processes to run.
+  batchSize: process.env.BATCH_SIZE || 20,
```
timcosgrove, line 19: This can be played with. I had better luck with 20 over 32. This was not verified scientifically, but I suspect that each runner has a limit on outgoing network connections.

scripts/check-broken-links.mjs
```diff
@@ -3,17 +3,60 @@ import { LinkChecker } from 'linkinator'
...
   verbose: process.env.VERBOSE || false,
   skipImageLinks: process.env.SKIP_IMAGES || false,
 }
 
+// List of patterns to skip checking. Most of these are to avoid false positives
+// from servers that disallow bot traffic like ours.
+// These can be regular expressions; no quotes around them in that case.
+const LINKS_TO_SKIP = [
```
timcosgrove, line 27: I do not love this mechanism at all, but it's necessary. It would be nice in the future to figure out how to check links on most of these domains successfully, rather than needing to exclude them, since this means we will never find legitimate broken links on these domains.

scripts/check-broken-links.mjs
```diff
@@ -25,16 +68,14 @@ const LOGGER_MAP = {
...
-  ],
-  timeout: 5000, // Fail a link if it doesn't resolve within 5s, otherwise linkinator will hang until it resolves.
+  linksToSkip: LINKS_TO_SKIP,
+  timeout: 10000, // Fail a link if it doesn't resolve, otherwise linkinator will hang until it resolves.
   urlRewriteExpressions: [
     // { pattern: '', replacement: '' }
   ],
   // recurse: true, // not recursing through links that are checked because we scan the full known sitemap
+  retryErrors: true,
+  retryErrorsCount: 3,
```
timcosgrove, line 78: This lowers the amount of false positives, but more importantly it ensures that the bulk of pages are checked. Without this, I was finding that a full script run was only scanning about 3000 of 25000 pages; presumably the remaining 22000 didn't load within the timeout.

scripts/check-broken-links.mjs
```diff
@@ -73,42 +114,78 @@ async function checkBrokenLinks() {
 
       linksChecked.push(result)
 
-      if (result.state === 'BROKEN') {
+      // 403 errors returned to the crawler by and large work for web users.
+      if (result.state === 'BROKEN' && result.status !== 403) {
```
timcosgrove, line 118: In testing 403s were almost always false positives.

scripts/check-broken-links.mjs
```diff
@@ -73,42 +114,78 @@ async function checkBrokenLinks() {
...
         brokenLinks.push(result)
       }
     })
+    .on('retry', (retryDetails) => {
+      console.log(`Retrying ${retryDetails.url} ; status: ${retryDetails.status}`)
```
timcosgrove, line 123: This isn't strictly necessary but leaving it here to provide some visibility into how retries are going for a particular run.

scripts/check-broken-links.mjs
```diff
@@ -73,42 +114,78 @@ async function checkBrokenLinks() {
...
+    })
 
   // Full array of sitemap defined URLs.
-  const paths = await getSitemapLocations(OPTIONS.sitemapUrl)
-  // Tiny array of paths for debugging this script.
-  // const paths = (await getSitemapLocations(OPTIONS.sitemapUrl)).slice(0, 100)
+  const allPaths = (await getSitemapLocations(OPTIONS.sitemapUrl))
+  const paths = getPagesSlice(allPaths, OPTIONS.totalInstances, OPTIONS.instanceNumber)
   console.log(`Number of pages to check: ${chalk.yellow(paths.length)}`)
+  const initialPathCount = paths.length
```
timcosgrove, line 130: Put this in to track how many page scans are getting dropped.

scripts/check-broken-links.mjs
```diff
@@ -131,43 +208,55 @@ async function checkBrokenLinks() {
...
       domain: OPTIONS.sitemapUrl,
+      initialPathCount: initialPathCount,
       pagesScanned: pagesChecked.length,
       linksChecked: linksChecked.length,
       brokenLinkCount: brokenLinks.length,
       time: formattedTime,
     },
-    brokenLinks: [],
+    brokenLinksByParent: {},
+    brokenLinksByLink: {},
```
timcosgrove, line 218: Added this here to allow for two views into the data. May not be necessary, since we're now providing a CSV which allows this sort of data view.

.github/workflows/broken-links-check.yml
```diff
@@ -0,0 +1,183 @@
+name: Broken link check (va.gov)
+
+on:
+  pull_request:
```
timcosgrove, line 4: `pull_request:` will be removed right before merge. Right now it's the only way to trigger the workflow.

package.json
```diff
@@ -21,8 +21,8 @@
     "format": "prettier --write .",
     "lint": "next lint",
     "plop": "plop",
-    "postinstall": "husky install && yarn certs",
-    "certs": "rm -f ./certs/VA-mozilla-combined.pem && cat node_modules/node_extra_ca_certs_mozilla_bundle/ca_bundle/ca_intermediate_root_bundle.pem ./certs/VA-Internal-S2-RCA-combined.pem > ./certs/VA-mozilla-combined.pem",
+    "postinstall": "yarn certs && husky install",
+    "certs": "rm -f ./certs/VA-mozilla-combined.pem && more node_modules/node_extra_ca_certs_mozilla_bundle/ca_bundle/ca_intermediate_root_bundle.pem > ./certs/VA-mozilla-combined.pem && more certs/VA-Internal-S2-RCA-combined.pem >> ./certs/VA-mozilla-combined.pem",
```
timcosgrove, line 25: Rearranged this a bit because I was having unreliable results with `cat`. Using `more` and doing it on each file individually seems to be working more reliably.

scripts/check-broken-links.mjs
```diff
@@ -73,42 +113,81 @@ async function checkBrokenLinks() {
 
       linksChecked.push(result)
 
-      if (result.state === 'BROKEN') {
+      // 403 errors returned to the crawler by and large work for web users.
+      if (result.state === 'BROKEN' && result.status !== 403) {
         brokenLinks.push(result)
       }
     })
+    .on('retry', (retryDetails) => {
```
timcosgrove, line 121: Not strictly necessary but at this stage it is good to have visibility into how retries are going.

playwright/utils/getSitemapLocations.js
tjheffner: not a blocker for this PR because we don't actually run the full a11y scan right now, but the initial stub (`playwright/tests/a11y.spec.js`) used the old `splitPagesIntoSegments` for its test batching. that will need to be reworked to match this.

scripts/broken-links-slack-payload.mjs
tjheffner: the env-loader package also uses commander to process cli flags into env vars, i'm fine with doing this separately here because it's a separate thing

scripts/broken-links-slack-payload.mjs
```diff
@@ -0,0 +1,81 @@
...
+        type: 'mrkdwn',
+        text: `
+          ${icon} *Broken link check* (Workflow run: <${RUN_URL}>)\n
+          *Pages scanned:* ${brokenLinks.metrics.pagesScanned}\n
+          *Broken links:* *${brokenLinks.metrics.brokenLinkCount}* out of ${brokenLinks.metrics.linksChecked} checked\n\n
+          *Markdown report:* ${options.markdownUrl}\n
+          *CSV report:* ${options.csvUrl}\n\n
+          cc: ${GROUP_TO_NOTIFY}`,
+      },
+    })
```
tjheffner, line 70: Screenshot 2024-02-12 at 11 28 27 AM

just noting this looks like this, links go to download of a given run's artifacts. informative and not noisy. thank you!

tjheffner (approved):
only one note about the batch functions probably breaking the a11y playwright test... but we should worry about that when we actually implement that to run regularly (it does not run at all right now)

PR #359, opened 2024-01-24
VACMS-8636: Adds component for rendering react-widget paragraph.

ryguyk:
## Description
Relates to: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/8636

## Testing done/QA steps
Test Locally
- [ ] Add various react-widget components inside `<Wrapper>` in `[[...slug]].tsx` and ensure they render the expected react widget.
```
    <Wrapper
      bannerData={bannerData}
      headerFooterData={headerFooterData}
      preview={preview}
      resource={resource}
    >
      <ReactWidget entityId={0} widgetType="pension-app-status" />

   .
   .
   .
```
- [ ] Navigate to any page (e.g. http://localhost:3000/minneapolis-health-care/stories/)
    - Examples:
        - [ ] `<ReactWidget entityId={0} widgetType="pension-app-status" />`
        - [ ] `<ReactWidget entityId={0} widgetType="direct-deposit" ctaWidget={true} />`

### DIFF
--- a/src/data/queries/reactWidget.ts
+++ b/src/data/queries/reactWidget.ts
@@ -6,18 +6,18 @@ export const formatter: QueryFormatter<ParagraphReactWidget, ReactWidget> = (
   entity: ParagraphReactWidget
 ) => {
   return {
-    entityId: entity?.drupal_internal__id,
-    ctaWidget: entity?.field_cta_widget,
-    widgetType: entity?.field_widget_type,
-    loadingMessage: entity?.field_loading_message,
-    timeout: entity?.field_timeout,
-    errorMessage: entity?.field_error_message.processed,
+    entityId: entity.drupal_internal__id,
+    ctaWidget: entity.field_cta_widget,
+    widgetType: entity.field_widget_type,
+    loadingMessage: entity.field_loading_message,
+    timeout: entity.field_timeout,
+    errorMessage: entity.field_error_message.processed,
     defaultLink: entity.field_default_link
       ? {
           url: entity.field_default_link.url,
           title: entity.field_default_link.title,
         }
       : null,
-    buttonFormat: entity?.field_button_format,
+    buttonFormat: entity.field_button_format,
   }
 }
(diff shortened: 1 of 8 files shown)


jtmst (dismissed):

src/data/queries/reactWidget.ts
```diff
@@ -0,0 +1,23 @@
+import { ParagraphReactWidget } from '@/types/drupal/paragraph'
+import { QueryFormatter } from 'next-drupal-query'
+import { ReactWidget } from '@/types/formatted/reactWidget'
+
+export const formatter: QueryFormatter<ParagraphReactWidget, ReactWidget> = (
+  entity: ParagraphReactWidget
+) => {
+  return {
```
tjheffner, line 8: not a blocker just a stylistic question. all of these have `entity?.` in some fashion, could we just `if (!entity) return` beforehand?

tjheffner (approved):
nice work! widgets appear to load just fine. I see the loading state

src/data/queries/reactWidget.ts
ryguyk: The returned object needs to comply with the type `ReactWidget` which requires `entityId` and `widgetType`. So, we can't just return an empty object.

That said, this did make me realize that I don't need all those `?` instances. `entity` will always be defined, so those are not necessary.

With regards to the repeated `entity.[some_field]`, there are other ways we could write this to not have that repeated everywhere, but they don't necessarily make the code any more concise (see example below).

```
  const {
    drupal_internal__id,
    field_widget_type,
    field_cta_widget,
    field_loading_message,
    field_timeout,
    field_error_message,
    field_default_link,
    field_button_format,
  } = entity

  return {
    entityId: drupal_internal__id,
    widgetType: field_widget_type,
    ctaWidget: field_cta_widget,
    loadingMessage: field_loading_message,
    timeout: field_timeout,
    errorMessage: field_error_message.processed,
    defaultLink: field_default_link
      ? {
          url: field_default_link.url,
          title: field_default_link.title,
        }
      : null,
    buttonFormat: field_button_format,
  }
```

tjheffner: yeah not worried at all about deconstructing off for the formatter, was mostly just curious about the `?`. if reactwidget always needs it / will always have it, cool

tjheffner (approved):

PR #360, opened 2024-01-25
paragraph--alert_single added, with logic for reusable and non-reusable alert types

jtmst:
## Description
Relates to:
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/8608
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/8608

Brings in queries and component for the alert_single type.  Also addresses the logic required to render Non reusable alerts in the same component.

Changes:
Adds alertSingle react component
Adds alertSingle query
Updates types and property names where necessary
Adds jest test for alert component
Adds storybook for AlertSingle

## Testing done
Local testing

## Screenshots
Screen Shot 2024-01-25 at 10 43 22 AM

## QA steps
Verify that component renders properly with storybook
Further verification when resources and support page is added and this component is actually used

### DIFF
--- a/src/data/queries/alertSingle.ts
+++ b/src/data/queries/alertSingle.ts
@@ -8,11 +8,11 @@ import { drupalClient } from '@/lib/drupal/drupalClient'
 import { queries } from '.'
 import { ParagraphAlertSingle } from '@/types/drupal/paragraph'
 import { AlertSingle } from '@/types/formatted/alertSingle'
+import { transformAlertData } from '@/lib/drupal/query'
 
 export const params: QueryParams<null> = () => {
   return queries
     .getParams()
-    .addPageLimit(3)
     .addInclude([
       'field_alert_block_reference',
       'field_alert_block_reference.field_alert_content',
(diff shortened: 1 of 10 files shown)


src/data/queries/alertSingle.ts
```diff
@@ -0,0 +1,64 @@
...
+} from 'next-drupal-query'
+import { drupalClient } from '@/lib/drupal/drupalClient'
+import { queries } from '.'
+import { ParagraphAlertSingle } from '@/types/drupal/paragraph'
+import { AlertSingle } from '@/types/formatted/alertSingle'
+
+export const params: QueryParams<null> = () => {
+  return queries
+    .getParams()
+    .addPageLimit(3)
```
tjheffner, line 15: i don't think we want a page limit on this (just an example when gathering data from the `_playground` route)

src/data/queries/index.ts
```diff
@@ -37,6 +38,7 @@ export const QUERIES_MAP = {
   'paragraph--link_teaser': LinkTeaser,
   'paragraph--rich_text_char_limit_1000': Wysiwyg,
   'paragraph--wysiwyg': Wysiwyg,
+  'paragragh--alert_single': AlertSingle,
   'media--image': MediaImage,
   'block--alert': Alert,
```
tjheffner, line 43: this `block--alert` feels very similar to the alertSingle query you've got, can it be refactored to use the same `transformAlert` function you've added in `alertSingle`? or removed if it's no longer needed?

tjheffner (changes requested):
one small change around the page limit, and one question around `block--alert`. otherwise, this looks great and looks good in storybook. nice!

src/data/queries/alertSingle.ts
jtmst: Good catch, fixed

src/data/queries/index.ts
jtmst: I refactored them both to use the transformAlert function. It feels natural to leave them separate in case there are further modifications needed in the future

tjheffner (approved):

PR #367, opened 2024-01-29
Fix AudienceTopics styles and add types

jtmst:
## Description
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/16617

Tags for resource and support taxonomies were largely already done.  This PR fixes up some styling and adds type safety to the getTags function.

Changes
- Styling for tag borders
- Styling for spacing
- Types in query function

## Testing done
Local

## Screenshots
image

## QA steps
Verify tags render as expected in storybook

### DIFF
--- a/src/data/queries/audienceTopics.ts
+++ b/src/data/queries/audienceTopics.ts
@@ -3,14 +3,21 @@ import { ParagraphAudienceTopics } from '@/types/drupal/paragraph'
 import { QueryFormatter } from 'next-drupal-query'
 import { AudienceTopic } from '@/types/formatted/audienceTopics'
 
-const getTagsList = (fieldTags) => {
-  if (!fieldTags) return null
+interface Tag {
+  id: string
+  href: string | undefined
+  name: string
+  categoryLabel: string
+}
+
+const getTagsList = (entity: ParagraphAudienceTopics): Tag[] | null => {
+  if (!entity) return null
 
   const {
     field_topics: fieldTopics = [],
     field_audience_beneficiares: fieldAudienceBeneficiares,
     fieldNonBeneficiares: fieldNonBeneficiares,
-  } = fieldTags
+  } = entity
 
   const topics = fieldTopics.map((topic) => ({
     id: topic.id,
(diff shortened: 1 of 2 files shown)


src/templates/components/audienceTopics/index.tsx
```diff
@@ -6,10 +6,11 @@ export function AudienceTopics({ tags }: FormattedAudienceTopics) {
   if (isEmpty(tags)) return null
   const tagsList = tags.map(({ id, href, name }) => (
     <div key={id}>
-      <div className="vads-u-margin-right--1 vads-u-margin-bottom--1 medium-screen:vads-u-margin-bottom--0">
+      <div className="vads-u-margin-right--1 vads-u-margin-bottom--1 medium-screen:vads-u-margin-bottom--1p5">
         <Link
           href={`${href}/${encodeURI(name)}`}
```
tjheffner, line 11: i don't think this `href` is correct. the `/${encodeURI(name)}` portion can be dropped.

looking at va.gov: 
Screenshot 2024-01-30 at 8 29 26 AM


this pr:
`/resources/tag/all-veterans/All%20Veterans`

tjheffner (changes requested):

src/data/queries/audienceTopics.ts
```diff
@@ -3,14 +3,21 @@ import { ParagraphAudienceTopics } from '@/types/drupal/paragraph'
 import { QueryFormatter } from 'next-drupal-query'
 import { AudienceTopic } from '@/types/formatted/audienceTopics'
 
-const getTagsList = (fieldTags) => {
-  if (!fieldTags) return null
+interface Tag {
```
tjheffner, line 6: is this needed?

if the formatter is returning `AudienceTopic[]` but it just calls getTagsList which returns `Tag[]`, can we re-use one or the other?

src/templates/components/audienceTopics/index.tsx
jtmst: good catch, ty

src/data/queries/audienceTopics.ts
jtmst: Missed that, thanks

tjheffner (approved):

PR #369, opened 2024-01-30
update banner endpoint call to match new cms endpoint

tjheffner:
## Description
Relates to #[16781](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/16781).

temporarily butchering banner types in next-build so they render properly with the updated endpoint

## Testing done
locally with a local up-to-date cms. created one banner of each type and verified they could render in next-build still.

## Screenshots
Screenshot 2024-01-30 at 2 59 42 PM

## QA steps

```[tasklist]
```


timcosgrove (approved):

src/data/queries/banners.ts
```diff
@@ -11,57 +11,58 @@ export type BannerDataOpts = {
 // A given page does not reference a banner node via entity reference, banner node types have a field that lists what
 // paths they are supposed to be visible on. This endpoint queries banners based on their path lists.
 // See docroot/modules/custom/va_gov_api/src/Resources/BannerAlerts.php in va.gov-cms for more info.
```
alexfinnarn, line 13: Might want to update the reference to the code that provides the banner response.

PR #370, opened 2024-02-01
Initial pass at Dockerfile and supporting files

jschmidt-civicactions:
## Description
Relates to #[16857](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/16857). Adds Dockerfile for next-build container.

## Testing done
Docker container builds, full testing will require k8s env and additional config pieces there.

## QA steps

```[tasklist]
```

### DIFF
--- a/.github/workflows/mirror-images.yml
+++ b/.github/workflows/mirror-images.yml
@@ -1,0 +1,37 @@
+name: Create and Commit Next-Build Docker Image
+
+on:
+  workflow_dispatch:
+    inputs:
+      image_tag:
+        description: Image tag version
+        type: string
+        required: true
+        default: "1.0.0"
+jobs:
+  mirror:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-region: us-gov-west-1
+          role-to-assume: ${{ vars.AWS_ASSUME_ROLE }}
+          role-duration-seconds: 900
+          role-session-name: vsp-vagov-next-build-githubaction
+
+      - name: Login to Amazon ECR
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@v2
+
+      - name: Build, tag, and push image to Amazon ECR
+        env:
+          ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
+          ECR_REPOSITORY: "dsva/next-build-node"
+          IMAGE_TAG: ${{ inputs.image_tag }}
+        run: |
+          BUILD_TYPE=vagovprod node ./scripts/yarn/vets-website-assets.mjs
+          docker build -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG -f docker/Dockerfile
+          docker push $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG
--- a/.github/workflows/mirror-images.yml
+++ b/.github/workflows/mirror-images.yml
@@ -13,6 +13,13 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v3
+        with:
+          path: next-build
+
+      - uses: actions/checkout@v3
+        with:
+          repository: department-of-veterans-affairs/vets-website
+          path: vets-website
 
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v1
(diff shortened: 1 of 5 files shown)


docker-run.sh
```diff
@@ -0,0 +1,12 @@
...
+# Shell script invoked by next-build docker container (see docker/Dockerfile).
+# Env vars must be present to populate cmd params.
+# Secrets (remaining required params) must be present at ./envs/.env.${ENV}.
+
+APP_ENV=${ENV} yarn dev \
+    --SITE_URL ${SITE_URL} \
+    --NEXT_PUBLIC_DRUPAL_BASE_URL ${NEXT_PUBLIC_DRUPAL_BASE_URL} \
+    --NEXT_IMAGE_DOMAIN ${NEXT_IMAGE_DOMAIN} \
+    --REDIS_URL ${REDIS_URL} \
+    --NEXT_PUBLIC_ASSETS_URL=./public/generated/
```
tjheffner, line 12: with the way the app is set up, the `APP_ENV` value ties it to a specific .env file. e.g. `APP_ENV=prod` will load env vars from the `envs/.env.prod` file. Do we need to overwrite env vars that would be found in those files?

docker-run.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+# Shell script invoked by next-build docker container (see docker/Dockerfile).
+# Env vars must be present to populate cmd params.
+# Secrets (remaining required params) must be present at ./envs/.env.${ENV}.
+
+APP_ENV=${ENV} yarn dev \
```
tjheffner, line 7: Seeing as these are for the BRD-adjacent preview servers, I think ultimately we will want to be running `yarn build:preview` && `yarn start` instead of `yarn dev`, but this is fine for now

docker-run.sh
jschmidt-civicactions: Right, my thought is that k8s supplies env-specific configs in two ways:
1. ConfigMap that sets all the vars found in `docker-run.sh`.
2. Then a k8s Secret creates the correct env file for the given environment, which does contain the remaining configs that happen to be actual secrets (eg. DRUPAL_PREVIEW_SECRET, etc).

jschmidt-civicactions: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/16857#issuecomment-1921388478

docker/.dockerignore
```diff
@@ -0,0 +1,7 @@
+Dockerfile
+.dockerignore
+node_modules
+npm-debug.log
+READMEs
+.next
```
tjheffner, line 6: i think we'll want to include `.next` eventually when running `yarn build` / `yarn start`

docker-run.sh
tjheffner: ah, got it. this is to keep a single image for all the different environments, i missed that note. makes sense 😄

jschmidt-civicactions:
@tjheffner I will be adding the Github actions changes for container builds to this branch as well, since Eric has a start on it and he's focusing on the datasync ticket, I'll just add that here. I'll also make the suggested changes to docker-run.sh and .dockerignore.

I'll ping you again when it's ready for final review (marking this as draft for now). Thanks!

.github/workflows/mirror-images.yml
```diff
@@ -0,0 +1,37 @@
...
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@v2
+
+      - name: Build, tag, and push image to Amazon ECR
+        env:
+          ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
+          ECR_REPOSITORY: "dsva/next-build-node"
+          IMAGE_TAG: ${{ inputs.image_tag }}
+        run: |
+          BUILD_TYPE=vagovprod node ./scripts/yarn/vets-website-assets.mjs
```
tjheffner, line 35: i think this may need to run inside the docker image (or as part of it?).

In addition to snagging files from the s3 bucket, this script does pull some files from a package in `node_modules` for fonts, so i think running this before the dockerfile runs `yarn install` may miss some assets

tjheffner (commented):
at a glance things look good, minus the one comment around when the vets-website script runs

also noting every file may need a new line at the end

docker/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+RUN mkdir -p /app/content-build
+RUN mkdir -p /app/vets-website
+
+COPY ./next-build /app/next-build
+COPY ./vets-website /app/vets-website
+
+WORKDIR /app/next-build
+RUN yarn install --frozen-lockfile
+RUN BUILD_TYPE=vagovprod node ./scripts/yarn/vets-website-assets.mjs
```
tjheffner, line 15: :boom:

tjheffner (changes requested):
one small change around NEXT_PUBLIC_ASSETS_URL, and a question around the other env vars. the rest all LGTM!

tjheffner (approved):

PR #381, opened 2024-02-05
VACMS-8619 contact info

tjheffner:
## Description
Closes [#8619](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/8619).
Closes [#8613](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/8613).
Closes [#8646](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/8646).

Creates larger ContactInfo component, and a re-usable DefaultContact component that can handle phone or email contacts.
Creates formatter for the larger component and for node--support_service (essentially a list of phone numbers)

Updates the generator templates to output correct casing when initially created.

## Testing done
Storybook, until the pages these are used on are implemented

## QA steps

```[tasklist]
```


src/data/queries/contactInfo.ts
```diff
@@ -0,0 +1,57 @@
+import { QueryParams, QueryFormatter } from 'next-drupal-query'
+import { ParagraphContactInformation } from '@/types/drupal/paragraph'
+import { ContactInfo } from '@/types/formatted/contactInfo'
```
jtmst, line 3: Could make this as FormattedContactInfo for consistency

src/data/queries/supportServices.ts
```diff
@@ -0,0 +1,22 @@
...
+  entities: NodeSupportService[]
+) => {
+  if (!entities) return null
+
+  return entities.map((entity) => {
+    if (entity.status) {
+      return {
+        title: entity.title,
+        value: entity.field_phone_number,
+        href: entity.field_link.uri,
```
ryguyk, line 16: We'll likely need to change this to `url` once we get the JSON:API fix in place.

src/data/queries/supportServices.ts
```diff
@@ -0,0 +1,22 @@
+import { QueryFormatter } from 'next-drupal-query'
+import { NodeSupportService } from '@/types/drupal/node'
+import { Contact } from '@/types/formatted/contactInfo'
```
jtmst, line 3: Same thing here

src/pages/_playground/api-explorer.tsx
```diff
@@ -21,9 +21,18 @@ export async function getStaticProps() {
     }
   }
 
-  const params = queries.getParams().addInclude(['field_cta'])
```
jtmst, line 24: curious if this got pushed in on accident or if you wanted to update

src/pages/_playground/api-explorer.tsx
```diff
@@ -21,9 +21,18 @@ export async function getStaticProps() {
...
+      'field_benefit_hub_contacts',
+      'field_benefit_hub_contacts.field_support_services',
+    ])
+    .addPageLimit(1)
   const data = await drupalClient.getResourceCollection(
-    'paragraph--featured_content',
+    'paragraph--contact_information',
     {
       params: params.getQueryObject(),
       withAuth: {
```
ryguyk, line 38: Do we need this file committed to source control?

src/types/formatted/emailContact.ts
```diff
@@ -1,5 +1,4 @@
 export type EmailContact = {
-  id: string
   label: string
   address: string
 }
```
ryguyk, line 4: Should this file be deleted? Is `EmailContact` now included in `/formatted/contactInfor.ts`?

src/pages/_playground/api-explorer.tsx
tjheffner: i think we do want it because it's a nice helper for formatting data for `@/mocks/` (we can `console.log(json.stringify(data))` in a given query too, but i prefer this helper route

src/pages/_playground/api-explorer.tsx
tjheffner: probably don't need to update it here, will revert

src/data/queries/contactInfo.ts
tjheffner: looking at this, the `Type as FormattedType` only comes into play in the templates, the queries all seem to import the type name as-is

jtmst (approved):
thanks for changes, lgtm

src/data/queries/contactInfo.ts
ryguyk: Yeah, this is a result of naming conflict in the template.

src/data/queries/contactInfo.ts
```diff
@@ -0,0 +1,57 @@
...
+// paragraph--contact_information is essentially a wrapper paragraph around several other entity references
+export const formatter: QueryFormatter<
+  ParagraphContactInformation,
+  ContactInfo
+> = (entity: ParagraphContactInformation) => {
+  const additionalContactType = entity.field_additional_contact?.type
+  let additionalContact = null
+
+  switch (additionalContactType) {
+    case 'paragraph--email_contact':
```
ryguyk, line 25: `paragraph--email_contact` should maybe be replaced with an enum/const reference, but this can be cleaned up when we revisit the type cleanup as a whole.

tjheffner: sounds good. going to merge as-is for now just to keep things moving / make these available for the layouts that need them knowing it will be iterated on

ryguyk: Though, this might go away. See comment below about how we might format this paragraph.

src/data/queries/contactInfo.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  return {
+    contactType: entity.field_contact_info_switch,
+    defaultContact: {
+      title: entity.field_contact_default.title,
+      value: entity.field_contact_default.field_phone_number,
+      href: entity.field_contact_default.field_link.uri,
+    },
+    additionalContact,
```
ryguyk, line 51: Some of the work I've done around the overarching structure of handling paragraphs includes the creation of a function `formatParagraph` that can take a paragraph of any type and return the formatted version. Since `field_additional_contact` is a field that can take variable types of paragraphs, I think this might eventually become:

`additionalContact: additionalContactType ? formatParagraph(entity.field_additional_contact) : null`

This, though, implies that `paragraph--phone_number` and `paragraph--email_contact` have their own formatter and templates. It seems like this PR actually moves away from that and removes email as its own standalone template. I'm curious about this.

PR #382, opened 2024-02-06
Use `class` instead of `className` in custom web component

ryguyk:
## Description
TIL that the use of `className` in place of `class` in JSX does not apply to custom web components. So, accordingly, this PR updates incorrect usage throughout the repo.


jtmst (approved):

tjheffner (approved):
verified this renders the classes as expected, and wasn't previously. good catch!

pr:
Screenshot 2024-02-06 at 8 13 15 AM

main:
Screenshot 2024-02-06 at 8 13 02 AM

PR #405, opened 2024-02-13
Adjustments for k8s app deploy and dockerignore fixes

jschmidt-civicactions:
## Description
Relates to #[16857](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/16857).

## QA steps

```[tasklist]
```

### DIFF
--- a/docker/.dockerignore
+++ b/docker/.dockerignore
@@ -1,6 +1,8 @@
-Dockerfile
-.dockerignore
-node_modules
-npm-debug.log
-READMEs
-.git
+# ignore directories
+**/node_modules/*
+**/envs/*
+**/READMEs/*
+**/.git/*
+**/.github/*
+# ignore files
+**/npm-debug.log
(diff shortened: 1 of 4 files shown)


docker/.dockerignore
```diff
@@ -1,6 +1,8 @@
-Dockerfile
-.dockerignore
```
tjheffner, line 2: just checking, do we still want to ignore the Dockerfile and .dockerignore, or is below all we need?

jschmidt-civicactions: I recently learned that ignoring these can result in odd/undefined behavior. So I dropped those.

Docker also expects some weirdness when the context is set the way I have it (one dir above next-build so that vets-website can be part of it), so that's also why I'm doing the cp shenanigans in the workflow. All sorts of oddness I wasn't previously aware of!

tjheffner: interesting, TIL. thanks!

tjheffner (approved):

PR #427, opened 2024-03-01
14370 q a component

jtmst:
## Description
Relates to #[14370](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/14370). (or closes?)

CHANGES:
QA component
QA Query
QA mock data
QA tests/types

## Testing done
Local, will verify in tugboat

## Screenshots
image

## QA steps
Verify QA component renders in storybook

```[tasklist]
```


src/types/drupal/paragraph.ts
```diff
@@ -166,6 +167,12 @@ export interface ParagraphQAGroup extends DrupalParagraph {
   field_section_header: string
 }
 
+export interface ParagraphQaParagraph extends DrupalParagraph {
```
tjheffner, line 170: `ParagraphQaParagraph` feels off. `ParagraphQA` or `ParagraphQuestionAnswer` ?

playwright/tests/qaParagraph.spec.js
```diff
@@ -0,0 +1,19 @@
+const { test, expect } = require('../utils/next-test')
```
tjheffner, line 1: you can remove this file, we'll only need playwright tests for actual layout pages

src/templates/layouts/qaParagraph/index.tsx
```diff
@@ -0,0 +1,21 @@
+import { QaParagraph as FormattedQaParagraph } from '@/types/formatted/qaParagraph'
```
tjheffner, line 1: i don't think this should live in `templates/layouts/qaParagraph`. I think something like `templates/components/qa` makes more sense

`QuestionAnswer` is fine too if the shorthand qa feels funky when Group and Section are also added

tjheffner (commented):
as an initial stub i have some nits around naming and template location.

just verifying that it's supposed to render as plain wysiwygs, instead of in an accordion

Screenshot 2024-03-01 at 8 54 50 AM

src/templates/layouts/qaParagraph/index.tsx
jtmst: I was trying to avoid having QuestionAnswer as a layout and QuestionAnswer as a component, but if you think that would be less confusing im open to it

playwright/tests/qaParagraph.spec.js
jtmst: Gothca

jtmst:
> as an initial stub i have some nits around naming and template location.
> 
> just verifying that it's supposed to render as plain wysiwygs, instead of in an accordion
> 
> Screenshot 2024-03-01 at 8 54 50 AM

Given that we're using the <Paragraph/> Component, this could render anything.  In the context of this going into q_a section, this QA component would primarily be used for wysiwyg rendering, as if its meant to be an accordion its more likely to use the q_a_collapsible panel template.

src/data/queries/qaParagraph.ts
```diff
@@ -0,0 +1,14 @@
...
+import { ParagraphQaParagraph } from '@/types/drupal/paragraph'
+import { QaParagraph } from '@/types/formatted/qaParagraph'
+
+export const formatter: QueryFormatter<ParagraphQaParagraph, QaParagraph> = (
+  entity: ParagraphQaParagraph
+) => {
+  return {
+    question: entity.field_question,
+    answers: entity.field_answer.map(
+      (paragraphWysiwyg) => paragraphWysiwyg.field_wysiwyg
```
tjheffner, line 11: i don't think this is correct as these `field_answer` can technically be more than just wysiwyg paragraphs.

we'll likely want an approach like the one in the resourcesSupport query here: https://github.com/department-of-veterans-affairs/next-build/blob/main/src/data/queries/resourcesSupport.ts#L84

src/data/queries/index.ts
```diff
@@ -68,12 +69,14 @@ export const QUERIES_MAP = {
...
   'paragraph--non_reusable_alert': AlertNonReusable,
   'paragraph--phone_number': PhoneNumber,
+  'paragraph--q_a': QaParagraph,
   'paragraph--react_widget': ReactWidget,
   'paragraph--rich_text_char_limit_1000': Wysiwyg,
   'paragraph--table': Table,
   'paragraph--wysiwyg': Wysiwyg,
   // Blocks
   'block--alert': AlertBlock,
+  'paragragh--alert_single': AlertSingle,
```
tjheffner, line 79: this exists on line 59 already

tjheffner (commented):
changes look good to me, just one extra thing snuck into the index query

src/data/queries/index.ts
jtmst: error on fixing merge conflicts, thanks

tjheffner (approved):

PR #437, opened 2024-03-05
VACMS-17399: Adds more granular definition of listing pages

ryguyk:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/17399

Event-listing pages do not need all "paged" pages generated statically because the paging mechanism is done client side. This PR introduces some additional config and helper functions to be able to more granularly define whether a content type is a listing page that needs static generation of all "paged" pages or if it is a listing page that does not need that (but still needs other listing-page treatment).

The biggest gain here is efficiency of the build, as this PR will remove some unnecessary processing that was taking place for event-listing pages. Also important, this should make the code much more easy to follow, and a much more accurate representation of how things are structured/treated.

## Testing done/QA steps
The biggest risk here is screwing up existing static-page generation for story-listing pages. So, the biggest thing to test is that story-listing pages still have subsequent "paged" pages statically built. We were never generating subsequent pages for event-listing pages, but we also want to test that they do not exist.

Some examples of things that _should_ exist:
- [ ] `/minneapolis-health-care/stories/page-2`
- [ ] `/minneapolis-health-care/events`

Some examples of things that _should not_ exist:
- [ ] `/minneapolis-health-care/events/page-2`

### DIFF
--- a/src/lib/drupal/listingPages.ts
+++ b/src/lib/drupal/listingPages.ts
@@ -15,7 +15,14 @@ const LISTING_RESOURCE_TYPES = [
   RESOURCE_TYPES.EVENT_LISTING,
 ] as const
 
+const SINGLE_PAGE_LISTING_RESOURCE_TYPES = [
+  RESOURCE_TYPES.EVENT_LISTING,
+] as const
+
 export type ListingResourceType = (typeof LISTING_RESOURCE_TYPES)[number]
+
+export type SinglePageListingResourceType =
+  (typeof SINGLE_PAGE_LISTING_RESOURCE_TYPES)[number]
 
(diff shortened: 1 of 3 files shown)


src/lib/drupal/listingPages.ts
```diff
@@ -15,8 +15,15 @@ const LISTING_RESOURCE_TYPES = [
   RESOURCE_TYPES.EVENT_LISTING,
 ] as const
 
+const SINGLE_PAGE_LISTING_RESOURCE_TYPES = [
```
timcosgrove, line 18: Can we document why these are distinct from other listing pages? A comment here with a simple explanation and a link to a README with some more detail would be great.

tjheffner (approved):
thank you

PR #440, opened 2024-03-06
VACMS-16863: Trigger manifest update from image creation workflow.

timcosgrove:
## Description
Relates to #[16863](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/16863). 

Triggers manifest update from image creation workflow.

### DIFF
--- a/.github/workflows/mirror-images.yml
+++ b/.github/workflows/mirror-images.yml
@@ -15,7 +15,7 @@ on:
   workflow_run:
     workflows: ['Create Production Tag']
     types: [completed]
-#    branches: [main]
+    branches: [main]
 
 jobs:
   mirror:
@@ -35,7 +35,7 @@ jobs:
           repository: department-of-veterans-affairs/vets-website
           path: vets-website
 
-      - name: 'Download artifact'
+      - name: 'Download tag artifact'
         if: ${{ github.event_name == 'workflow_run' }}
         uses: actions/github-script@v6
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/mirror-images.yml
```diff
@@ -88,4 +88,13 @@ jobs:
           docker build . -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG -f ./Dockerfile
           docker push $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG
 
+      - name: Write release name as artifact
```
timcosgrove, line 91: This is the same logic as in the production tag creation workflow.

.github/workflows/update-manifest.yml
```diff
@@ -22,40 +22,110 @@ on:
...
         required: true
         default: "staging"
+  workflow_run:
+    workflows: ['Create and Commit Next-Build Docker Image']
+    types: [completed]
+    branches: [main]
 
 jobs:
   update-manifest:
+    if: ${{ !(github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'failure') }}
```
timcosgrove, line 32: Same as before: if this is triggered by workflow_run and that was a failure, do not run.

.github/workflows/update-manifest.yml
```diff
@@ -22,40 +22,110 @@ on:
...
     runs-on: ubuntu-20.04
+    env:
+      IMAGE_TAG: ${{ inputs.image_tag }}
+    strategy:
+      fail-fast: false
+      max-parallel: 4
+      matrix:
```
timcosgrove, line 39: This setup is so we can deploy all environments at once.

.github/workflows/update-manifest.yml
```diff
@@ -22,40 +22,110 @@ on:
...
           cwd: vsp-infra-application-manifests/apps/${{ inputs.app_name }}/${{ inputs.environment }}
           author_name: va-vsp-bot
           author_email: devops@va.gov
-          message: 'auto update next-build images and helm chart'
\ No newline at end of file
+          message: 'auto update next-build images and helm chart'
+
+      # If this is triggered via workflow_run, run as a matrix and deploy all
+      # apps and envs.
+      - name: Update image and helm chart versions (triggered by upstream)
```
timcosgrove, line 116: If these steps run (because triggered by workflow_run), it will cause the job to be split out into 4 parallel jobs which will deploy each environment.

tjheffner (approved):

PR #441, opened 2024-03-06
VACMS-16863: Further work on automated manifest update

timcosgrove:
## Description
Relates to #[16863](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/16863). 

Previously we tried using a matrix - so 4 parallel instances - to update the 4 environments for Next Build EKS.

This doesn't work because manifest updates are a 'gitops' operation. We can't do these in parallel because it creates merge conflicts.

### DIFF
--- a/.github/workflows/update-manifest.yml
+++ b/.github/workflows/update-manifest.yml
@@ -33,12 +33,6 @@ jobs:
     runs-on: ubuntu-20.04
     env:
       IMAGE_TAG: ${{ inputs.image_tag }}
-    strategy:
-      fail-fast: false
-      max-parallel: 4
-      matrix:
-        app_name: ['next-build', 'next-build-test']
-        environment: ['staging', prod]
     steps:
       - name: 'Download tag artifact'
         if: ${{ github.event_name == 'workflow_run' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/update-manifest.yml
```diff
@@ -111,21 +105,71 @@ jobs:
...
+          message: 'auto update next-build images and helm chart'
+
+      - name: Update image and helm chart versions, Prod Next Build Test
+        if: ${{ github.event_name == 'workflow_run' }}
+        run: |
+            cd vsp-infra-application-manifests/apps/next-build-test/prod
+            yq e -i '.deployment.container.image = "008577686731.dkr.ecr.us-gov-west-1.amazonaws.com/dsva/next-build-node:${{ env.IMAGE_TAG }}"' values.yaml
+            git diff
+
+      - name: Add and Commit file, Prod Next Build Test
```
jschmidt-civicactions, line 167: I think this can be simplified to a single add-and-commit at the end here. If the working dir is 'vsp-infra-application-manifests/apps' (or higher) the add for '*.yaml' should recursively add all changed yaml files and it will be one commit.

tjheffner (approved):

PR #443, opened 2024-03-07
8629 qa group section

jtmst:
## Description
issues:
 [#8629](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/8629)
[#17322](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/17322)

Changes
Adds q_a_section and q_a_group queries/tests/types
adds q_a_section component with tests
adds qaCollapsible panel component with tests
modifies existing qaParagraph component and types slightly

## Testing done
Local Visual
verified components render on R&S pages
## Screenshots
Screen Shot 2024-03-07 at 10 33 20 AM
image

## QA steps
Verify components render as expected in storybook
Verify QA components render on R&S pages, (ie: http://localhost:3000/resources/creating-an-account-for-vagov/)

```[tasklist]
```

### DIFF
--- a/src/data/queries/qaGroup.ts
+++ b/src/data/queries/qaGroup.ts
@@ -1,0 +1,21 @@
+import { QueryFormatter } from 'next-drupal-query'
+import { queries } from '.'
+import { ParagraphQaGroup } from '@/types/drupal/paragraph'
+import { QaGroup } from '@/types/formatted/qaGroup'
+
+export const formatter: QueryFormatter<ParagraphQaGroup, QaGroup> = (
+  entity: ParagraphQaGroup
+) => {
+  return {
+    questions:
+      entity.field_q_as.map?.((item) => ({
+        title: item.title,
+        answer: queries.formatData('paragraph--wysiwyg', item.field_answer),
+      })) || [],
+    type: entity.type,
+    id: entity.id,
+    intro: entity.field_section_intro,
+    header: entity.field_section_header,
+    displayAccordion: entity.field_accordion_display,
+  }
+}
--- a/src/data/queries/qaGroup.ts
+++ b/src/data/queries/qaGroup.ts
@@ -1,6 +1,7 @@
 import { QueryFormatter } from 'next-drupal-query'
 import { ParagraphQaGroup } from '@/types/drupal/paragraph'
 import { QaGroup } from '@/types/formatted/qaGroup'
+import { QaGroupQa } from '@/types/formatted/qaGroup'
 import { formatParagraph } from '@/lib/drupal/paragraphs'
 
 export const formatter: QueryFormatter<ParagraphQaGroup, QaGroup> = (
@@ -15,10 +16,10 @@ export const formatter: QueryFormatter<ParagraphQaGroup, QaGroup> = (
       filteredQAs.map?.((item) => ({
         question: item.title,
         answers: [formatParagraph(item.field_answer)],
-        type: item.type,
+        type: item.type as QaGroupQa['type'],
         id: item.id,
       })) || [],
-    type: entity.type,
+    type: entity.type as QaGroup['type'],
     id: entity.id,
     entityId: entity.drupal_internal__id,
     intro: entity.field_section_intro || null,
--- a/src/data/queries/qaGroup.ts
+++ b/src/data/queries/qaGroup.ts
@@ -6,17 +6,23 @@ import { formatParagraph } from '@/lib/drupal/paragraphs'
 export const formatter: QueryFormatter<ParagraphQaGroup, QaGroup> = (
   entity: ParagraphQaGroup
 ) => {
+  // This handles filtering for field_q_as that are not published
+  const filteredQAs = entity.field_q_as.filter(
+    (item) => item.title || item.field_answer
+  )
   return {
     questions:
-      entity.field_q_as.map?.((item) => ({
-        title: item.title,
-        answer: formatParagraph(item.field_answer),
+      filteredQAs.map?.((item) => ({
+        question: item.title,
+        answers: [formatParagraph(item.field_answer)],
+        type: item.type,
+        id: item.id,
       })) || [],
     type: entity.type,
     id: entity.id,
     entityId: entity.drupal_internal__id,
-    intro: entity.field_section_intro ?? null,
-    header: entity.field_section_header ?? null,
-    displayAccordion: entity.field_accordion_display ?? false,
+    intro: entity.field_section_intro || null,
+    header: entity.field_section_header || null,
+    displayAccordion: entity.field_accordion_display || false,
   }
 }
--- a/src/data/queries/qaGroup.ts
+++ b/src/data/queries/qaGroup.ts
@@ -1,7 +1,7 @@
 import { QueryFormatter } from 'next-drupal-query'
-import { queries } from '.'
 import { ParagraphQaGroup } from '@/types/drupal/paragraph'
 import { QaGroup } from '@/types/formatted/qaGroup'
+import { formatParagraph } from '@/lib/drupal/paragraphs'
 
 export const formatter: QueryFormatter<ParagraphQaGroup, QaGroup> = (
   entity: ParagraphQaGroup
@@ -10,12 +10,13 @@ export const formatter: QueryFormatter<ParagraphQaGroup, QaGroup> = (
     questions:
       entity.field_q_as.map?.((item) => ({
         title: item.title,
-        answer: queries.formatData('paragraph--wysiwyg', item.field_answer),
+        answer: formatParagraph(item.field_answer),
       })) || [],
     type: entity.type,
     id: entity.id,
-    intro: entity.field_section_intro,
-    header: entity.field_section_header,
-    displayAccordion: entity.field_accordion_display,
+    entityId: entity.drupal_internal__id,
+    intro: entity.field_section_intro ?? null,
+    header: entity.field_section_header ?? null,
+    displayAccordion: entity.field_accordion_display ?? false,
   }
 }
--- a/src/lib/constants/resourceTypes.ts
+++ b/src/lib/constants/resourceTypes.ts
@@ -31,7 +31,7 @@ export const PARAGRAPH_RESOURCE_TYPES = {
   // STAFF_PROFILE: 'paragraph--staff_profile',
   TABLE: 'paragraph--table',
   WYSIWYG: 'paragraph--wysiwyg',
-  QA_PARAGRAPH: 'paragraph--q_a',
+  QA: 'paragraph--q_a',
   QA_SECTION: 'paragraph--q_a_section',
   QA_GROUP: 'paragraph--q_a_group',
 } as const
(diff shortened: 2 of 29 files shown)


tjheffner:
Nothing stands out to me in this PR, but I do notice that tugboat fails on some pages now:

https://tugboat.vfs.va.gov/65e9dfad9341203883c2165a
```
> Export encountered errors on following paths:
        /[[...slug]]: /resources/about-our-va-community-care-network-and-covered-services
        /[[...slug]]: /resources/creating-an-account-for-vagov
        /[[...slug]]: /resources/getting-urgent-care-at-va-or-in-network-community-providers
        /[[...slug]]: /resources/va-health-care-income-limits
```

Also, maybe move `Qa Paragraph` from `Components` down to the `Paragraphs` section of Storybook (happens in the label of the stories.tsx file itself)
Screenshot 2024-03-08 at 9 22 19 AM

src/lib/constants/resourceTypes.ts
```diff
@@ -31,6 +31,9 @@ export const PARAGRAPH_RESOURCE_TYPES = {
   // STAFF_PROFILE: 'paragraph--staff_profile',
   TABLE: 'paragraph--table',
   WYSIWYG: 'paragraph--wysiwyg',
+  QA_PARAGRAPH: 'paragraph--q_a',
```
ryguyk, line 34: I think I would prefer `QA` rather than `QA_PARAGRAPH`. 

We discussed naming a bit previously, so I apologize if this conflicts with anything I previously shared. But everything in here is a paragraph so it feels confusing to see `QA_PARAGRAPH` and then `QA_SECTION` and `QA_GROUP`. I actually wondered for a second if all of these were types of paragraphs.

src/data/queries/qaGroup.ts
```diff
@@ -0,0 +1,21 @@
...
+    questions:
+      entity.field_q_as.map?.((item) => ({
+        title: item.title,
+        answer: queries.formatData('paragraph--wysiwyg', item.field_answer),
+      })) || [],
+    type: entity.type,
+    id: entity.id,
+    intro: entity.field_section_intro,
+    header: entity.field_section_header,
+    displayAccordion: entity.field_accordion_display,
```
ryguyk, line 19: We want to include `id`, `type` and `entityId` for all paragraphs.

Then, the type definition `QaGroup` will need to be an intersection type with `PublishedParagraph` (see more commentary on the type-definition file).

src/types/formatted/qaGroup.ts
```diff
@@ -0,0 +1,15 @@
+import { Wysiwyg as FormattedWysiwyg } from './wysiwyg'
+
+type QaGroupQa = {
+  title: string
+  answer: FormattedWysiwyg
+}
+
+export type QaGroup = {
```
ryguyk, line 8: `PublishedParagraph & {...}`

src/types/formatted/qaSection.ts
```diff
@@ -0,0 +1,12 @@
...
+import { PublishedParagraph } from './publishedEntity'
+
+export type QaSection = PublishedParagraph & {
+  header: string
+  intro: string
+  displayAccordion: boolean
+  questions: FormattedParagraph[]
+  id: string
+  type: string
+  headerTagLevel?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6'
```
ryguyk, line 11: There's a type `HeadingLevel` for this defined in `src/types/formatted/headingElement.ts`

src/types/formatted/qaGroup.ts
ryguyk: We'll need to add all of these paragraph types to the `Paragraph` component in `src/templates/components/paragraph/index.tsx`

src/data/queries/qaGroup.ts
```diff
@@ -0,0 +1,21 @@
...
+import { QaGroup } from '@/types/formatted/qaGroup'
+
+export const formatter: QueryFormatter<ParagraphQaGroup, QaGroup> = (
+  entity: ParagraphQaGroup
+) => {
+  return {
+    questions:
+      entity.field_q_as.map?.((item) => ({
+        title: item.title,
+        answer: queries.formatData('paragraph--wysiwyg', item.field_answer),
```
ryguyk, line 13: It looks like `field_answer` can be `paragraph--rich_text_char_limit_1000`:

```
"field_answer": {
  "data": {
    "type": "paragraph--rich_text_char_limit_1000",
    "id": "4a228710-bd01-4107-ac29-dc8411400039",
    "meta": {
      "target_revision_id": 769238,
      "drupal_internal__target_id": 13976
    }
  },
  "links": {
    "related": {
      "href": "https://content-build-medc0xjkxm4jmpzxl3tfbcs7qcddsivh.ci.cms.va.gov/jsonapi/node/q_a/9299a086-e253-4ef4-a483-16057b7d43a0/field_answer?resourceVersion=id%3A697026"
    },
    "self": {
      "href": "https://content-build-medc0xjkxm4jmpzxl3tfbcs7qcddsivh.ci.cms.va.gov/jsonapi/node/q_a/9299a086-e253-4ef4-a483-16057b7d43a0/relationships/field_answer?resourceVersion=id%3A697026"
    }
  }
},
```

This is effectively the same thing as `paragraph--wysiwyg`, but perhaps we can just use `formatParagraph` here and take advantage of that abstraction and not having to define the type of paragraph at this layer.

ryguyk: I also wonder if we should just be formatting this entire thing using `queries.formatData('questionAnswer`, ...)`. But I'm not sure that query is properly set up yet.

ryguyk (changes requested):
I think this needs a few things in order to fit into the larger puzzle. Looks great overall!

tjheffner:
one page still has an error when tugboat builds:
```
> Export encountered errors on following paths:
        /[[...slug]]: /resources/creating-an-account-for-vagov
```

src/templates/components/qaSection/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+}: FormattedQaSection | FormattedQaGroup) {
+  const setHeaderh3 = header ? true : false
+  return (
+    <div data-template="paragraphs/q_a_section">
+      {header && <h2>{header}</h2>}
+      {intro && <p>{intro}</p>}
+      {displayAccordion ? (
+        <QaCollapsiblePanel questions={questions} />
+      ) : (
+        questions.map((questionObject: FormattedQaParagraph, index) => (
```
ryguyk, line 21: This feels interesting to me. Despite TypeScript not yelling at this, it feels like `FormattedQaParagraph` can't work here because `questions` can actually be `FormattedParagraph[]` or `QaGroupQa[]`. Since `FormattedParagraph` encompasses a number of different types, they won't all map to `FormattedQaParagraph`.

It seems to me that we need to type `questions` differently. I'm leaving some other comments in some other places to this end.

src/types/formatted/qaSection.ts
```diff
@@ -0,0 +1,12 @@
+import { FormattedParagraph } from '@/data/queries'
+import { PublishedParagraph } from './publishedEntity'
+import { HeadingLevel } from './headingElement'
+
+export type QaSection = PublishedParagraph & {
+  header: string
+  intro: string
+  displayAccordion: boolean
+  questions: FormattedParagraph[]
```
ryguyk, line 9: This can be one of two paragraph types:
image

I think we should type this accordingly.

Notably, it's a bit interesting that this can be a React Widget. I only see one example of this in the current content, and it's a "test" with "do not publish". But the IA allows for it, so I think we need to accommodate it. It's important to consider that this will likely change how the template works, though, since it will need to handle the possibility that it's a React Widget as well. I think this can be handled by using the general `<Paragraph ...>` component in the template rather than the more specific `<QaParagraph ...>` component.

src/types/formatted/qaGroup.ts
```diff
@@ -0,0 +1,16 @@
...
+
+type QaGroupQa = {
+  question: string
+  answers: FormattedParagraph[]
+  type: string
+  id: string
+}
+
+export type QaGroup = PublishedParagraph & {
+  questions: QaGroupQa[]
```
ryguyk, line 12: I'm thinking more and more that we should get the QA node query/formatter set up properly and then type `questions` as `QuestionAnswer`, since that is how things are set up in Drupal. I know we want to get this over the line, so perhaps this is follow-on work, but I just don't think it makes a ton of sense to define `QaGroupQa` as a sort of quick-and-dirty knockoff of the actual node that we'll have to have eventually anyway.

ryguyk (changes requested):

src/types/formatted/qaGroup.ts
tjheffner: agreed. I think we can make that a fast follow so that we can get Vet Centers unblocked and in, but this should be cleaned up.

src/types/formatted/qaSection.ts
```diff
@@ -0,0 +1,12 @@
...
+  type: 'paragraph--q_a_section' | 'paragraph--q_a_group'
+  headerTagLevel?: HeadingLevel
```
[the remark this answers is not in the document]
jtmst: Looking again, this was meant to be refactored out as unnecessary as it isnt currently used in content-builds latest templates either. will clean up

src/types/formatted/qaSection.ts
```diff
@@ -0,0 +1,12 @@
+import { FormattedParagraph } from '@/data/queries'
+import { PublishedParagraph } from './publishedEntity'
+import { HeadingLevel } from './headingElement'
+
+export type QaSection = PublishedParagraph & {
+  header: string
+  intro: string
+  displayAccordion: boolean
+  questions: FormattedParagraph[]
+  type: 'paragraph--q_a_section' | 'paragraph--q_a_group'
```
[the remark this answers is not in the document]
jtmst: will do

src/types/formatted/qaSection.ts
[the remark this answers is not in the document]
jtmst: With that out of the way we largely have similar data that is being shaped similarly by formaters.  the group formatter is currently being used in resources and support, and I will be able to implement the section formatted version with vet-centers.  if it proves to be an issue there I can look to just create separated templates, but i think this could work

src/types/formatted/qaSection.ts
jtmst: Agree with this,  will plug in paragraph component instead

src/types/formatted/qaGroup.ts
jtmst: Going to comment that QaGroupQa should be removed once QuestionAnswer is revisted.

src/templates/components/qaSection/index.tsx
jtmst: It will need to be formattedParagraph to accommodate the change below to <Paragraph>

src/types/formatted/qaSection.ts
ryguyk: Just a note that Josh and I discussed and agreed that this might be better remaining typed as `questions: FormattedParagraph[]`. There are currently only two paragraph types allowed, but there could theoretically be more added. Typing it more specifically offers some advantages in type safety, IDE type hinting, etc. but this feels like a good tradeoff in this case.

ryguyk (approved):
Nice work!

PR #456, opened 2024-03-15
VACMS-17523: Type cleanup

ryguyk:
## Description

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/17523
- Cleans up type definitions
- Small tweaks to bring other things more in line with architecture that surrounds these type definitions

## Testing done/QA Steps
- [x] Build w/out errors
- [x] Manually check pages of each built content type
  - [x] Story Listing: [/minneapolis-health-care/stories/](https://pr456-smurhj5z9pibvrniwtgtidibnnkugjjh.tugboat.vfs.va.gov/minneapolis-health-care/stories/)
  - [x] Story Listing subsequent page: [/minneapolis-health-care/stories/page-3/](https://pr456-smurhj5z9pibvrniwtgtidibnnkugjjh.tugboat.vfs.va.gov/minneapolis-health-care/stories/page-3/)
  - [x] News Story: [/minneapolis-health-care/stories/my-life-my-story-by-dave/](https://pr456-smurhj5z9pibvrniwtgtidibnnkugjjh.tugboat.vfs.va.gov/minneapolis-health-care/stories/my-life-my-story-by-dave/)
  - [x] Event Listing: [/minneapolis-health-care/events/](https://pr456-smurhj5z9pibvrniwtgtidibnnkugjjh.tugboat.vfs.va.gov/minneapolis-health-care/events/)
  - [x] Event: [/minneapolis-health-care/events/65635/](https://pr456-smurhj5z9pibvrniwtgtidibnnkugjjh.tugboat.vfs.va.gov/minneapolis-health-care/events/65635/)
  - [x] Resources and Support: [/resources/covid-19-testing-at-va/](https://pr456-smurhj5z9pibvrniwtgtidibnnkugjjh.tugboat.vfs.va.gov/resources/covid-19-testing-at-va/)

### DIFF
--- a/src/data/queries/index.ts
+++ b/src/data/queries/index.ts
@@ -44,52 +44,57 @@ import * as HealthServices from './healthServices'
 import {
   ResourceType,
   ParagraphResourceType,
+  RESOURCE_TYPES,
+  PARAGRAPH_RESOURCE_TYPES,
+  PageResourceType,
 } from '@/lib/constants/resourceTypes'
 
 export const QUERIES_MAP = {
   // Standard Drupal entity data queries
   // Nodes
-  'node--news_story': NewsStory,
-  'node--news_story--teaser': NewsStoryTeaser,
-  'node--story_listing': StoryListing,
-  'node--q_a': QuestionAnswer,
-  'node--event': Event,
-  'node--event--teaser': EventTeaser,
-  'node--event_listing': EventListing,
-  'node--person_profile': PersonProfile,
-  'node--landing_page': BenefitsHub, // "Benefits Hub Landing Page"
-  'node--support_service': SupportServices,
-  'node--support_resources_detail_page': ResourcesSupport,
-  'node--vet_center': VetCenter,
-  'node--vet_center_facility_health_servi': HealthServices,
+  [RESOURCE_TYPES.STORY]: NewsStory,
+  [`${RESOURCE_TYPES.STORY}--teaser` as const]: NewsStoryTeaser,
+  [RESOURCE_TYPES.STORY_LISTING]: StoryListing,
+  [RESOURCE_TYPES.QA]: QuestionAnswer,
+  [RESOURCE_TYPES.EVENT]: Event,
+  [`${RESOURCE_TYPES.EVENT}--teaser` as const]: EventTeaser,
+  [RESOURCE_TYPES.EVENT_LISTING]: EventListing,
+  [RESOURCE_TYPES.PERSON_PROFILE]: PersonProfile,
+  [RESOURCE_TYPES.BENEFITS_HUB]: BenefitsHub, // "Benefits Hub Landing Page"
+  [RESOURCE_TYPES.SUPPORT_SERVICES]: SupportServices,
+  [RESOURCE_TYPES.RESOURCES_SUPPORT]: ResourcesSupport,
+  [RESOURCE_TYPES.VET_CENTER]: VetCenter,
+  [RESOURCE_TYPES.HEALTH_SERVICES]: HealthServices,
 
   // Paragraphs
-  'paragraph--alert': Alert,
-  'paragraph--alert_single': AlertSingle,
-  'paragraph--audience_topics': AudienceTopics,
-  'paragraph--basic_accordion': Accordion,
-  'paragraph--button': Button,
-  'paragraph--collapsible_panel': CollapsiblePanel,
-  'paragraph--collapsible_panel_item': CollapsiblePanelItem,
-  'paragraph--contact_information': ContactInfo,
-  'paragraph--email_contact': EmailContact,
-  'paragraph--expandable_text': ExpandableText,
-  'paragraph--featured_content': FeaturedContent,
-  'paragraph--link_teaser': LinkTeaser,
-  'paragraph--non_reusable_alert': AlertNonReusable,
-  'paragraph--number_callout': NumberCallout,
-  'paragraph--phone_number': PhoneNumber,
-  'paragraph--process': ProcessList,
-  'paragraph--q_a': QaParagraph,
-  'paragraph--q_a_section': QaSection,
-  'paragraph--q_a_group': QaGroup,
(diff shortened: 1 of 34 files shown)


src/data/queries/index.ts
```diff
@@ -100,20 +105,53 @@ export const QUERIES_MAP = {
...
 
-// Type representing all possible object shapes returned from querying and formatting Drupal data.
-// E.g. StoryListingType | NewsStoryType | (other future resource type)
-// E.g. Accordion | Button (other paragraphs)
-// Type constructed by:
-//  1. Consider all ResourceType or ParagraphResourceType types
-//  2. Take subset of those types that have a key in QUERIES_MAP
-//  3. Map that subset of keys to their respective values, which are modules for querying data
-//  4. Within each of those modules, grab the return type of the `formatter` function
+// All resource types that have a `data` function defined
```
tjheffner, line 108: I really like this splitting up & individual comments for everything. A little easier to follow.

src/data/queries/personProfile.ts
```diff
@@ -9,41 +9,40 @@ import { DrupalJsonApiParams } from 'drupal-jsonapi-params'
...
 
-// Define the query params for fetching node--news_story.
-export const params: QueryParams<null> = () => {
-  return new DrupalJsonApiParams().addInclude([
-    'field_media.image',
-    'field_office',
-    'field_complete_biography',
-  ])
-}
+// // Define the query params for fetching node--person_profile.
```
tjheffner, line 15: Is this query & data loader commented out only because we haven't officially gone through and (re-)built this page type?

src/data/queries/personProfile.ts
```diff
@@ -9,41 +9,40 @@ import { DrupalJsonApiParams } from 'drupal-jsonapi-params'
...
+//     params: params().getQueryObject(),
+//   })
+//   return entities
+// }
 
-export const formatter: QueryFormatter<NodePersonProfile, PersonProfile[]> = (
-  entities: NodePersonProfile
+export const formatter: QueryFormatter<NodePersonProfile, PersonProfile> = (
+  entity: NodePersonProfile
 ) => {
```
tjheffner, line 41: much better here, thank you

src/lib/drupal/listingPages.test.ts
```diff
@@ -35,7 +35,7 @@ describe('isSinglePageListingResourceType', () => {
   })
 
   test('should return false when not single-page listing resource type', () => {
-    const storyListingResult = isListingResourceType(
+    const storyListingResult = isSinglePageListingResourceType(
```
tjheffner, line 38: thank you for fixing this test

src/lib/drupal/lovell/tests/utils.test.ts
```diff
@@ -39,11 +39,12 @@ describe('isLovellResourceType', () => {
     expect(storyResult).toBe(true)
   })
 
-  // Currently, all resource types are Lovell resource types (Story, Story Listing)
```
tjheffner, line 42: one less commented test warning 🎉

tjheffner (approved):
A lot of really nice cleanup in here, thank you!


This type of thing in `queries/index.tsx` is the only thing I'm not immediately loving, but I see why it is done and good for consistency / less chance of typos as those strings get littered throughout. So thank you for doing it.
```js
 [RESOURCE_TYPES.STORY]: NewsStory,
  [`${RESOURCE_TYPES.STORY}--teaser` as const]: NewsStoryTeaser,
```

src/data/queries/personProfile.ts
ryguyk: Yes, mostly. I originally commented it out because it (along with a few other resource types) weren't fully aligned with the others that we are building (e.g. missing properties). As a result, type errors were being raised in places referencing `FormattedResource` (e.g. `[[...slug]].tsx` default export - the template). So, initially, I started to try to fix these files, and `personProfile` was the first one. I realized things were not set up correctly with the query or the formatter, so I commented out the query (not needed currently) and fixed the formatter.

But, ultimately, I ended up introducing `FormattedPageResource` to differentiate between nodes that generate pages and nodes that are only entity references (for now, at least). This made it unnecessary to update all the other files accordingly, and allowed me to just "move on". There are still some resources that are in the same situation as `personProfile`, but they weren't addressed here. Specifically, the query is likely unnecessary (it can probably just be deleted rather than commented in `personProfile`) because, although it's a node, we aren't going to query it individually. In this way, it's more like a paragraph than a page-generating node.

ryguyk:
> A lot of really nice cleanup in here, thank you!
> 
> This type of thing in `queries/index.tsx` is the only thing I'm not immediately loving, but I see why it is done and good for consistency / less chance of typos as those strings get littered throughout. So thank you for doing it.
> 
> ```js
>  [RESOURCE_TYPES.STORY]: NewsStory,
>   [`${RESOURCE_TYPES.STORY}--teaser` as const]: NewsStoryTeaser,
> ```

I considered this for a while. Ultimately, I think we want to use the enumeration/lookup table anywhere we are referencing these resource-type names. In this case, especially, this will prevent the addition of a query without the addition of an entry in the lookup table.

Notably, the `...-teaser` entries are interesting because these are not actually resource types. The resource type of `node--news_story--teaser`, for example, is `node--news_story`, but we have a different formatter. I wondered about having a single query file for these and somehow passing in an option to indicate which format was desired, but it looks like that is not possible (the `formatter` function doesn't accept options). So, rather than make a non-resource entry in the lookup table, this was the more clean approach, in my opinion.

src/data/queries/personProfile.ts
tjheffner: Thank you for the explanation, that does make sense. I do want to note that `personProfile` will generate pages at some point (though to your point, there are also profile paragraphs that do not generate pages). 
e.g. https://www.va.gov/butler-health-care/staff-profiles/sharon-coyle/

tjheffner (approved):

PR #463, opened 2024-03-20
Update sitemap a11y scan

tjheffner:
## Description
Relates to #[16117](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/16117).

## Testing done

Running test locally with all the options via CLI, e.g. `BASE_URL=https://va.gov USE_PROXY=false PW_WIDTH_VALUE=768 PW_HEIGHT_VALUE=720 PW_BROWSER_VALUE=firefox SEGMENT_INDEX=0 yarn playwright test --project=a11y`

## Screenshots

Screenshot 2024-04-05 at 11 18 40 AM

## QA steps

1. Go to the a11y workflow section on GH: https://github.com/department-of-veterans-affairs/next-build/actions/workflows/a11y.yml
2. Look at the latest run to see it is green.
3. Download the `a11y-scan-1` artifact and open it to confirm violations are listed by URL with a valid JSON syntax. The browser name and viewport size should be included with each URL listed.
4. Re-run the latest job and see it passes. The run should take around 20 minutes.
5. Repeat steps 2 and 3 to see they still pass.


alexfinnarn:
This branch runs locally with `BASE_URL=https://va.gov USE_PROXY=false yarn test:playwright:a11y` and completes five segments.

Tests running:
Screenshot 2024-03-20 at 1 36 11 PM

HTML report home page:
Screenshot 2024-03-20 at 1 40 23 PM

HTML report for specific test:
Screenshot 2024-03-20 at 1 41 21 PM

alexfinnarn (dismissed):
LGTM! I agree with not processing the sitemap on each request but this works locally now and can be modified later.

.github/workflows/a11y.yml
```diff
@@ -0,0 +1,158 @@
...
+
+      - name: Download blob reports from GitHub Actions Artifacts
+        uses: actions/download-artifact@v4
+        with:
+          pattern: failed-pages-segment-*
+          merge-multiple: true
+
+      - name: Merge reports in valid JSON format
+        run: |
+          if [ ! -f failed-pages-segment-*.json ]; then
```
tjheffner, line 142: the last workflow didn't find any failed pages here, but several links that 502'd were included in the scan upload instead. I'd expect to see pages that 502 in this report instead

```json
  {
    "url": "http://www.va.gov/central-ohio-health-care/work-with-us/internships-and-fellowships/pgy-1-pharmacy-residency-program/",
    "browserName": "webkit",
    "viewportSize": {
      "width": 1920,
      "height": 1440
    },
    "violations": [
      {
        "id": "html-has-lang",
        "impact": "serious",
        "tags": [
          "cat.language",
          "wcag2a",
          "wcag311",
          "TTv5",
          "TT11.a",
          "EN-301-549",
          "EN-9.3.1.1",
          "ACT"
        ],
        "description": "Ensures every HTML document has a lang attribute",
        "help": "<html> element must have a lang attribute",
        "helpUrl": "https://dequeuniversity.com/rules/axe/4.8/html-has-lang?application=playwright",
        "nodes": [
          {
            "any": [
              {
                "id": "has-lang",
                "data": {
                  "messageKey": "noLang"
                },
                "relatedNodes": [],
                "impact": "serious",
                "message": "The <html> element does not have a lang attribute"
              }
            ],
            "all": [],
            "none": [],
            "impact": "serious",
            "html": "<html><head><title>502 Bad Gateway</title></head>\n<body>\n<center><h1>502 Bad Gateway</h1></center>\n\n\n</body></html>",
            "target": [
              "html"
            ],
            "failureSummary": "Fix any of the following:\n  The <html> element does not have a lang attribute"
          }
        ]
      }
    ]
  },
```

playwright/tests/a11y.spec.js
```diff
@@ -1,66 +1,145 @@
...
+        `failed-pages-segment-${segmentNumber}.json`,
+        trimmedFailedPages
       )
-    })
-  }
+    }
+  })
 })
+
+function splitArray(array, numChunks) {
```
tjheffner, line 132: i'm curious why this was needed versus re-using the `splitPagesIntoBatches` and `getPagesSlice` from the sitemapLocations utility

playwright/tests/a11y.spec.js
```diff
@@ -1,66 +1,145 @@
...
+          .exclude('footer')
+          .analyze()
+
+        console.log('page violations:', accessibilityScanResults.violations)
+        if (accessibilityScanResults.violations.length > 0) {
+          scanResults.push({
+            url: pageUrl,
+            browserName,
+            viewportSize,
+            violations: accessibilityScanResults.violations,
```
tjheffner, line 89: i think this object could be slightly more helpful / trimmed down some, with something like:

```js
scanResults.push({
             url: pageUrl,
             browserName,
             viewportSize,
             violations: accessibilityScanResults.violations.map((v) => {
                 id: v.id,
                 description: v.description,
                 impact: v.impact,
                 nodes: `${v.nodes.map((node) => node.html).join(', ')}`
             })
})
```

playwright/tests/a11y.spec.js
```diff
@@ -1,66 +1,145 @@
...
-    }) => {
-      await runA11yTestsForPages(
-        pageSegments[i],
-        `Segment ${i + 1}`,
-        page,
-        makeAxeBuilder
+    if (scanResults.length > 0) {
+      // Remove root array from scanResults so we can merge results more cleanly.
+      const trimmedScanResults =
+        JSON.stringify(scanResults, null, 2).replace(/^\[|]$/g, '') +
```
tjheffner, line 111: this is hard to parse at a glance. is it just I'm curious why we need to do this / add a trailing comma... Let's keep the individual scans as valid JSON imo.

something like
`jq -s '.' *.json | jq  'flatten' > merged-report.json` will take all json files and merge them into one array of objects without any fussing with removing brackets or adding commas to the individual files.

.github/workflows/a11y.yml
```diff
@@ -0,0 +1,158 @@
+name: A11y Tests
+on:
+  pull_request:
```
tjheffner, line 3: once we have all the kinks worked out, let's update this to run on a schedule like the`broken-links-check.yml` workflow

Something like:
```
on:
  # Manual trigger
  workflow_dispatch:
  # Schedule; currently once nightly, at 3:20 UTC.
  schedule:
    - cron: 20 3 * * *
```
(I've adjusted the hours from the broken link schedule to offset load on the runners)

.github/workflows/a11y.yml
```diff
@@ -0,0 +1,158 @@
...
+      - name: Build proxy-fetcher dist
+        run: |
+          yarn tsc -b ./packages/proxy-fetcher/tsconfig.json
+
+      - name: Run Playwright tests
+        run: |
+          SEGMENT_INDEX=${{ matrix.shardIndex }} yarn playwright test --project=a11y
+
+      - name: Upload scan report to GitHub Actions Artifacts
+        if: always()
```
tjheffner, line 68: These uploads should probably check for existence of a file rather than `always()`.

The summary page has a ton of warnings that a file doesn't exist, so nothing was uploaded: https://github.com/department-of-veterans-affairs/next-build/actions/runs/8560535467?pr=463

READMEs/testing.md
```diff
@@ -75,12 +75,136 @@ This project can be tested for a11y compliance in several ways:
...
-- A full site scan of all urls known to next-build (generated in the sitemap) using Playwright: `yarn test:playwright:a11y`
-
-The full scan will run (TKTK: some cadence, weekly?) in CI.
-You can run it manually after generating the sitemap with a few steps:
-
-1. `yarn export` to generate the static pages for the site
-2. `yarn build:sitemap` to generate the sitemap for pages from step 1.
-3. `yarn export:serve` to host the static pages locally
-4. `yarn test:playwright:a11y` to run the scan. This runs `playwright/tests/a11y.spec.js` which loops over the sitemap and tests each page individually using `@axe-core/playwright`.
+- A full site scan of all urls known to next-build (generated in the sitemap) using Playwright:
```
tjheffner, line 78: I love all of this. thank you.

playwright/tests/a11y.spec.js
```diff
@@ -1,66 +1,145 @@
...
+        console.log('skipping scan of excluded page:', pageUrl)
+        continue
+      }
+
+      try {
+        console.log('testing page:', pageUrl)
+        await page.goto(pageUrl)
+
+        // @todo The shared "makeAxeBuilder" never reports errors for whatever reason so not using for now.
+        // const accessibilityScanResults = await makeAxeBuilder({ page }).analyze()
```
tjheffner, line 73: just confirming that this is still true for you? it has reported errors previously in the individual spec tests

playwright/tests/a11y.spec.js
alexfinnarn: I don't fully remember but I was looking to split an array into equal chunks and the `splitPagesIntoBatches` wasn't giving me as equal of chunks. But could of been some other issue I was facing...

playwright/tests/a11y.spec.js
alexfinnarn: What does the accessibility person reviewing the report want to see? I'm all for making it more helpful for them to review.

playwright/tests/a11y.spec.js
alexfinnarn: Okay if that works, I'm fine with it.

.github/workflows/a11y.yml
tjheffner: i believe it means it 502'd during the scan

.github/workflows/a11y.yml
```diff
@@ -0,0 +1,150 @@
...
+          pattern: segment-*
+          merge-multiple: true
+
+      - name: Merge reports in valid JSON format
+        run: |
+          if [ ! -f segment-*.json ]; then
+              echo "No segment-*.json files found."
+              exit 0
+          fi
+          jq -s '.' *.json | jq 'flatten' > all-segments.json
```
tjheffner, line 108: this probably needs to be `segment-*.json` as the full greedy glob pulls in `package.json` & `tsconfig.json` too.

playwright/tests/a11y.spec.js
```diff
@@ -1,66 +1,148 @@
...
+  'http://www.va.gov/track-claims/',
+  'http://www.va.gov/burials-and-memorials-v2/application/530/',
+  'http://www.va.gov/profile/',
+  'http://www.va.gov/my-va/',
+  'http://www.va.gov/disability/view-disability-rating/rating/',
+  'http://www.va.gov/view-change-dependents/view/',
+  'http://www.va.gov/supporting-forms-for-claims/substitute-claimant-form-21P-0847/',
+  'http://www.va.gov/supporting-forms-for-claims/alternate-signer-form-21-0972/',
+  'http://www.va.gov/careers-employment/education-and-career-counseling/apply-career-guidance-form-25-8832/',
+]
```
tjheffner, line 42: let's also add:
```
'http://www.va.gov/js/smartbanner/README/',
'http://www.va.gov/js/smartbanner/LICENSE/'
```

why these two are in the sitemap in the first place is a different story, but let's not scan them. they always report non-actionable errors

playwright/tests/a11y.spec.js
tjheffner: the content-build reports have expired, so can't see the actual logs... but the file itself has this:

https://github.com/department-of-veterans-affairs/content-build/blob/main/src/platform/testing/e2e/cypress/support/commands/axeCheck.js#L22

so the violations are being trimmed somewhat. The example I shared above was basically modified from the original output of this scan prior to your updates:

```js
    if (accessibilityScanResults.violations.length > 0) {
      accessibilityScanResults.violations.forEach((violation) => {
        console.log(`URL: ${pageUrl}`)
        console.log(`Violation: ${violation.id}`)
        console.log(`Description: ${violation.description}`)
        console.log(`Impact: ${violation.impact}`)
        console.log(
          `Nodes: ${violation.nodes.map((node) => node.html).join(', ')}`
        )
      })

      a11yFailures.push({
        url: pageUrl,
        violations: accessibilityScanResults.violations,
      })
```

I think there's a happy medium in there somewhere... the default violation output includes a lot of stuff that I think isn't particularly relevant or helpful.

playwright/tests/a11y.spec.js
alexfinnarn: I added the fixture back and the report shows scan results. So the fixture looks to be working.

playwright/tests/a11y.spec.js
alexfinnarn: Updated this part of the code to have a tidier report basically in the same shape.

playwright/tests/a11y.spec.js
alexfinnarn: I deleted that function I made and used `splitPagesIntoBatches` instead.

.github/workflows/a11y.yml
alexfinnarn: yep, I didn't see the 502 in the HTML key till now...I think maybe the HTML could be checked for a string of something like `<title>502 Bad Gateway</title>` but I didn't add that yet. We could do all kinds of checks with the results and modify the report to include/exclude/change output.

.github/workflows/a11y.yml
alexfinnarn: I added a check using `if: hashFiles('segment-${{ matrix.shardIndex }}.json') != ''` on the step, but I think it is only for whole jobs and not steps in jobs. So I went with an option to ignore if no files are found instead of printing a warning...but now GH Actions is having problems so that run is stuck in queued: https://www.githubstatus.com/incidents/xcjgp37zx5dt

I think I've gotten to all the comments in the review except the change of schedule, but I could have missed one .

.github/workflows/a11y.yml
tjheffner: @alexfinnarn everything LGTM, let's bump this to be scheduled and we can merge

tjheffner (commented):
there's one false positive with a 502 in the latest run, but otherwise everything LGTM. i'd like to get this merged and we can iterate as needed on the actual results. great work on this!

alexfinnarn (approved):

PR #466, opened 2024-03-22
VACMS-17582: Rearchitects query parameters

ryguyk:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/17582

There was this `TODO` in [src/data/queries/alertSingle.ts](https://github.com/department-of-veterans-affairs/next-build/blob/a10b0ff1f9bab4e796d6774b073c473cd2370774/src/data/queries/alertSingle.ts#L11-L57). This PR implements a solution to this.

```
// TODO:
//
// EDIT:
// It seems *maybe* that next-drupal-query is designed to handle this already.
// You can pass an `id` to queries.getParams().
// E.g.
// queries.getParams('paragraph--alert_single'))
//
// We don't need this here:
// export const params: QueryParams<null> = () => {
//   return new DrupalJsonApiParams()
//     .addInclude([
//       'field_alert_block_reference',
//       'field_alert_block_reference.field_alert_content',
//       'field_alert_non_reusable_ref',
//       'field_alert_non_reusable_ref.field_va_paragraphs',
//     ])
// }
//
// ...but it seems like we might want to define
// paragraph-specific `includes` within the paragraph's
// query file (this file), and then somehow reference these in
// the node query when we need to include fields of a specific
// paragraph type.
//
// E.g.
// export const include = [
//   'field_alert_block_reference',
//   'field_alert_block_reference.field_alert_content',
//   'field_alert_non_reusable_ref',
//   'field_alert_non_reusable_ref.field_va_paragraphs',
// ]
//
// /* Some util file*/
// export const getIncludedSubFields = (
//   parentField: string,
//   subFields: string[]
// ) => [parentField, ...subFields.map((subField) => `${parentField}.${subField}`)]
//
// /* src/data/queries/resourcesSupport.ts */
// import { include } from '@/data/queries/alertSingle'
// export const params: QueryParams<null> = () => {
//   return new DrupalJsonApiParams().addInclude([
//     ...getIncludedSubFields('field_alert_single', include)
//   ])
// }
```

Briefly, the goal is to remove the need for engineers writing page-level queries to need to understand the inner details of content that is included. For example:

A Resources-and-Support page has a field `field_alert_single`. This field references a `paragraph--alert_single`. Prior to this PR, the query to build this type of page needed to include this:
```
    'field_alert_single',
    'field_alert_single.field_alert_block_reference',
    'field_alert_single.field_alert_block_reference.field_alert_content',
    'field_alert_single.field_alert_non_reusable_ref',
    'field_alert_single.field_alert_non_reusable_ref.field_va_paragraphs',
```

But all those included fields are really an implementation detail of the `paragraph`, so requiring the R&S node query to know these details is not good architecture, and puts undue burden on the engineers tasked with building these page-level queries like R&S. This isn't even to mention, if that paragraph structure ever changed, we'd have to update _every_ query that repeated this structure. So, now, after this PR, this is what can be used instead:

```
    ...getNestedIncludes(
      'field_alert_single',
      PARAGRAPH_RESOURCE_TYPES.ALERT_SINGLE
    ),
```

Now, the page-level node query (R&S in this case) doesn't need to know anything about the paragraph. It can simply say, "give me the data for this paragraph found at this field."

### DIFF
--- a/src/lib/utils/queries.ts
+++ b/src/lib/utils/queries.ts
@@ -1,0 +1,20 @@
+import { queries, ParamsType } from '@/data/queries'
+
+export const getNestedIncludes = (
+  fieldName: string,
+  resourceType: ParamsType
+) => {
+  const includedFields = queries
+    .getParams(resourceType)
+    .getQueryObject()
+    .include.split(',')
+
+  return fieldName
+    ? [
+        fieldName,
+        ...includedFields.map(
+          (includedField) => `${fieldName}.${includedField}`
+        ),
+      ]
+    : includedFields
+}
--- a/src/lib/utils/queries.ts
+++ b/src/lib/utils/queries.ts
@@ -2,12 +2,15 @@ import { queries, ParamsType } from '@/data/queries'
 
 export const getNestedIncludes = (
   fieldName: string,
-  resourceType: ParamsType
+  resourceType: ParamsType | ParamsType[]
 ) => {
-  const includedFields = queries
-    .getParams(resourceType)
-    .getQueryObject()
-    .include.split(',')
+  const resourceTypes = Array.isArray(resourceType)
+    ? resourceType
+    : [resourceType]
+
+  const includedFields = resourceTypes.flatMap((resourceType) =>
+    queries.getParams(resourceType).getQueryObject().include.split(',')
+  )
 
   return fieldName
     ? [
--- a/src/data/queries/alertSingle.ts
+++ b/src/data/queries/alertSingle.ts
@@ -1,4 +1,5 @@
-import { QueryFormatter } from 'next-drupal-query'
+import { QueryFormatter, QueryParams } from 'next-drupal-query'
+import { DrupalJsonApiParams } from 'drupal-jsonapi-params'
 import { ParagraphAlertSingle } from '@/types/drupal/paragraph'
 import {
   AlertSingle,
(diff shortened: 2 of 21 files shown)


src/lib/utils/queries.ts
```diff
@@ -0,0 +1,22 @@
+import { queries, ParamsType } from '@/data/queries'
+
+jest.mock('')
```
tjheffner, line 3: this jest mock feels out-of-place here

ryguyk: 😂 Just a bit. I'm not sure how that got there. I had to introduce a mock in the test file for this code, so it's not a complete shock, but I clearly started typing in the wrong place.

jtmst (approved):
This will make some of these queries much cleaner. Non-blocking but I'm curious if expanding this to recursively include additional nested objects like we see in field_contact_information.field_benefit_hub_contacts.field_support_services, or if that is impractical

timcosgrove:
> This will make some of these queries much cleaner. Non-blocking but I'm curious if expanding this to recursively include additional nested objects like we see in field_contact_information.field_benefit_hub_contacts.field_support_services, or if that is impractical

Ideally I think it should do this, but as Josh says it may be impractical.

ryguyk:
@jtmst Very good question, and the answer is yes. I tried to get as much of this as possible, but some places were not as immediately workable as others. In the example you provided, there was some weirdness around how the benefit hubs query/formatter is working (it's a "teaser" only), so I initially skipped so as to not get slowed by those details. Your question prompted me to return to it with fresh eyes, and I was able to make things _mostly_ better. Our benefit hubs formatter is still an interesting one, and I've added a comment to the contact-info formatter that explains this:
```
    //TODO:
    // This should likely be: `queries.formatData(RESOURCE_TYPES.BENEFITS_HUB, entity.field_benefit_hub_contacts)
    // since `entity.field_benefit_hub_contacts` will be a reference to a Benefits Hub Landing Page,
    // but the formatter for Benefits Hub Landing Pages is currently just a "partial"/"teaser" formatter
    // that doesn't include everything we'd need. That likely needs to be broken into  "full" and "teaser"
    // files so we have a formatter for each.
    //
    // For now, we can drill down into `field_support_services` and pass that to the formatter for
    // RESOURCE_TYPES.SUPPORT_SERVICE.
    benefitHubContacts:
      entity.field_benefit_hub_contacts?.field_support_services.map(
        (supportService) =>
          queries.formatData(RESOURCE_TYPES.SUPPORT_SERVICES, supportService)
      ),
```

@timcosgrove 
> Can this recurse?

Yes! You can use `getNestedIncludes` to call a query-param generator that itself calls `getNestedIncludes`. The result will be built from the "bottom up".

jtmst (approved):
Looks great!

src/lib/utils/queries.ts
```diff
@@ -0,0 +1,23 @@
+import { queries, ParamsType } from '@/data/queries'
+
+export const getNestedIncludes = (
```
tjheffner, line 3: this is a very elegant little helper, thank you!

tjheffner (approved):

PR #487, opened 2024-04-03
VACMS-17551: quiet pRetry logging unless DEBUG flag is present

ryguyk:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/17551

- Adds DEBUG flag as CLI option
  - E.g. `yarn export --DEBUG` or `yarn export -d`
- Adjusts pRetry to only log failed fetch on last attempt unless debug flag is set
  - If `--DEBUG` is present, log all failed requests, as has been the case
  - If `--DEBUG` is not present, the default value is `false`, so only log on the last failure (6th attempt, currently)
- Brings USE_REDIS and SSG in line with DEBUG to be a flag rather than require true/false string on CLI
  - Previously, we were setting, for example, `--USE_REDIS=true` and `--SSG=false`
    - We chose this because when the value gets passed from the env-loader into the main application code, the boolean values would become strings anyway, so we needed to ensure that our application code was checking for strings rather than booleans.
    - But this requirement for application code can remain while at the same time allowing for the CLI values to be boolean (understanding that they will be converted to strings)
  - This PR adds `--DEBUG` as a boolean option, so it was a good time to clean up the others and align them all to what seems like the better approach.

## Testing done/QA
- [ ] `yarn export` or `yarn build` should never log failed first, second, etc. attempts.
- [ ] `yarn export` should build all pages successfully, thus proving that the changes to `SSG` flag handling work as expected.
- [ ] Changes were made to `docker-run.sh` and github workflow files `archive.yml` and `content-release.yml` to adjust the usage of the CLI options `--USE_REDIS`. These changes should be reviewed by @jschmidt-civicactions or @tjheffner. Note that `--USE_REDIS` defaults to `false` (as do all boolean CLI options). It will always be `false` (or `"false"`) unless `--USE_REDIS` is explicitly added.


tjheffner:
You may need to pass `--USE_REDIS` to the tugboat build explicitly now (line 75 in `.tugboat/config.yml`), if it is defaulting to false:

```
2024-04-03T17:04:04.781Z - 660d8beab4e1649ad167c2cd# /bin/sh -c NODE_EXTRA_CA_CERTS=/etc/ssl/certs/ca-certificates.crt yarn export
Using environment variables from: envs/.env.tugboat
   Linting and checking validity of types ...

./src/lib/drupal/lovell/tests/staticProps.test.ts
46:1  Warning: Some tests seem to be commented  jest/no-commented-out-tests
47:1  Warning: Some tests seem to be commented  jest/no-commented-out-tests
48:1  Warning: Some tests seem to be commented  jest/no-commented-out-tests
88:1  Warning: Some tests seem to be commented  jest/no-commented-out-tests
128:1  Warning: Some tests seem to be commented  jest/no-commented-out-tests
162:1  Warning: Some tests seem to be commented  jest/no-commented-out-tests

info  - Need to disable some ESLint rules? Learn more here: https://nextjs.org/docs/basic-features/eslint#disabling-rules
   Creating an optimized production build ...
Browserslist: caniuse-lite is outdated. Please run:
  npx browserslist@latest --update-db
  Why you should do it regularly: https://github.com/browserslist/browserslist#browsers-data-updating
 ✓ Compiled successfully
   Collecting page data ...
Redis connected
Redis connected
Redis ready!
Redis connected
Redis ready!
Redis ready!
Redis connected
Redis ready!
JsonApiErrors: Bad Gateway
    at /var/lib/tugboat/node_modules/next-drupal/dist/index.js:2550:19
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  errors: 'Bad Gateway',
  statusCode: 502
}

> Build error occurred
Error: Failed to collect page data for /[[...slug]]
    at /var/lib/tugboat/node_modules/next/dist/build/utils.js:1171:15
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  type: 'Error'
}
2024-04-03T17:04:30.782Z - Command Failed (Tugboat Error 1064): Exit code: 1; Command: NODE_EXTRA_CA_CERTS=/etc/ssl/certs/ca-certificates.crt yarn export

```

ryguyk:
> You may need to pass `--USE_REDIS` to the tugboat build explicitly now (line 75 in `.tugboat/config.yml`), if it is defaulting to false:

Where is this value getting used, though? From what I can tell, the application code needs `process.env.USE_REDIS` to explicitly be `"true"` everywhere it uses it. If it wasn't being set on line 75 in `.tugboat/config.yml`, how was it working? Is this flag being used somewhere outside our application code?

tjheffner:
> Where is this value getting used, though? From what I can tell, the application code needs process.env.USE_REDIS to explicitly be "true" everywhere it uses it. If it wasn't being set on line 75 in .tugboat/config.yml, how was it working? Is this flag being used somewhere outside our application code?

I don't think it is. AFAIK, it's only being checked inside of drupalClient:

```js
cache:
    process.env.USE_REDIS === 'true'
      ? redisCache(createRedisClient(process.env.REDIS_URL))
      : null,
```

We were explicitly setting this to true in package json before, so it would be used on tugboat: 
`"export": "yarn build:env-loader && node ./scripts/yarn/export.js --USE_REDIS=true",`


This PR updates that to `--USE_REDIS`, is the boolean not being converted to string as expected? Or should the check in drupalClient be updated to just `process.env.USE_REDIS ? redisCache(...) : null`?

ryguyk:
@tjheffner I found the issue. 

`USE_REDIS` was getting set just fine. In fact, we could see in the log a number of occurrences of `Redis connected`.

The issue can probably best be described by just pasting the code that needed to be changed. The error was not being thrown correctly, and it needed to be to have the retry mechanism work properly:

BEFORE:
```
      if (!response.ok && logFailedRequest) {
        /*eslint-disable-next-line*/
        console.log(
          `Failed request (Attempt ${attempt} of ${retryCount + 1}): ${
            response.url
          }`
        )
        throw new Error('Failed request')
      }
```

AFTER:
```
      if (!response.ok) {
        if (logFailedRequest) {
          /*eslint-disable-next-line*/
          console.log(
            `Failed request (Attempt ${attempt} of ${retryCount + 1}): ${
              response.url
            }`
          )
        }
        throw new Error('Failed request')
      }
```

ryguyk:
Note that this quiets the logs with regard to the retry mechanism, but we are still getting quite a bit of messages about page data size:

```
Warning: data for page "/[[...slug]]" (path "/charleston-health-care/stories/page-14/") is 77 kB which exceeds the threshold of 550 B, this amount of data can reduce performance.
See more info here: https://nextjs.org/docs/messages/large-page-data
Warning: data for page "/[[...slug]]" (path "/mountain-home-health-care/stories/page-2/") is 74.7 kB which exceeds the threshold of 550 B, this amount of data can reduce performance.
See more info here: https://nextjs.org/docs/messages/large-page-data
Warning: data for page "/[[...slug]]" (path "/tennessee-valley-health-care/stories/page-10/") is 507 kB which exceeds the threshold of 550 B, this amount of data can reduce performance.
See more info here: https://nextjs.org/docs/messages/large-page-data
Warning: data for page "/[[...slug]]" (path "/washington-dc-health-care/stories/page-10/") is 556 kB which exceeds the threshold of 550 B, this amount of data can reduce performance.
See more info here: https://nextjs.org/docs/messages/large-page-data
Warning: data for page "/[[...slug]]" (path "/pacific-islands-health-care/stories/page-10/") is 656 kB which exceeds the threshold of 550 B, this amount of data can reduce performance.
See more info here: https://nextjs.org/docs/messages/large-page-data
Warning: data for page "/[[...slug]]" (path "/richmond-health-care/stories/page-2/") is 72.9 kB which exceeds the threshold of 550 B, this amount of data can reduce performance.
See more info here: https://nextjs.org/docs/messages/large-page-data
Warning: data for page "/[[...slug]]" (path "/richmond-health-care/stories/") is 65.4 kB which exceeds the threshold of 550 B, this amount of data can reduce performance.
See more info here: https://nextjs.org/docs/messages/large-page-data
Warning: data for page "/[[...slug]]" (path "/northern-california-health-care/stories/page-3/") is 65.6 kB which exceeds the threshold of 550 B, this amount of data can reduce performance.
See more info here: https://nextjs.org/docs/messages/large-page-data
Warning: data for page "/[[...slug]]" (path "/charleston-health-care/stories/page-10/") is 594 kB which exceeds the threshold of 550 B, this amount of data can reduce performance.
See more info here: https://nextjs.org/docs/messages/large-page-data
```

.github/workflows/archive.yml
```diff
@@ -132,7 +132,7 @@ jobs:
         max_attempts: 3
         retry_on: error
         timeout_minutes: 30
-        command: cd main && yarn export --DRUPAL_CLIENT_ID ${{ secrets.PROD_DRUPAL_CLIENT_ID }} --DRUPAL_CLIENT_SECRET ${{ secrets.PROD_DRUPAL_CLIENT_SECRET }} --USE_REDIS=false
+        command: cd main && yarn export --DRUPAL_CLIENT_ID ${{ secrets.PROD_DRUPAL_CLIENT_ID }} --DRUPAL_CLIENT_SECRET ${{ secrets.PROD_DRUPAL_CLIENT_SECRET }}
```
tjheffner, line 135: we'll still need to pass `false` to this command in some way.

package.json has: `"export": "yarn build:env-loader && node ./scripts/yarn/export.js --SSG --USE_REDIS",` so it will default to true when run in the workflows. this was fine before, because we still overrode it here

Locally, I was able to do that by adding: `.option('-nr, --no-USE_REDIS', 'Disable redis')` but you may know a cleaner way to negate boolean values

ryguyk: This is exactly some of the interaction I was worried about with this. I wasn't sure if there were any situations where a flag was overriding another flag.

The solution you mention is the path forward. I didn't add the option because I added defaults (`false`) to all the boolean flags. But I assumed we'd eventually need to be able to explicitly negate them for some reason. I'm going to add to this PR the ability to negate all of them, and then fix this line in `archive.yml` as well.

.github/workflows/content-release.yml
```diff
@@ -150,7 +150,7 @@ jobs:
         max_attempts: 3
         retry_on: error
         timeout_minutes: 30
-        command: cd main && yarn export --DRUPAL_CLIENT_ID ${{ secrets.PROD_DRUPAL_CLIENT_ID }} --DRUPAL_CLIENT_SECRET ${{ secrets.PROD_DRUPAL_CLIENT_SECRET }} --USE_REDIS=false
+        command: cd main && yarn export --DRUPAL_CLIENT_ID ${{ secrets.PROD_DRUPAL_CLIENT_ID }} --DRUPAL_CLIENT_SECRET ${{ secrets.PROD_DRUPAL_CLIENT_SECRET }}
```
tjheffner, line 153: this one would also need `--no-USE_REDIS`

tjheffner (commented):
LGTM once the second worfklow (`content-release.yml`) is also not using redis for the time being

tjheffner (approved):
Thank you!

PR #518, opened 2024-04-23
VACMS-17588: Adds some additional information to existing docs on Typescript in this repo

ryguyk:
## Description
Relates to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/17588

## Testing done
N/A. This is documentation only.

## QA steps
- [ ] Ensure completeness and accuracy of documentation.

### DIFF
--- a/READMEs/typescript.md
+++ b/READMEs/typescript.md
@@ -1,40 +1,162 @@
-# Typescript
+# Typescript in `next-build`
 
-Next-build is primarily written in [Typescript](https://www.typescriptlang.org/), a strongly typed language based on Javascript that gives us better tooling for writing intentional, error-free code. Typescript is compiled to standard Javascript as part of the build process, so what is delivered to the browser is always Javascript that will run on all browsers and devices we support.
+## Introduction
+
+This application (`next-build`) is primarily written in [Typescript](https://www.typescriptlang.org/), a strongly typed language based on Javascript that gives us better tooling for writing intentional, error-free code. Typescript is compiled to standard Javascript as part of the build process, so what is delivered to the browser is always Javascript that will run on all browsers and devices we support.
 
 ## Types
 
-next-build is built on top of [Next-Drupal](https://next-drupal.org/), which is also written in Typescript. [Next-Drupal provides a number of types](https://github.com/chapter-three/next-drupal/blob/main/packages/next-drupal/src/types.ts) for standard Drupal data structures and also for its internal operations.
+Typescript provides [basic types](https://www.typescriptlang.org/docs/handbook/basic-types.html) that form the building blocks for more complex, application-specific types, and it's critical that engineers working on this project have a strong grasp of these fundamentals before proceeding. In this document, we'll focus on the more complex types that are specific to our application. Importantly, we'll examine them from the context of the application architecture that gives rise and meaning to the types in question.
+
+The salient pieces of that application architecture can be summarized via the following steps through which data flows in our application:
+
+1. Data entities are fetched from **Drupal**.
+2. Data entities are transformed into standardized **formatted** data structures.
+3. Formatted data entities are passed to **components** (templates).
 
 ### Drupal data types
 
-Every time we are working with a Drupal entity - what Drupal considers a data object with a defined structure - there should be a type definition for that entity. Drupal itself defines types for these entities, and the types we use in next-build should reflect Drupal's structures.
+As we follow the data flow through the steps outlined above, the first important piece of the flow is fetching data from our CMS, in this case Drupal. In order to reap the benefits of strongly typed programming, we need to be able to define types for all of the possible data structures that Drupal might give us (e.g. News Story, Event, etc.)
 
-Our Drupal Typescript types should extend the basic types provided by Next-Drupal. For example, Next-Drupal provides a type `DrupalNode` which provides a defined structure that is shared by any node data. Our defined types for our Drupal nodes should extend that basic type, adding information about the specific fields the node makes use of. Example:
+Fortunately, we have some additional building blocks at our disposal for defining each of these types. `next-build` is built on top of [Next.js for Drupal](https://next-drupal.org/) (`next-drupal`) - which is also written in Typescript - and `next-drupal`[provides a number of types](https://github.com/chapter-three/next-drupal/blob/main/packages/next-drupal/src/types.ts) for standard Drupal data structures. For example, `next-drupal` defines a type `DrupalNode` which provides a defined structure that is shared by any node data. Our defined types for our Drupal nodes extend this basic type, adding information about the specific fields the node makes use of. The News Story node provides a good example:
 
 ```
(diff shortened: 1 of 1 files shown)


READMEs/typescript.md
```diff
@@ -1,40 +1,162 @@
...
+
+```
+  <Button
+    id="1"
+    label="This is a button"
+    url="https://somesite.org/some/path"
+  />
+```
+
+Notably, this accommodation is only in place for paragraph types. It is not in place for our node types. It's not strictly necessary, but it is technically more accurate. **Going forward, it might make sense to align these approaches.** Either remove this work around paragraphs, or add it in for node type handling.
```
tjheffner, line 155: I think this section could be re-worked or potentially removed. We do actually need `type` for both Paragraphs and Nodes.

The underlying templates may not use `type` (whether layouts or paragraph components), but we filter on `type` for both entity classes in `[[...slug]].tsx` or in `paragraph/index.tsx`... it is a necessary field for this reason. You do call that out earlier with this line:

>  This field is necessary during the earlier steps of the process when we need to know what type of data an object represents. The most notable reason for needing that property is so that we can call the appropriate template (this step) to render the data in question.

but it is confusing to me to then have this portion later. The node types do not have this workaround, and behave just fine spreading in the full data object (type not omitted, but also not used in the template proper):

```js
          {resource.type === RESOURCE_TYPES.STORY && (
            <NewsStory {...(resource as FormattedNewsStory)} />
          )}
```

Paragraphs behave the same way (full object spread in, type not removed and not used in the template proper)

```js
    case PARAGRAPH_RESOURCE_TYPES.ALERT_SINGLE:
      return <AlertSingle {...(paragraph as FormattedAlertSingle)} />
```
or
```js
              {/* Tags */}
              {tags && <AudienceTopics {...tags} />}
```

Am I missing something here?

tjheffner: > Yes, definitely, but not in the components/templates. It's unnecessary at best and wrong at worst. The important point here is something that appears in both my writing and your commentary. From your commentary above:
> 
> The underlying templates may not use type (whether layouts or paragraph components)

my commentary was `may not`, not `does not`. Some of them do use the type, and need it to be present. i.e the slug template or any template that does need to render an arbitrary paragraph (resourcesSupport, for example does this:

```js
                {/* Main Content */}
                {/* Note: eventually this key can just be `pargraph.id`, but only once we implement all paragraph types for this resource*/}
                {mainContent.map((paragraph, index) => (
                  <Paragraph key={paragraph?.id || index} {...paragraph} />
                ))}
```

if the `{...paragraph}` being spread in there does not contain `type`, it all falls apart, because the appropriate paragraph will not be found. Same thing with `resource.type` inside of `[[...slug]].tsx`. 

This is why I feel including `type` is important. When a component is calling a paragraph template directly (instead of using the Paragraph iterator), it may not need the type there. But we including type in paragraph formatters, because it may be helpful to have that information (and we do not know where every paragraph will be used at all times!)

To be clear, I like the approach of `ParagraphComponent`, to make the type optional (and thus not require it when using components directly), and I agree when you say:
> A component should not require something that it doesn't use. 


I am just saying, that the `type` key _does_ get used and is valuable information to include with these data objects when we are passing them around. Whether the underlying component requires it or not is almost irrelevant, because it has the potential to be used by the iterators (whether in `...slug` or `Paragraph`) and should be included for that reason.

`ParagraphComponent` is a separate type from `PublishedParagraph`, they serve different functions. Node layouts could use a similar abstraction perhaps (while still accounting for Lovell/Pagination/etc as needed)... but I don't feel as strongly as you that the current approach is "wrong". It's simplified, to account for some of the resource's data rendering in slug vs the "lower" layout template. i.e. the metatag, breadcrumbs, html comment vs editor-entered content

ryguyk: There's nothing in your response with which I wholly disagree. It highlights an important point - namely, that _some_ components do, in fact, require the `type` property. So, I acknowledge that my discussion of "templates/components" is too general. Not all components can be lumped into the same category. I didn't intend to suggest that, but it might easily be confused.

But the components that I _was_ referencing (individual templates, as opposed to 'routing' components like `[[...slug]]` and `Paragraph`) most certainly comply with the rationale I laid out. They don't use `type` and, as such, I implemented a solution that would rightly define their signatures (i.e. no `type` property required).

You said something important above that I want to touch on:
> the type key does get used and is valuable information to include with these data objects when we are passing them around

I agree with this, and I'd even add that it's not only "valuable" but it's _necessary_. But I'd highlight the qualification, "when we are passing them around". We definitely need `type` when we are passing them around. There are no two ways around that. It's a critical piece of the architecture. So the types that are passed to `[[...slug]]` and `Paragraph` _need_ to have `type`. Again, there are no two ways around that. Importantly, though it follows necessarily, the current implementation accommodates this.

But my relatively simple point is that the components at the lower layer don't use it, and, as such, shouldn't require it. Again, as such, I implemented things accordingly.

But I think in the midst of this (important) conversation, we (or maybe I) may have lost sight of the initial feedback you gave about the documentation. As I mentioned in my previous commentary, I think there are two issues at play here - the implementation and the documentation. We've discussed a lot about the implementation, and I'm happy to discuss that further, but my purpose here is to document the implementation so that it's clear to anybody jumping into this application. So, my question to you - is there something about the documentation that feels inaccurate, confusing, etc.? Or is it the implementation that feels confusing, wrong, etc.?

tjheffner: I appreciate the discussion! I think the following statement (that this chain is originally connected to) is what was confusing to me in the documentation:

> Notably, this accommodation is only in place for paragraph types. It is not in place for our node types. It's not strictly necessary, but it is technically more accurate. **Going forward, it might make sense to align these approaches.** Either remove this work around paragraphs, or add it in for node type handling.

I think this is confusing to me because Paragraphs and Node types are different on several levels. Basically what I was trying to get at, is there are more keys than just `type` that would need to be omitted for a similar abstraction on nodes, but would feel weird to me to do so.... the layout templates aren't really referenced anywhere but from `[[...slug]]`. There are a number of other fields from `PublishedEntity` that are used in the higher level slug template but not necessarily on the "lower" layout template. That simplification of types may feel "wrong" because the fields aren't used on the layout template, but seeing as `[[...slug]]` is a layout wrapper, it feels "right" to me, to keep them all bundled on the node resource object.

ryguyk: Ok! I was curious whether the conversation that followed did anything to change your mind on whether that was still confusing.

We can consider changing that wording. Perhaps we can chat about it.

tjheffner: After chatting with Ryan, my mind has been changed. This line is fine, it's documenting the current state of the system and its caveats. The ~incoming~ new AP team can decide for themselves if they'd like to align some of these types (e.g. perhaps add an additional layer of types beyond `drupal` + `formatted`, for purely components)

tjheffner (approved):

PR #526, opened 2024-04-30
devops doc updates; clean up docker-run.sh

jschmidt-civicactions:
## Description
Relates to #[17706](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/17706).

### DIFF
--- a/docker-run.sh
+++ b/docker-run.sh
@@ -1,7 +1,3 @@
 #!/bin/sh
 
-APP_ENV=$APP_ENV yarn build:preview && yarn start \
-    --SITE_URL $SITE_URL \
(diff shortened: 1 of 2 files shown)


docker-run.sh
```diff
@@ -1,7 +1,3 @@
 #!/bin/sh
 
-APP_ENV=$APP_ENV yarn build:preview && yarn start \
```
tjheffner, line 3: these inline vars are no longer required because they exist inside of the `secrets.yaml` and `configmap.yaml` files?

jschmidt-civicactions: That's right, this is vestigial code from when we were thinking of using env vars from the command line, and not setting up the .env files as we do now. Leaving it is confusing.

tjheffner (approved):

PR #560, opened 2024-06-18
VANB-18071: Adding response header to change file paths in urls to point at the correct s3 bucket.

JakeBapple:
## Description
Relates to #[18071](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/18071). (or closes?)

## Testing done
Local testing

## QA steps
 - URL paths on assets should point to an s3 bucket
 - Site behaves normally and assets appear.

```[tasklist]
```


timcosgrove (changes requested):
I think we should incorporate the suggestion in the review, so that we keep the default headers intact.

timcosgrove (approved):

PR #604, opened 2024-07-15
QA Fixes :  Event Layout 

ttran-Therisa:
## Description
Relates to #[18494](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/18494). (or closes?)

## Testing done

- [x] yarn test -- -u
- [x] yarn test:format
- [x] yarn test:coverage

## Screenshots
Prod
Screenshot 2024-07-15 at 3 41 03 PM
Local
Screenshot 2024-07-15 at 3 41 03 PM

## QA steps
What variations of circumstances  need to be checked?
So Production shows the Register option, while Local says the event has passed. From the information from the article including the dropdown "View other times for this events," local is accurate.
However, when the user clicks the "This is an online event" link, you will see that the event is still on-going, but the newer times are not updated in the article. This would indicate Production is right. 
I need a second opinion on which version should we follow? Should our local version be correct since the information from the the article indicates there are no other events passed May or should we figure out why aren't the other events from eventbrite updating to this article?

```[tasklist]
```


mreed101 (commented):
Approved by me pending feedback button change

mreed101 (approved):

Issue #669, opened 2024-08-13
Next Build Get Directions link


mreed101:
## Description

On the event page the get directions link should be using the `<va-link>` component.

## Acceptance Criteria 

- [ ] Replace a tag with va-link component

PR #673, opened 2024-08-13
Issue 669 va-link for google directions

nfpappas-oddball:
# Description
fix issue 669 and add va-link for google directions

## Ticket
https://github.com/department-of-veterans-affairs/next-build/issues/669

## Testing Steps
rewrite index.test.tsx to accommodate the new component
run yarn test

## Screenshots
Before:
image

After:
image


src/templates/common/googleMapsDirections/index.test.tsx
```diff
@@ -11,10 +11,13 @@ describe('GoogleMapsDirections Component', () => {
 
     render(<GoogleMapsDirections title={title} address={address} />)
 
-    const anchorElement = screen.getByRole('link', {
-      name: /get directions on google maps/i,
-    })
+    // const anchorElement = screen.getByRole('va-link', {
```
timcosgrove, line 14: If tests can't grab and use the web components, that's fine, but we should remove this commented code if we don't intend to use it.

timcosgrove (commented):
Do we need this commented out code in the test?

src/templates/common/googleMapsDirections/index.test.tsx
nfpappas-oddball: I left the old code because I wanted to get feedback on the way it worked previously (commented out section) and the new code. I feel like there may be a better way to do the selector but couldn't figure out how so i gave it a test ID. Test still works as it previously did. Don't know if any of this affects 508 either.

ttran-Therisa (approved):
Looks good to me

PR #768, opened 2024-10-02
fix: remove sitemap deploy

nfpappas-oddball:


timcosgrove:
This will not be the solution.

PR #836, opened 2024-12-11
fix: Do not use assets URL for fonts

ph-One:
- Do not use `${ASSETS_URL}` for anything other than `*.js` and `*.css`
- JS and CSS point to the assets bucket https://github.com/department-of-veterans-affairs/vsp-platform-revproxy/blob/main/template-rendering/revproxy-vagov/templates/nginx_website_server.conf.j2#L332
- All other resources are pointed at the [next]content bucket https://github.com/department-of-veterans-affairs/vsp-platform-revproxy/blob/main/template-rendering/revproxy-vagov/templates/nginx_website_server.conf.j2#L402

### DIFF
--- a/src/pages/_document.tsx
+++ b/src/pages/_document.tsx
@@ -36,28 +36,28 @@ const Document = () => {
         {/* Preload main fonts */}
         <link
           rel="preload"
-          href={`${ASSETS_URL}sourcesanspro-bold-webfont.woff2`}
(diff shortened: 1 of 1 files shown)


ph-One:
The root cause was missing files in the assets bucket. Nothing to change. Closing this PR.

timcosgrove (changes requested):
`/generated` still pulls from vets-website ultimately. This wouldn't solve anything.

PR #840, opened 2024-12-12
Check banner object for field before trying to iterate over it.

timcosgrove:
# Description
Banner data did not contain a field we were expecting and the banner code attempted to iterate over it.

This threw an error which caused the entire page to not be built, which is a problem for us.

## QA steps
Visit /boston-health-care/events on this Tugboat instance; it should show a page.


eselkin (approved):
LGTM
At some future point it might be good to restructure this be a little simpler
```
(banner.field_banner_alert_vamcs || []).map...
```
But it works

7hunderbird:
Looks good to me +1 

Went to https://pr840-xxegfpomzm0ryj6qekdo8eodojtrjvbo.tugboat.vfs.va.gov/boston-health-care/events/ and it's loading with the banner.

CleanShot 2024-12-12 at 15 50 19

PR #881, opened 2025-02-13
Scaffold runbook docs

cvalarida:
# Description
This is the first pass of the template migration runbook. For now, it contains mostly stuff from [this Canvas](https://dsva.slack.com/canvas/CT4GZBM8F), but I made a place for us to add more information about how to actually _implement_ that migration.

For now, I don't _really_ know what that looks like, so it's mostly a placeholder.

## Ticket
Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/20317


laflannery:
@cvalarida (I don't actually know how to do fancy comments) I just have a few things:

- In the post-launch file can we update the **Responsible** people to be `CMS Team & the team doing the migration`
  - I can see this being complicated to implement in real life and I think we are going to have to see how this works the first few templates but currently we have both teams doing the work so technically they both will be responsible (Erika requested that the CMS team do the content-build clean up stuff and Michelle recommended the migrating team do the Prod verification)
- I know this is just a first draft but in the Active migration doc, I think I would add a step in the **How do we do it section** right before "Create a new template in `next-build` to match the `content-build` template" that says something like "Confirm there is no existing template for the content-build template you are going to migrate" (because there are already some in there that were started).
- This is super picky but in the **How do we do it** sections for both the Active Migration doc and the Post-launch doc, the steps are all number 1, can you update them to be 1, 2, etc.?

cvalarida:
> @cvalarida (I don't actually know how to do fancy comments) I just have a few things:
> 
> * In the post-launch file can we update the **Responsible** people to be `CMS Team & the team doing the migration`
>   
>   * I can see this being complicated to implement in real life and I think we are going to have to see how this works the first few templates but currently we have both teams doing the work so technically they both will be responsible (Erika requested that the CMS team do the content-build clean up stuff and Michelle recommended the migrating team do the Prod verification)
> * I know this is just a first draft but in the Active migration doc, I think I would add a step in the **How do we do it section** right before "Create a new template in `next-build` to match the `content-build` template" that says something like "Confirm there is no existing template for the content-build template you are going to migrate" (because there are already some in there that were started).
> * This is super picky but in the **How do we do it** sections for both the Active Migration doc and the Post-launch doc, the steps are all number 1, can you update them to be 1, 2, etc.?

I think the first two edits make sense. 👍 

For the numbering, I'd prefer to keep it 1s because we'll be making changes to them and renumbering is a pain. The rendered markdown will order them correctly. 🤓 
image
image

cvalarida:
Just pushed up those changes, @laflannery.

laflannery (approved):

brianseek (approved):
Makes sense to me. Looks like a good start!

PR #899, opened 2025-02-21
Don't retry 404s

cvalarida:
# Description
When `proxy-fetch` gets a non-ok response, it retries up to 5 times (total of 6 attempts). This doesn't make sense when the response is a 404 error.


packages/proxy-fetcher/src/index.ts
```diff
@@ -87,7 +87,7 @@ export const getFetcher = (
       //  In non-debug mode: only on final attempt
       const logFailedRequest = debug ? true : attempt === retryCount + 1
 
-      if (!response.ok) {
+      if (!response.ok && response.status !== 404) {
```
brianseek, line 90: Unless I'm reading this wrong, this still retries 404, it just doesn't log them. Is that the intent?

brianseek: If we wanted to abort for 404, on the next line change it to be ` if (logFailedRequest || response.status === 404)`

cvalarida: 🤦 You're absolutely right. Just updated.

packages/proxy-fetcher/src/index.ts
```diff
@@ -101,8 +101,17 @@ export const getFetcher = (
...
-        throw new Error('Failed request')
+        throw new Error(
+          `Failed request to ${response.url}: ${response.status} ${response.statusText}`
+        )
       }
+
+      // Don't retry if we shouldn't
+      if ([404, 403].includes(response.status)) {
+        const { AbortError } = await import('p-retry')
+        throw new AbortError(response.statusText)
```
cvalarida, line 112: Looks like this is the right way to prevent retries: https://www.npmjs.com/package/p-retry

brianseek: TIL `await import()`

brianseek (approved):
LGTM!

PR #904, opened 2025-02-25
API Explorer enhancements

cvalarida:
# Description
The API explorer was a little bit of a pain to use, so I made it interactive. 😇 

image

## Features
- Explore the API dynamically! Doesn't require a code change to look at the response
- Once you fetch a content type, it'll grab the `relationshipNames` and allow you to quick-fill those by pressing the buttons
- If the request fails due to a circular reference or something, it'll show you
   image
- JSON output has syntax highlighting, collapsible objects, object counts, and "copy to clipboard" features

## Note
This _does_ add another API route. It's note currently gated on `process.env.BUILD_TYPE` (or whatever), but we can 100% make that a thing. It also removes the "404 this page if it's not SSR" bit, but I don't know that that's a problem.

If we go to prod with this as a persistent server, we'll need to gate this based on a build type or (ideally) a feature toggle. (Ideally one that we can toggle on for us, but off for everybody else.)


src/pages/_playground/api-explorer.tsx
```diff
@@ -1,38 +1,367 @@
...
  * in our mocks. Ideally I would like it to become an interactive query explorer.
  */
 
-export default function Page(props) {
-  const output = JSON.stringify(props.data, null, 2)
-  return <pre>{output}</pre>
+/**
+ * Simple JSON syntax highlighter component
+ */
+const SyntaxHighlightedJson: React.FC<{ data: unknown }> = ({ data }) => {
```
cvalarida, line 31: This is all AI-generated and I have no opinion. We can pull in an external library for this if we care. It seemed to work well enough as-is, though, so I didn't even bother trying.

src/pages/_playground/api-explorer.tsx
cvalarida, line 1: Basically everything in here is AI-generated. I didn't spend much time doing more than make sure it works as expected because this is just a quick dev tool to help us migrate content types. 🤷 Didn't bother me overmuch if it's kind of ugly.

Still, if you have opinions and want to see changes, let me know and I'd be happy to actually spend a little time on it.

timcosgrove (commented):
I love it. A couple minor usability requests, if they're not too much trouble:
- It would be nice if clicking an active relationshipName button deactivated it and removed it from the query. I do see that if you manually remove the relationship name, the active state also gets removed from the button
- It would be cool if the debug information always showed; or, better, if the constructed `DrupalJsonApiParams` object with method calls were printed out, so it could be lifted directly into a query.
- I don't know if it's possible to add `relationshipNames` as we drill down. Like, if we add `field_media`, we get a set of relationshipNames within that item. It would be nice if we could then add `field_media.image`, `field_media.thumbnail`, etc, as options (that is the correct include syntax)

image


You are selling me on AI as an assistive tool, though. This is great.

cvalarida:
> It would be cool if the debug information always showed; or, better, if the constructed `DrupalJsonApiParams` object with method calls were printed out, so it could be lifted directly into a query.

So have the API return an object like:

```json
{
  "DrupalJsonApiParams": { ... },
  "queryResult": { ... }
}
```

Then we can use the `queryResult` property for what we're already displaying, and `DrupalJsonApiParams` for both debugging and easy copy pasta when creating queries. That sound about right?

> I don't know if it's possible to add `relationshipNames` as we drill down. Like, if we add `field_media`, we get a set of `relationshipNames` within that item. It would be nice if we could then add `field_media.image`, `field_media.thumbnail`, etc, as options (that is the correct include syntax)

From what I can tell, when we add string literals into the `.addInclude([])` array, it may add those includes to all entity references as well. Not 100% on that, though; I just know that when I add `field_local_health_care_service_` to the includes, it goes belly up because of a circular reference.

timcosgrove:
> Then we can use the `queryResult` property for what we're already displaying, and `DrupalJsonApiParams` for both debugging and easy copy pasta when creating queries. That sound about right?

More or less. Queries are currently constructed more or less like so: https://github.com/department-of-veterans-affairs/next-build/blob/main/src/data/queries/event.ts#L19C1-L26C2

It would be nice to lift what the api-explorer is using to run its query into a usable form like that. 

That said, probably much easier said than done, especially because of the `...getNestedInclude` things and also the use of `RESOURCE_TYPE` constants in place of literals. I wouldn't worry about it at this stage.

I think y'all will find the friction points as you do the work and figure out what will be useful to you and can refine this. Probably not needed right now.

timcosgrove:
> From what I can tell, when we add string literals into the `.addInclude([])` array, it may add those includes to all entity references as well. Not 100% on that, though; I just know that when I add `field_local_health_care_service_` to the includes, it goes belly up because of a circular reference.

I think the `field_local_health_care_service_` issue may be separate from what I'm describing. The query syntax for Drupal JSON API allows for sub-includes like `field_media.thumbnail`: https://www.drupal.org/docs/core-modules-and-themes/core-modules/jsonapi-module/includes

Actually, that whole section is probably interesting to you, so that you know and can verify what REST requests are being generated underlyingly: https://www.drupal.org/docs/core-modules-and-themes/core-modules/jsonapi-module

A note when comparing an actual Drupal JSON API response to what the api-explorer shows you: next-drupal-query does some flattening and normalizing to the data along the way. If you compare a direct Drupal API response to what you get from next-drupal-query, you'll see subtle differences. 
Example direct response: https://prod.cms.va.gov/jsonapi/node/health_care_local_facility/fb83cadd-3383-4a22-9859-8644cd7ca808

Mostly just telling you this so that if you ever do compare the Drupal JSON API version to the massaged next-drupal-query version you know why.

timcosgrove (approved):
This is great.

PR #908, opened 2025-02-27
Scaffold vamc facility

cvalarida:
# Description
This is just the very starting scaffolding of the VAMC Facilities. I ran into some trouble with getting the types all worked out, and there was some existing crust to work through, but I finally got me a blank page on which we can start building stuff.

Almost everything in this is auto-generated and untouched, but I've noted in my review comments where I actually made some changes.

## Ticket
Relates to #[#20436](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/20436)

## Testing Steps

1. `FEATURE_NEXT_BUILD_CONTENT_ALL=true yarn dev`
2. Navigate to http://localhost:3000/memphis-health-care/locations/lt-col-luke-weathers-jr-va-medical-center/ (or [port 3999](http://localhost:3999/memphis-health-care/locations/lt-col-luke-weathers-jr-va-medical-center/), if that's what you've got going)
3. See the glorious blank page! The important part is that it doesn't 404

## QA steps

## Screenshots
Before: 🤷 
After:
image

# Reviewer

### Reviewing a PR

## Standard Checks

```
```

## Merging an Approved Layout

> [!NOTE]
> I added this new page type to `PAGE_RESOURCE_TYPES` because I needed to for it to render the page, but as far as I can tell, it won't actually start building the page in production unless the `FEATURE_NEXT_BUILD_CONTENT_${typeName}` environment variable is true—which it shouldn't be in prod. @timcosgrove please make sure I'm not mistaken there! 😅


src/types/drupal/node.ts
cvalarida, line 1: I'm gonna have to go back and change this with the follow-up PRs, but the short story is that this type already existed, but it wasn't accurate. For now, the mock data I've got doesn't match this version because I don't have hydrated entity references in there. That'll come in the follow-ups as well.

src/types/formatted/healthCareLocalFacility.ts
```diff
@@ -0,0 +1,5 @@
+import { PublishedEntity } from './publishedEntity'
+
+export type HealthCareLocalFacility = PublishedEntity & {
+  title: string
+}
```
cvalarida, line 5: This isn't true. 🙈 Changes will be coming in the next PRs.

src/lib/constants/resourceTypes.ts
```diff
@@ -10,7 +10,7 @@ export const RESOURCE_TYPES = {
   STORY: 'node--news_story',
   STORY_LISTING: 'node--story_listing',
   SUPPORT_SERVICES: 'node--support_service',
-  VAMC_FACILITY: 'node--health_care_local_facility',
+  HEALTH_CARE_LOCAL_FACILITY: 'node--health_care_local_facility',
```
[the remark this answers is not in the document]
timcosgrove: I think this is fine, though note line 10. I think when we set these up, our intention was that the left would be based on the human/shared name of the type of content. The right is explicitly tied to Drupal's machine name for it. 

See for example this page: https://prod.cms.va.gov/admin/structure/types/manage/health_care_local_facility

from the URL you can see Drupal's machine name for the thing; at the page, you can see the label/human name for the thing, also visible here: 
image

src/mocks/healthCareLocalFacility.mock.json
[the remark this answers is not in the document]
timcosgrove: I'd say do what makes sense. Provided the mock is self-contained - it doesn't rely on API or DB calls to live data - it doesn't necessarily matter. But, it might be nice to reference a working facility.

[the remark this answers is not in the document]
timcosgrove: You could create a fictional facility on Ringworld.... 🤔

src/data/queries/healthCareLocalFacility.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  return entity
+}
+
+export const formatter: QueryFormatter<
+  NodeHealthCareLocalFacility,
+  HealthCareLocalFacility
+> = (entity: NodeHealthCareLocalFacility) => {
+  return {
+    ...entityBaseFields(entity),
```
timcosgrove, line 49: I assume this will be fleshed out eventually.

cvalarida: Yup! When I actually do the thing. 👍 This PR is just for scaffolding up a blank page. Next up, we'll build out features, which will require changes to this.

timcosgrove (approved):

PR #910, opened 2025-02-27
Fixing Event Page When section formatting

moanu404:
# Description
This PR creates a new date formatter function that addresses a defect currently leading to showing of the date in the "When" section inconsistently and it formats the date/time and time zone to be consistent with the AC. I tried to keep the formatting as consistent as possible with the format on the event listing page.
Now also using date-fns, date-fns-tz library
## Ticket
Relates to #[19696](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/19696)

## Developer Task

```
- [ ] Alerted the #accelerated-publishing Slack channel to request a PR review.
```

## Testing Steps
- [ ] Run all unit tests `yarn test`
- [ ] Load up https://www.va.gov/outreach-and-events/events 
- [ ] Click into the event and observe the "When:" area and see the incorrect formatting 
- [ ] EG https://www.va.gov/dayton-health-care/events/76691/ 
- [ ] Load up your local server and pull down this branch
- [ ] check http://localhost:3999/outreach-and-events/events/ (takes forever) OR http://localhost:3999/dayton-health-care/events/
- [ ] do the same parity check and ensure the formatting is consistent and so is the time zone is in your local timezone

## QA steps

## Screenshots
Before: (PROD)
image

After: (Local)
image

# Reviewer

### Reviewing a PR

## Standard Checks

```
```

## Merging an Approved Layout

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "clsx": "^2.1.0",
     "commander": "^12.0.0",
     "date-fns": "^4.1.0",
+    "date-fns-tz": "^3.2.0",
     "debug": "^4.3.4",
     "drupal-jsonapi-params": "^2.3.1",
     "env-loader": "0.1.0",
--- a/src/lib/utils/date.test.ts
+++ b/src/lib/utils/date.test.ts
@@ -1,3 +1,10 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import { jest } from '@jest/globals'
+
+jest.mock('date-fns-tz', () => ({
+  formatInTimeZone: jest.fn(),
+}))
+
 import {
   isISOString,
   parseMilliseconds,
@@ -346,45 +353,87 @@ test('parseDate should return null for inputs that do not match any date format'
 
 describe('formatEventDateTime', () => {
   beforeEach(() => {
-    // Check if we're in UTC (CI environment)
-    const isUTC = new Date().getTimezoneOffset() === 0
-    const mockTimezone = isUTC ? 'EST' : 'PST'
-
-    jest.spyOn(Intl, 'DateTimeFormat').mockImplementation(() => ({
-      formatToParts: () => [
-        {
-          type: 'timeZoneName',
-          value: mockTimezone,
-        },
-      ],
-      format: () => '',
-      resolvedOptions: () => ({
-        locale: 'en-US',
-        calendar: 'gregory',
-        numberingSystem: 'latn',
-        timeZone: isUTC ? 'America/New_York' : 'America/Los_Angeles',
-      }),
-      formatRange: () => '',
-      formatRangeToParts: () => [],
-    }))
+    jest.restoreAllMocks()
   })
 
   afterEach(() => {
     jest.restoreAllMocks()
   })
 
-  it('formats date and time with timezone', () => {
-    const mostRecentDate = {
-      value: 1683216000,
-      endValue: 1683223200,
-    }
+  /**
+   * Helper function to mock the system time zone
+   */
+  const mockUserTimeZone = (mockTimeZone: string) => {
+    jest
+      .spyOn(Intl.DateTimeFormat.prototype, 'resolvedOptions')
+      .mockReturnValue({
+        locale: 'en-US',
+        calendar: 'gregory',
+        numberingSystem: 'latn',
+        timeZone: mockTimeZone,
+      })
+  }
 
+  it('should format time based on mocked Eastern Time (ET)', () => {
+    mockUserTimeZone('America/New_York')
+    const mostRecentDate = {
+      value: 1683201600,
+      endValue: 1683208800,
+    }
     const result = formatEventDateTime(mostRecentDate)
-    const [datePart, timePart] = result.split(' – ')
-    expect(datePart).toMatch(/Thu\. May 4, 2023, \d{1,2}:\d{2} [ap]\.m\./)
-    expect(timePart).toMatch(/\d{1,2}:\d{2} [ap]\.m\. [A-Z]T/)
+    expect(result).toMatch(/Thu\. May 4, 2023, 8:00 a\.m\. – 10:00 a\.m\. ET/)
   })
-  it('returns empty string for null input', () => {
-    expect(formatEventDateTime(null)).toBe('')
+
+  it('should format time based on mocked Pacific Time (PT)', () => {
+    mockUserTimeZone('America/Los_Angeles')
+    const mostRecentDate = {
+      value: 1683201600,
+      endValue: 1683208800,
+    }
+    const result = formatEventDateTime(mostRecentDate)
+    expect(result).toMatch(/Thu\. May 4, 2023, 5:00 a\.m\. – 7:00 a\.m\. PT/)
+  })
+
+  it('should format time to User time even when mostRecentDate has undefined timezone', () => {
+    mockUserTimeZone('America/Chicago')
+    const mostRecentDate = {
+      value: 1683201600,
+      endValue: 1683208800,
+      timezone: undefined,
+    }
+    const result = formatEventDateTime(mostRecentDate)
+    expect(result).toMatch(/Thu\. May 4, 2023, 7:00 a\.m\. – 9:00 a\.m\. CT/)
+  })
+
+  it('should return "Invalid date" for incorrect timestamps', () => {
+    mockUserTimeZone('America/New_York')
+    const mostRecentDate = {
+      value: -1,
+      endValue: -1,
+    }
+    const result = formatEventDateTime(mostRecentDate)
+    expect(result).toBe('Invalid date – Invalid date ET')
+  })
+
+  it('should format event in US timezone if user is international (Asia)', () => {
+    mockUserTimeZone('Asia/Tokyo')
+    const mostRecentDate = {
+      value: 1683201600,
+      endValue: 1683208800,
+      timezone: 'America/New_York',
+    }
+    const result = formatEventDateTime(mostRecentDate)
+    expect(result).toMatch(/Thu\. May 4, 2023, 8:00 a\.m\. – 10:00 a\.m\. ET/)
+  })
+
+  it('should correctly format an event from 12:00 a.m. to 12:00 p.m.', () => {
+    mockUserTimeZone('America/New_York')
+    const mostRecentDate = {
+      value: 1683172800,
+      endValue: 1683216000,
+      timezone: 'America/New_York',
+    }
+    const result = formatEventDateTime(mostRecentDate)
+    expect(result).toBe('Thu. May 4, 2023, 12:00 a.m. – 12:00 p.m. ET')
   })
 })
--- a/src/lib/utils/date.ts
+++ b/src/lib/utils/date.ts
@@ -257,10 +257,11 @@ export const deriveFormattedTimestamp = (datetime) => {
   return reformatTime(formattedTime)
 }
 
-import { format } from 'date-fns'
-
+import { formatInTimeZone } from 'date-fns-tz'
 /**
- * Bonus points for describing this object, but the name is pretty descriptive by itself. 🤷‍♂️
+ * Represents a time range for events or appointments using Unix timestamps.
+ * Includes optional timezone support for display.
+ * All timestamps are in seconds since Unix epoch.
  */
 interface DateTimeRange {
   /**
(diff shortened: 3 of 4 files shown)


src/lib/utils/date.test.ts
```diff
@@ -342,3 +343,48 @@ test('getDateParts should return null for falsy values', () => {
 test('parseDate should return null for inputs that do not match any date format', () => {
   expect(parseDate('')).toBe(null)
 })
+
+describe('formatEventDateTime', () => {
+  beforeEach(() => {
+    // Check if we're in UTC (CI environment)
```
moanu404, line 349: this makes me sad but the tests kept failing in CI cuz UTC

src/lib/utils/date.ts
```diff
@@ -257,6 +257,36 @@ export const deriveFormattedTimestamp = (datetime) => {
...
 
+import { format } from 'date-fns'
+
+interface DateTimeRange {
+  value: number
+  endValue: number
+  timezone?: string
+}
+
+export const formatEventDateTime = (mostRecentDate: DateTimeRange | null) => {
```
moanu404, line 268: Tried to make this more or less adherent to vets-website except not using moment

moanu404:
@laflannery while tugboat is being flakey on me I had a quick question about formatting
https://github.com/department-of-veterans-affairs/next-build/pull/910 here we can see that 12 pm should be noon and 12 am midnight etc, should we account for this here?

laflannery:
@moanu404 The formatting looks great! A couple comments:

1. Regarding your noon/midnight question, we weren't doing this on Prod prior to the migration so it's ok if we use 12:00 am/pm here as well. I actually think it makes a bit more sense to use the numeric times here.
2. On your stories PR, you changed [this common file](https://github.com/department-of-veterans-affairs/next-build/blob/main/src/templates/common/socialLinks/index.tsx) and it looks like it updated the spacing on the social links in the Events detail pages. I didn't catch it because I didn't realize it was using a shared component. Can you take a look? This is a screenshot from a previous ticket to show what it should look like (ignore the red lines, those were for the purpose of the original ticket I took this from):
image

laflannery:
@moanu404 I re-reviewed the recurring event times and those look good now, thanks!

laflannery:
After discussion with @moanu404 we decided to go ahead and merge this and handle the spacing issue I noted above in [a separate ticket](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/20874)

package.json
```diff
@@ -56,6 +56,7 @@
     "chrome-launcher": "^1.1.2",
     "clsx": "^2.1.0",
     "commander": "^12.0.0",
+    "date-fns": "^4.1.0",
```
cvalarida, line 59: https://blog.date-fns.org/v40-with-time-zone-support/ 🎉 

This is weirdly exciting. 😂 🤓

src/lib/utils/date.test.ts
cvalarida: 🤔 I don't think I understand why this is necessary. If we're in UTC (CI), we say the mock time zone is EST, but if it isn't (including for actual EST), we say the mock time zone is PST. Why not stick with one or the other regardless of what the system time zone is?

How was this failing?

src/lib/utils/date.test.ts
```diff
@@ -342,3 +343,48 @@ test('getDateParts should return null for falsy values', () => {
...
+  it('formats date and time with timezone', () => {
+    const mostRecentDate = {
+      value: 1683216000,
+      endValue: 1683223200,
+    }
+
+    const result = formatEventDateTime(mostRecentDate)
+    const [datePart, timePart] = result.split(' – ')
+    expect(datePart).toMatch(/Thu\. May 4, 2023, \d{1,2}:\d{2} [ap]\.m\./)
+    expect(timePart).toMatch(/\d{1,2}:\d{2} [ap]\.m\. [A-Z]T/)
```
cvalarida, line 385: I would expect this to expect a specific value, since we're providing a specific value. Like, either it's `9:00 a.m. - 11:00 a.m. PT` or `noon - 3:00 p.m. ET`.

Can we add tests to check that the time zone is being translated correctly as well as the a.m. / p.m. / noon / midnight? This test checks that the format _looks_ correct, not that the times actually _are_ correct.

src/lib/utils/date.ts
```diff
@@ -257,6 +257,36 @@ export const deriveFormattedTimestamp = (datetime) => {
   return reformatTime(formattedTime)
 }
 
+import { format } from 'date-fns'
+
+interface DateTimeRange {
+  value: number
+  endValue: number
+  timezone?: string
+}
```
cvalarida, line 266: Is this the same shape as in `vets-website`? If not, I recommend renaming `value` and `endValue` to `startTime` and `endTime`. In either case, I recommend putting a JSDoc comment on the properties to describe them.
```suggestion
/**
 * Bonus points for describing this object, but the name is pretty descriptive by itself. 🤷‍♂️
 */
interface DateTimeRange {
  /**
   * Unix timestamp for the beginning of the date time range.
   */
  value: number
  /**
   * Unix timestamp for the end of the date time range.
   */
  endValue: number
  /**
   * ISO timezone string (e.g. 'America/Los_Angeles')
   */
  timezone?: string
}
```

It feels pretty verbose, but we get to leverage the power of our language server hover docs that way. 🤓

src/lib/utils/date.ts
cvalarida: Can you add a JSDoc comment to this as well? I'd like to see at a glance what the format is and call out that this is _specifically_ for formatting dates and times for _events_. Bonus points for linking to the [design system content style guide for dates and time](https://design.va.gov/content-style-guide/dates-and-numbers#times-and-time-zones) for reference.

**Pro tip:** Get the AI to write it for you, then correct any mistakes it may make. 🤓 

We might could extract this formatting logic later since it seems more broadly applicable than for _just_ events.

src/lib/utils/date.ts
```diff
@@ -257,6 +257,36 @@ export const deriveFormattedTimestamp = (datetime) => {
...
+  if (!mostRecentDate) return ''
+
+  const startsAtUnix = mostRecentDate.value
+  const endsAtUnix = mostRecentDate.endValue
+  const formattedStartsAt = startsAtUnix
+    ? format(new Date(startsAtUnix * 1000), 'EEE. MMM d, yyyy, h:mm aaaa')
+    : ''
+  const formattedEndsAt = endsAtUnix
+    ? format(new Date(endsAtUnix * 1000), 'h:mm aaaa')
+    : ''
```
cvalarida, line 278: Alright, so this gets a little funky.

1. If I pass in `formatEventDateTime({ value: 0, endValue: 0 })`, it looks like the output is going to be ` –  PT`; is that right?
2. Neither of these account for the `noon` or `midnight` edge cases as described by the style guide

src/lib/utils/date.ts
```diff
@@ -257,6 +257,36 @@ export const deriveFormattedTimestamp = (datetime) => {
...
+  const endsAtTimezone =
+    new Intl.DateTimeFormat('en-US', {
+      timeZoneName: 'short',
+    })
+      .formatToParts(new Date())
+      .find((part) => part.type === 'timeZoneName')
+      ?.value?.replace(/S|D/i, '') || ''
```
cvalarida, line 285: > When referencing international time zones, use the time zone name for that country or the UTC (coordinated universal time) offset. (Example: Central European Time; Korea Standard Time; UTC +2)

A couple things here:
1. `timeZoneName: 'short'` is apparently _only_ for US timezones (maybe only continental US; I don't know)
2. If we just blindly remove the first `S` or `D` like this `.replace()` does _and_ we use a time zone outside of the US, we may end up with a formatted timezone that looks like `Korea tandard Time` 😅

src/lib/utils/date.ts
```diff
@@ -257,6 +257,36 @@ export const deriveFormattedTimestamp = (datetime) => {
...
+
+  return `${formattedStartsAt} – ${formattedEndsAt} ${endsAtTimezone}`
```
cvalarida, line 287: > - In body copy, write out ranges using sentence construction: We’re open 9:00 a.m. to 6:00 p.m. ET, Monday through Friday.
> - Where space is limited, indicate ranges with the - hyphen without spaces on either side. Don’t combine the hyphen with from/between sentence construction.

Not sure if this function will be used in these cases, but food for thought: What would this look like if we added an `options?` parameter that we could use to configure the output based on where we're using it?

I think until we run into a situation where we for sure want to use `formatEventDateTime`, we shouldn't worry about it, but food for thought. :homer-disappear:

cvalarida (commented):
Looks like a solid start! I think there are some edge cases we need to consider, and the testing needs to be a little more robust.

src/lib/utils/date.ts
moanu404: 1. Good catch, I can definitely add some error handling around that, however would that be a valid input given from what I am looking up a timestamp of 0 means January 1, 1970 00:00:00 UTC and given in the VA events system we only look for events in the future (everything else shows up as an invalid date atm in prod) the scenario seems unlikely. Does not hurt to add it though either ways. 

2. So I confirmed with Laura, and she commented here too -  that current behavior in Prod, does not have noon/midnight so it isn't necessary to implement here

src/lib/utils/date.ts
moanu404: Oh yes the code I based this off seemed to be consistent from the code i was referring to on `vets-website` like which appeared to be US specific  -  
https://github.com/department-of-veterans-affairs/vets-website/blob/5c0d4b46de7f87095b88d44146092c09ec583acf/src/applications/static-pages/events/components/Results/index.js#L84

This is getting more funky (when have dates not been lol) But if we are using end users timezones now, and someone is not logging in from within US from the style guide i suppose we can display a UTC offset?

src/lib/utils/date.ts
moanu404: thats a good thought for sure if we see another use case for it, until then, it lives in the utils class if we ever need it again

src/lib/utils/date.test.ts
moanu404: I was trying this a bit ago and I was checking against a hardcoded string and it kept failing against it, I can spend some more time to brainstorm alternate ways to better ways to test this but I was hitting a wall against the UTC timezone results going against my expects and failing the CI

src/lib/utils/date.ts
cvalarida: 1. :party-wizard:
2. :design-system-twitch-intensifies:

package.json
moanu404: I used `date-fns-tz` for refactoring the edge cases since I found it easier to do our funky formatting and it had more wider usage 😅

src/lib/utils/date.test.ts
```diff
@@ -342,3 +350,90 @@ test('getDateParts should return null for falsy values', () => {
...
+    }
+    const result = formatEventDateTime(mostRecentDate)
+    expect(result).toBe('Invalid date – Invalid date ET')
+  })
+
+  it('should format event in US timezone if user is international (Asia)', () => {
+    mockUserTimeZone('Asia/Tokyo')
+    const mostRecentDate = {
+      value: 1683201600,
+      endValue: 1683208800,
```
moanu404, line 422: AI helped a lot to get these stamps 🤖

moanu404:
New change is such that `mostRecentDate` timezone will now be used if the system user is located internationally, and if they are in the US, then the event times will be shown in their timezone

laflannery:
Looks good to me, I tested a couple international timezones and saw CT and PT for the events as expected

cvalarida (approved):
I apparently never actually submitted this. 🫠 

I've got a few questions / comments, but nothing merge-blocking, I think.

package.json
moanu404: Oops posted this in the wrong spot since I have so many PR's open lol
So I tried that initially and it did not work (I cannot pull up that terminal log alas) but it seemed to require me to have `date-fns` in addition for it to pass, which was weird. I'll try to dig it up some more and see what exact error it was showing me. The vets-website counterpart is now using `date-fns-tz`  and I saw more examples with formatting with it is what I meant.

src/lib/utils/date.test.ts
```diff
@@ -1,3 +1,10 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import { jest } from '@jest/globals'
+
+jest.mock('date-fns-tz', () => ({
+  formatInTimeZone: jest.fn(),
+}))
```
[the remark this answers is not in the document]
moanu404: Closed my terminal window (argh) but the pre-commit linter was unhappy about the presence of jest in a test file, so I had to add the `/* eslint-disable import/no-extraneous-dependencies */` as far as I can recall

src/lib/utils/date.ts
```diff
@@ -257,6 +257,96 @@ export const deriveFormattedTimestamp = (datetime) => {
...
+
+  // Format start and end times in target timezone
+  const formattedStartsAt =
+    startsAtUnix && startsAtUnix > 0
+      ? formatInTimeZone(
+          new Date(startsAtUnix * 1000),
+          targetTimeZone,
+          'EEE. MMM d, yyyy, h:mm aaaa'
+        )
+      : 'Invalid date'
```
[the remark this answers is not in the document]
laflannery: FYI - Past Events cannot be reached unless:
- You know the URL exactly and can go directly there or
- You search and the event shows up in the search results
However that is only temporary, eventually past events won't be on the site or reachable at all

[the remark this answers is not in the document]
laflannery: Ya - if it's for overall monitoring I think that makes sense but if it's just for the past events Invalid date I don't think it makes sense, only because they are going away. But yes, I think a ticket makes sense and then we can discuss and refine it, thanks!

PR #917, opened 2025-03-04
VACMS-16756 Staff profile layout

chriskim2311:
# Description
This PR adds the staff profile layout and updates the query that pulls in the person_profile data that this layout uses.

## Ticket
[#16756](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/16756)

## Developer Task

```
- [ ] Alerted the #accelerated-publishing Slack channel to request a PR review.
```

## Testing Steps

## QA steps

## Screenshots
Before:
After:

# Reviewer

### Reviewing a PR

## Standard Checks

```
```

## Merging an Approved Layout


src/data/queries/staticPathResources.ts
```diff
@@ -56,11 +60,23 @@ export const formatter: QueryFormatter<
...
+    if (!resource.path) {
+      return false
+    }
+
+    // Filter out resources that don't have an alias
+    if (!resource.path.alias) {
+      return false
+    }
+    return true
+  })
```
cvalarida, line 74: Can we simplify it like this?
```suggestion
  // Filter out resources that don't have a path or path alias
  const filteredResources = resources.filter((resource) => Boolean(resource.path?.alias))
```

src/lib/constants/resourceTypes.ts
```diff
@@ -53,7 +54,7 @@ export const PARAGRAPH_RESOURCE_TYPES = {
   PROCESS_LIST: 'paragraph--process',
   REACT_WIDGET: 'paragraph--react_widget',
   RICH_TEXT_CHAR_LIMIT_1000: 'paragraph--rich_text_char_limit_1000',
-  STAFF_PROFILE: 'paragraph--staff_profile',
+  STAFF_PROFILE_PARAGRAPH: 'paragraph--staff_profile',
```
cvalarida, line 57: This isn't needed.

src/lib/drupal/facilityNoDrupalPageSideNav.ts
```diff
@@ -0,0 +1,23 @@
+import { SidebarData } from '@/types/formatted/staffProfile'
```
cvalarida, line 1: If this is going to be a utility in `src/lib/drupal/`, it shouldn't rely on a `staffProfile` type. Can we extract this `SidebarData` into a new module?

src/lib/drupal/facilityNoDrupalPageSideNav.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const buildSidebarData = (
```
cvalarida, line 3: This seems like a not specific enough name? 😕 

Can you either give it a clearer name (there's already a function to build some shape of sidebar data floating around here) or at least document this to make it clear when we'd use this function instead of the other?

src/lib/drupal/query.ts
```diff
@@ -125,6 +125,9 @@ export async function fetchSingleEntityOrPreview(opts, type, params) {
 
 // Helper function to return a consistent set of base fields for resources.
 export const entityBaseFields = (entity: NodeTypes): PublishedEntity => {
```
cvalarida, line 127:
```suggestion
export const entityBaseFields = (entity: NodeTypes): PublishedEntity | null => {
```

This makes sense, I think, but when we go to enable strict mode, I expect this will cause a lot of ruckus. 😅

src/templates/components/staffNewsProfile/index.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+describe('StaffNewsProfile', () => {
+  test('renders byline with title and description', () => {
+    const props = {
+      title: 'John Smith',
+      description: 'Staff Writer',
+    }
+    render(<StaffNewsProfile {...props} />)
+    expect(screen.getByText('By John Smith, Staff Writer')).toBeInTheDocument()
+    expect(screen.getByText('By John Smith, Staff Writer').tagName).toBe('P')
```
cvalarida, line 12: 🫠

src/templates/layouts/staffProfile/index.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+    expect(screen.getByText('Prachi Asher')).toBeInTheDocument()
+    expect(screen.getByText('Deputy Director')).toBeInTheDocument()
+    expect(
+      screen.getByText('VA Pittsburgh Healthcare System')
+    ).toBeInTheDocument()
+    expect(screen.getByText('Phone:')).toBeInTheDocument()
+    expect(screen.getByTestId('phone')).toBeInTheDocument()
+    expect(screen.getByText(mockProfile.introText)).toBeInTheDocument()
+    expect(screen.queryByText(/Download full size photo/)).toBeInTheDocument()
+    expect(screen.queryByText(/Download full bio/)).toBeInTheDocument()
```
cvalarida, line 82: It's a nit, but we should stick with either `getBy*` in these tests.

Technically, we could just `screen.getByText('whatever')` without the `expect` and call it a day because they'll throw an error and fail the test if the element isn't found, but our eslint rules get mad because they're a little naive. (Which is fine. 🤷)

`queryBy*` can return null without throwing an error, so we need to use that for the `expect().not.toBeInTheDocument()`s.

src/types/formatted/media.ts
```diff
@@ -14,8 +14,8 @@ export type MediaImage = {
   id: string
   alt: string
   title: string
-  width: number
-  height: number
+  width?: number
+  height?: number
```
cvalarida, line 18: Interesting. Why?

cvalarida (commented):
Not gonna lie, my eyes kinda glazed over after a while. 🫠 

There are a few places where I think we can clean it up, and I've got a few questions. For the bits where my eyes glazed over, I'm mostly trusting Laura's QA to catch anything blatantly bad. (I didn't pay close attention to the templates, so I assume they're correct if they pass Laura's inspection.)

src/assets/styles/globals.css
```diff
@@ -52,3 +52,8 @@ body {
   height: 42px;
   width: 45px;
 }
+
+/* staff profile styles */
+.staff-profile-image {
+  max-width: 160px;
+}
```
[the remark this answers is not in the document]
brianseek: I don't see it used anywhere else, I'm in favor of inlining it

src/data/queries/index.ts
```diff
@@ -68,7 +68,7 @@ export const QUERIES_MAP = {
   [`${RESOURCE_TYPES.STORY}--teaser` as const]: NewsStoryTeaser,
   [RESOURCE_TYPES.STORY_LISTING]: StoryListing,
   [RESOURCE_TYPES.QA]: QuestionAnswer,
-  [RESOURCE_TYPES.PERSON_PROFILE]: PersonProfile,
+  [RESOURCE_TYPES.STAFF_PROFILE]: PersonProfile,
```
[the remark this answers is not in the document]
brianseek: I think this was an issue of crossed wires. PERSON_PROFILE already existed and then Chris added STAFF_PROFILE targeting the same content type. I consolidated everything favoring Chris's naming since that was new...but PERSON_PROFILE is probably the better name space. I'm kind of 🤷 on it. What do you think?

src/data/queries/personProfile.ts
```diff
@@ -1,46 +1,84 @@
...
+  getMenu,
+} from '@/lib/drupal/query'
+import { buildSidebarData } from '@/lib/drupal/facilityNoDrupalPageSideNav'
 
-// // Define the option types for the data loader.
-// type DataOpts = QueryOpts<null>
+// Define the query params for fetching node--person_profile.
+export const params: QueryParams<null> = () => {
+  return new DrupalJsonApiParams().addInclude([
+    'field_media.image',
```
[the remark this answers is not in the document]
brianseek: I guess not, credit to Chris K on that 🤷

src/data/queries/staticPathResources.ts
```diff
@@ -18,15 +18,19 @@ const PAGE_SIZE = PAGE_SIZES.MAX
...
+    // Note: We can't put `path` first:
+    // See:
+    //  https://next-drupal.org/guides/page-limit
+    //  https://dsva.slack.com/archives/C01SR56755H/p1695244241079879?thread_ts=1695070010.697129&cid=C01SR56755H
+    .addFields(resourceType, ['field_administration', 'title', 'path'])
+    .addInclude(['field_administration'])
+  // Filter out profiles that shouldn't have a page.
+  if (resourceType === 'node--person_profile') {
+    params.addFilter('field_complete_biography_create', '1')
+  }
```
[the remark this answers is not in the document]
brianseek: I do think we could build something in to do this per type. However this might be the only additional param that we will ever add for the path resources. Time will tell as we start migrating more.

src/lib/constants/resourceTypes.ts
brianseek: There is a paragraph of staff profile, but we aren't handling it yet. I'll remove it for now.

src/data/queries/personProfile.ts
```diff
@@ -1,46 +1,84 @@
...
+export const formatter: QueryFormatter<PersonProfileData, StaffProfile> = (
+  options
+) => {
+  const entity = options?.entity
+  const menu = options?.menu
+  let formattedMenu = null
+  if (menu !== null)
+    formattedMenu = entity?.path
+      ? buildSidebarData(entity.title, entity.field_office.path.alias)
+      : null
```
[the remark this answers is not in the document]
brianseek: this is a fun bit of stuff that I'm not sure how it is intended to work. After digging in, menu is always undefined, which means that `menu !== null` is true. This is just a bad bit of code that I need to rework. I might need to study content build a bit to grasp it better.

src/lib/drupal/facilityNoDrupalPageSideNav.ts
brianseek: I renamed this because it is only related to staff profile the confusing name is from content-build and its only used in profiles over there.

src/lib/drupal/facilityNoDrupalPageSideNav.ts
brianseek: Renamed along with everything related to `facilityNoDrupalPageSideNav`

src/templates/layouts/staffProfile/index.test.tsx
brianseek: yep on the same page there, that was me rushing though tests

src/data/queries/index.ts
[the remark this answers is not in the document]
cvalarida: I also don't really care but would prefer if we stick with one. But then, I'm making `[RESOURCE_TYPES.VAMC_FACILITY]: HealthCareLocalFacility` 🫠 🤷 

We inherit some names and have no background... 🤷 

The fundamental struggle here is whether we use the machine name or the name folks actually _use._ To answer that question, I don't know which is better.

src/lib/constants/resourceTypes.ts
cvalarida: Er, specifically, I meant the rename isn't needed. `PARAGRAPH_RESOURCE_TYPES.STAFF_PROFILE` isn't likely to be conflicting with `RESOURCE_TYPES.STAFF_PROFILE` unless we're deconstructing both objects in the same scope for some reason, thus causing a naming conflict. 🤷

src/templates/components/staffNewsProfile/index.test.tsx
cvalarida: For what it's worth, I just rediscovered the `{ selector: 'p' }` option for these queries. So this could be simply `expect(screen.getByText('By John Smith, Staff Writer', { selector: 'p' })).toBeInTheDocument()`.

src/lib/drupal/staffProfileSideNav.ts
```diff
@@ -0,0 +1,23 @@
+import { SidebarData } from '@/types/formatted/staffProfile'
+
+export const buildStaffProfileSidebarData = (
+  title: string,
+  fieldOfficePath: string
+): SidebarData => {
+  const formattedTitle = title.replace(/"/g, '&quot;')
```
cvalarida, line 7: This is fine, but I wonder if it's even necessary now that we're rendering the title in a React element, which should handle it without this encoding, no problem. 🤷

src/templates/layouts/staffProfile/index.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+    expect(screen.getByText('Prachi Asher')).toBeInTheDocument()
+    expect(screen.getByText('Deputy Director')).toBeInTheDocument()
+    expect(
+      screen.getByText('VA Pittsburgh Healthcare System')
+    ).toBeInTheDocument()
+    expect(screen.getByText('Phone:')).toBeInTheDocument()
+    expect(screen.getByTestId('phone')).toBeInTheDocument()
+    expect(screen.getByText(mockProfile.introText)).toBeInTheDocument()
+    expect(screen.getByText(/Download full size photo/)).toBeInTheDocument()
+    expect(screen.getByText(/Download full bio/)).toBeInTheDocument()
```
cvalarida, line 82: Seems like snapshot testing might be a better approach here if all we're doing is checking that some expected stuff gets rendered? (But not even the value of the phone numbers or that the "download" elements are links, buttons, or marquee. 🙈 )

cvalarida (approved):
I left a couple comments, but feel free to ignore them and move on with your life; they're just nits at this point. 🙈

src/templates/layouts/staffProfile/index.test.tsx
brianseek: yeah it could be better, I just sort of tacked these on since this conditional wasn't being tested. I'll add a note to make this better in an upcoming defect ticket

src/lib/drupal/staffProfileSideNav.ts
brianseek: worth testing, i'll add this to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21160

src/templates/layouts/staffProfile/index.test.tsx
brianseek: to this: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21154

PR #921, opened 2025-03-05
Add a filter option to the api explorer

cvalarida:
# Description

Adds a filter option to the API explorer so we can find entities that have the data we need to reshape.

image

## Generated description
This pull request introduces a new filtering feature to the API Explorer, allowing users to add, remove, and update filters for their queries. The changes include modifications to both the frontend and backend to support this new functionality.

### Frontend changes in `src/pages/_playground/api-explorer.tsx`:

* Added a new `Filter` interface and integrated it into the `QueryState` interface. [[1]](diffhunk://#diff-f7fabf4e20e95c192c9124c7be6cd43bf315666f1c091c2be8f67fc03fcfba64R71-R76) [[2]](diffhunk://#diff-f7fabf4e20e95c192c9124c7be6cd43bf315666f1c091c2be8f67fc03fcfba64R90)
* Modified the initial state of `queryState` to include an empty `filters` array and changed the default `pageLimit` value.
* Implemented `addFilter`, `removeFilter`, and `updateFilter` functions to manage the filters in the state.
* Updated the `executeQuery` function to add filters to the query parameters before executing the query.
* Added UI elements to allow users to input, update, and remove filters dynamically.

### Backend changes in `src/pages/api/drupal-query.ts`:

* Added a new `Filter` interface to handle filter parameters in the API request.
* Modified the request handler to extract filters from the request body and add them to the query parameters. [[1]](diffhunk://#diff-113fe32dcc5b614d9c8db2e7c832aa34769eb5cca9d60457fe38b593bd816affL14-R20) [[2]](diffhunk://#diff-113fe32dcc5b614d9c8db2e7c832aa34769eb5cca9d60457fe38b593bd816affR36-R44)

### DIFF
--- a/src/pages/_playground/api-explorer.tsx
+++ b/src/pages/_playground/api-explorer.tsx
@@ -321,6 +321,12 @@ export default function ApiExplorer() {
           <label className="vads-u-display--block vads-u-margin-bottom--1">
             Filters
           </label>
+          {queryState.filters.length === 0 && (
+            <p>
+              <strong>Note:</strong> Filtering by fields on a paragraph entity
+              doesn&apos;t work right now.
+            </p>
+          )}
           {queryState.filters.map((filter, index) => (
             <div
               key={index}
--- a/src/pages/_playground/api-explorer.tsx
+++ b/src/pages/_playground/api-explorer.tsx
@@ -321,7 +321,7 @@ export default function ApiExplorer() {
           <label className="vads-u-display--block vads-u-margin-bottom--1">
             Filters
           </label>
-          {queryState.filters.length === 0 && (
+          {queryState.filters.length > 0 && (
             <p>
               <strong>Note:</strong> Filtering by fields on a paragraph entity
               doesn&apos;t work right now.
--- a/src/pages/_playground/api-explorer.tsx
+++ b/src/pages/_playground/api-explorer.tsx
@@ -68,6 +68,12 @@ const SyntaxHighlightedJson: React.FC<{ data: unknown }> = ({ data }) => {
   )
 }
 
+interface Filter {
+  field: string
+  operator: string
+  value: string
+}
+
 interface QueryState {
   resourceType: string
   includes: string[]
@@ -81,6 +87,7 @@ interface QueryState {
   }
   availableRelationships: string[]
   viewMode: 'tree' | 'raw'
+  filters: Filter[]
 }
 
 const AVAILABLE_RESOURCE_TYPES = {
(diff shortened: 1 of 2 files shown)


src/pages/_playground/api-explorer.tsx
```diff
@@ -125,35 +133,51 @@ export default function ApiExplorer() {
...
+  }
+
   const executeQuery = async () => {
     setQueryState((prev) => ({ ...prev, loading: true, error: null }))
     try {
-      const params = new DrupalJsonApiParams()
-        .addInclude(queryState.includes)
-        .addPageLimit(queryState.pageLimit)
-
-      const queryParams = params.getQueryObject()
```
cvalarida, line 135: We weren't actually using these `queryParams`. 🤷

src/pages/_playground/api-explorer.tsx
```diff
@@ -108,7 +116,7 @@ export default function ApiExplorer() {
   }
 
   const handleIncludesChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
-    const includes = e.target.value.split('\n').filter((line) => line.trim())
+    const includes = e.target.value.split('\n')
```
cvalarida, line 119: This made it so we couldn't add new lines to the input manually because they'd be filtered out immediately. I instead filter out blank lines when we go to _use_ the includes in the `params` during the `executeQuery` callback. Neglecting to do so results in a failed request, alas.

src/pages/_playground/api-explorer.tsx
```diff
@@ -293,6 +317,71 @@ export default function ApiExplorer() {
...
+                style={{ flex: 1 }}
+              >
+                <option value="=">=</option>
+                <option value="<>">≠</option>
+                <option value=">">&gt;</option>
+                <option value=">=">&gt;=</option>
+                <option value="<">&lt;</option>
+                <option value="<=">&lt;=</option>
+                <option value="IS NOT NULL">IS NOT NULL</option>
+                <option value="IS NULL">IS NULL</option>
```
cvalarida, line 352: Fun fact: [This is also how you'd test for empty arrays](https://www.drupal.org/docs/core-modules-and-themes/core-modules/jsonapi-module/filtering#s-11-filtering-on-empty-array-fields). 🫠

brianseek:
@cvalarida this is a really great enhancement! But it does seem to work only some of the time. I haven't dug in yet, but I am wondering if the response data is actually raw data or just rendered in some way. For example searching in `paragraph--phone_number` with a filter of `field_phone_label = Appointments` returns nothing even though that value is present when you run with no filter. `Appointments` could be the rendered label of a taxonomy or the value to a key. Not sure yet but will dig a bit more.

cvalarida:
Oh, interesting. Looks like I can't get any of the filters to work on paragraph types. They seem to on nodes, though. Any Drupal reason you can think of? [These docs](https://www.drupal.org/docs/core-modules-and-themes/core-modules/jsonapi-module/filtering) don't mention anything about it.


image
image

brianseek:
huh, weird. I didn't notice that it was just a paragraph thing in general. Maybe for now just slap a disclaimer on the filter and we will figure it out later ¯\_(ツ)_/¯

brianseek (approved):
Disclaimer looks great! :shipit:

ataker (approved):

PR #929, opened 2025-03-12
VAMC facility sidebar nav

cvalarida:
# Description
Adds the sidebar nav.

## Ticket
Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/20765

## Screenshots
image

## Generated description

This pull request introduces several changes to the `healthCareLocalFacility` data handling and formatting, including the addition of a side navigation menu. The most significant changes include updating the query to fetch menu data, modifying the data formatter to include the side navigation, and updating tests to reflect these changes.

### Changes to data fetching and formatting:

* [`src/data/queries/healthCareLocalFacility.ts`](diffhunk://#diff-35c09c75baec6b07a198e9d78d40d7db3a47ac9fc1c9a1e68a3566c1faf7303fR10-R18): Updated the query to fetch the `field_region_page` and dynamically fetch the menu based on the `field_office` reference. Added a new type `LocalFacilityData` to include both entity and menu data. Modified the data formatter to include the side navigation built from the fetched menu. [[1]](diffhunk://#diff-35c09c75baec6b07a198e9d78d40d7db3a47ac9fc1c9a1e68a3566c1faf7303fR10-R18) [[2]](diffhunk://#diff-35c09c75baec6b07a198e9d78d40d7db3a47ac9fc1c9a1e68a3566c1faf7303fR31-R40) [[3]](diffhunk://#diff-35c09c75baec6b07a198e9d78d40d7db3a47ac9fc1c9a1e68a3566c1faf7303fL40-R72)

### Changes to tests:

* [`src/data/queries/tests/__snapshots__/healthCareLocalFacility.test.tsx.snap`](diffhunk://#diff-209da7eaba46ff4f3c51597305f96e3c943c67839359c0e0533a1e1b099d5376R140-R143): Updated snapshot to include the `sideNav` field in the formatted data output.
* [`src/data/queries/tests/healthCareLocalFacility.test.tsx`](diffhunk://#diff-abaa973f859904255936f31061621151002986900215c47b41a4c6af371af61fL22-R28): Modified test to include the `menu` data in the formatData function call and added a new test to check if `sideNav` is added to the `window` object.

### Changes to API and component:

* [`src/pages/api/drupal-query.ts`](diffhunk://#diff-113fe32dcc5b614d9c8db2e7c832aa34769eb5cca9d60457fe38b593bd816affR4): Added authentication using username and password or client ID and secret for fetching resource collections. [[1]](diffhunk://#diff-113fe32dcc5b614d9c8db2e7c832aa34769eb5cca9d60457fe38b593bd816affR4) [[2]](diffhunk://#diff-113fe32dcc5b614d9c8db2e7c832aa34769eb5cca9d60457fe38b593bd816affL45-R60)
* [`src/templates/layouts/healthCareLocalFacility/index.tsx`](diffhunk://#diff-dfb29969d1983625591fe3e87de074afa2660839bba7565b48c78b1d25b3aeefR2-R30): Added logic to populate the `sideNav` data on the `window` object for use by a separate script.

### Changes to types:

* [`src/types/formatted/healthCareLocalFacility.ts`](diffhunk://#diff-f1c2e1972ee32d794f599ac7233134a307e92da47943ff8019fbe3442c585589R3-R9): Added `sideNav` field to the `HealthCareLocalFacility` type.


src/templates/layouts/healthCareLocalFacility/index.test.tsx
```diff
@@ -24,4 +32,16 @@ describe('HealthCareLocalFacility with valid data', () => {
...
+  // Once window.sideNav is populated, the static-pages app will render the menu
+  // in the nav[data-widget-type="side-nav"] element, but testing for that
+  // render would require side-loading that app bundle, which is outside the
+  // scope of this unit test
+  test('adds the sideNav to window.sideNav', () => {
+    render(<HealthCareLocalFacility {...mockData} />)
+
+    // @ts-expect-error window.sideNav is not a default window property, but
+    // we're adding it
+    expect(window.sideNav).toEqual(mockData.sideNav)
```
cvalarida, line 45: I could do the whole
```ts
// Allows additions to window object without overwriting global type
interface customWindow extends Window {
  sideNav?: SideNavMenu
}
declare const window: customWindow
```
thing, but 🤷

src/templates/layouts/healthCareLocalFacility/index.test.tsx
```diff
@@ -24,4 +32,16 @@ describe('HealthCareLocalFacility with valid data', () => {
       expect(screen.getByText(mockData[key])).toBeInTheDocument()
     )
   })
+
+  // Once window.sideNav is populated, the static-pages app will render the menu
+  // in the nav[data-widget-type="side-nav"] element, but testing for that
+  // render would require side-loading that app bundle, which is outside the
+  // scope of this unit test
```
cvalarida, line 39: `static-pages` is built in `vets-website`, so including it for this test is non-trivial. If we care deeply, I can figure out how to make that happen in a follow-up, though.

src/pages/api/drupal-query.ts
cvalarida, line 1: I can move this to a separate PR if'n y'like. It's only in here because I was using it to update the mock data. 🤷 

The story goes like this: Using just the client ID and secret, when we apply filters, it just sorta doesn't work. This is because Drupal let's us make the query with filters, but returns an empty data set and a 200 response code. Very coy. When we authenticate using username and password, it'll actually apply the filters and return us some data.

cvalarida:
@laflannery From what I can tell, that has nothing to do with this code, but rather the data in the review instance. I'm not sure where the review instance content comes from, but it looks like the menu data is...wonky? 🤷 When looking through the `content-build`'s sidebar nav code, I don't see any switch in there for handling Lovell side navs any differently, so that leads me to believe this isn't a case I'm not handling, but that it's an issue with the data we're getting for these.

image

Any idea how we might test that?

In the meantime, I'm gonna fiddle with a way to record requests, change the data, and replay the responses. I wanna see if I can change the menu data to match prod and see if that changes anything.

laflannery:
@cvalarida Can you look at the Lovell events page: https://www.va.gov/lovell-federal-health-care-va/events/ Does that sidenav have the parentID that you point out in those screenshots?

I tried to look but I'm having trouble finding it even on the Production page, what you have is different than what I have

cvalarida:
I said this all in Slack too, but for posterity...

---

Update on the Lovell side nav stuff:

Soooo...I don't really know what's going on yet. :sweat_smile: Here's what I did:

1. Set up MITM proxy in front of my locally-hosted CMS to record all requests and responses between NextJS (not the browser) and the CMS
1. Found the query for the side nav menu items
1. Navigated to that query on prod
   - Copied the JSON
1. Set up a request intercept for that particular query, replaced the contents with the prod data
   - I even saved the flows and replayed them for posterity
1. Observed that the data getting injected into window.sideNav is different locally than in prod :face_exhaling: 
   - I then copied the window.sideNav from prod into window.sideNav locally (instead of using the data returned from the query), and I see the menu as expected

What this tells me is that I've gotta do more spelunking in content-build to see how that data is manipulated and stuffed into window.sideNav, and how that's different from how we're doing it in next-build. :face_exhaling:

src/lib/drupal/staticProps.ts
```diff
@@ -89,7 +89,7 @@ export function getStaticPropsQueryOpts(
   // most types will simply need an ID, preview mode needs context
   const defaultQueryOpts = {
     id,
-    context: context.preview ? context : null,
+    context,
```
cvalarida, line 92: This function is only used in `getDefaultStaticPropsResource()`, and the return value is (eventually) passed through to `queries.getData()`. This populates the `opts.context` that gets passed through to our data fetcher functions in each query module.

image

As you can see, we're not actually using that context anywhere that I can tell, so I think the impact is minimal. (Though to be clear, I don't know why "preview mode needs context" as the comment above says, so there's probably stuff I'm missing. I don't _think_ always providing the context hurts anything, though. 🤞)

src/data/queries/healthCareLocalFacility.ts
```diff
@@ -37,16 +48,32 @@ export const data: QueryData<
...
+  LocalFacilityData,
   HealthCareLocalFacility
-> = (entity: NodeHealthCareLocalFacility) => {
+> = ({ entity, menu, lovell }) => {
+  let formattedMenu =
+    menu !== null ? buildSideNavDataFromMenu(entity.path.alias, menu) : null
+
+  if (lovell?.isLovellVariantPage) {
+    formattedMenu = getLovellVariantOfMenu(formattedMenu, lovell?.variant)
+  }
```
cvalarida, line 71: Instead of pulling it up a level and making some broad assumptions, it felt cleaner to me to format the menu specifically for Lovell right here in the data formatter function.

cvalarida (commented):
Fixed up the Lovell facility issue. 🎉 

image

laflannery:
@cvalarida This looks good to me, confirmed multiple non-Lovell facilities and as well as all the Lovell Facilities. I also looked at All existing templates (Events, News, Stories) and everything still loks the same as it dido

FYI I did create a new ticket for [Lovell breadcrumbs](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/20935) because the system crumbs needs to be updated

src/data/queries/healthCareLocalFacility.ts
```diff
@@ -24,11 +29,17 @@ export type HealthCareLocalFacilityDataOpts = {
   context?: ExpandedStaticPropsContext
 }
 
+type LocalFacilityData = {
```
moanu404, line 32: would it be helpful to add a JSdoc/blurb to this?

src/types/formatted/healthCareLocalFacility.ts
```diff
@@ -1,8 +1,10 @@
 import { FacilityOperatingStatusFlags } from '../drupal/node'
 import { PublishedEntity } from './publishedEntity'
+import { SideNavMenu } from './sideNav'
 
 export type HealthCareLocalFacility = PublishedEntity & {
   // Other attributes here
   introText: string | null
   operatingStatusFacility: FacilityOperatingStatusFlags
+  sideNav: SideNavMenu | null
```
moanu404, line 9: not a big deal but mostly curious if `sideNav?: SideNavMenu` or `introText?: string` might be more readable? i am no ts wiz though

src/templates/layouts/healthCareLocalFacility/index.tsx
```diff
@@ -1,15 +1,32 @@
...
 
 export function HealthCareLocalFacility({
   title,
   introText,
   operatingStatusFacility,
+  sideNav,
 }: FormattedHealthCareLocalFacility) {
+  // Populate the side nav data for the side nav widget to fill in
+  // Note: The side nav widget is in a separate app in the static-pages bundle
+  useEffect(() => {
```
moanu404, line 19: should we add a check if `sideNav` is null? not sure how much we foresee this use case

moanu404 (approved):
looks good, left a couple of questions, minor suggestions

src/templates/layouts/healthCareLocalFacility/index.tsx
cvalarida: I'm sure it'll happen, but that's alright. If it's `null`, the side nav just won't get rendered. I'll double-check that it doesn't break unexpectedly, though. (Another `null` vs `property doesn't exist` issue—not to be confused with `undefined`, of course. 🫠 )

src/types/drupal/node.ts
```diff
@@ -153,8 +153,8 @@ export interface NodeHealthCareLocalFacility extends DrupalNode {
   field_operating_status_facility: FacilityOperatingStatusFlags
   field_operating_status_more_info: string
   field_phone_number: string
-  // TODO: Uncomment this
-  // field_region_page: NodeHealthCareRegionPage
+  // We don't need the field_media in here for now ¯\_(ツ)_/¯
```
timcosgrove, line 156: Remove the shrug :)

envs/.env.example
```diff
@@ -25,6 +25,10 @@ DRUPAL_PREVIEW_SECRET=secret
 DRUPAL_CLIENT_ID=Retrieve this from AWS SSM /cms/consumers/next-build/client_id
 DRUPAL_CLIENT_SECRET=Retrieve this from AWS SSM /cms/consumers/next-build/client_secret
 
+# For API explorer, so we can filter
+# DRUPAL_USERNAME=username@email.com
+# DRUPAL_PASSWORD=some-password
+
```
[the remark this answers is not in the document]
cvalarida: The client ID + secret _does_ act as authentication just fine. My current credentials are, however, associated with a user that doesn't have proper permissions, like you say, and I didn't know how to update them.

It's totally unrelated to the rest of this PR, so I can rip this out and push up another PR for it if we should have a more in-depth conversation about this. 👍

src/pages/api/drupal-query.ts
```diff
@@ -42,12 +43,21 @@ export default async function handler(
       }
     })
 
+    // The username & password will let us apply filters, so prefer that
```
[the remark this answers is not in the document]
cvalarida: Any time I applied any filters (except the `IS NULL` or `IS NOT NULL` sometimes?), I'd get a 200 response with an empty data set. 😕 With proper permissions (username + password), the filters actually worked as expected.

timcosgrove (dismissed):

moanu404 (approved):
🎉

timcosgrove (approved):

src/data/queries/healthCareLocalFacility.ts
moanu404: So I tried that initially and it did not work (I cannot pull up that terminal log alas) but it seemed to require me to have `date-fns` in addition for it to pass, which was weird. I'll try to dig it up some more and see what exact error it was showing me

PR #959, opened 2025-03-31
VACMS 19771 feedback analytics

brianseek:
# Description

A bit of a lift and shift of the medallia event listeners from content-build

## Generated description

This pull request includes several updates to the `MedalliaAssets` component and its associated tests to improve event handling and analytics recording. The most important changes include adding the `recordEvent` function, implementing event listeners for various Medallia events, and updating the test cases to mock and verify these events.

### Improvements to Event Handling and Analytics:

* [`src/templates/common/medallia/index.tsx`](diffhunk://#diff-5cdda7136985d4843bf8f9b7f45b1ea45c1e29d17657c729f9fae7a9ad47266cR10): Added the `recordEvent` function and implemented event listeners for various Medallia events, such as `ShowForm_Called`, `Form_Displayed`, `Feedback_Submit`, and others, to record analytics data. [[1]](diffhunk://#diff-5cdda7136985d4843bf8f9b7f45b1ea45c1e29d17657c729f9fae7a9ad47266cR10) [[2]](diffhunk://#diff-5cdda7136985d4843bf8f9b7f45b1ea45c1e29d17657c729f9fae7a9ad47266cR34-R122)

### Test Updates:

* [`src/templates/common/medallia/index.test.tsx`](diffhunk://#diff-3b60010f20560fdedac940fdb7a82f612412c7f0eaf08631f1477367e64641bdL3-R3): Removed unnecessary `useEffect` mock, added a mock for `recordEvent`, and implemented new test cases to verify that event listeners are correctly added and that `recordEvent` is called with the appropriate data when Medallia events are triggered. [[1]](diffhunk://#diff-3b60010f20560fdedac940fdb7a82f612412c7f0eaf08631f1477367e64641bdL3-R3) [[2]](diffhunk://#diff-3b60010f20560fdedac940fdb7a82f612412c7f0eaf08631f1477367e64641bdR18-R89) [[3]](diffhunk://#diff-3b60010f20560fdedac940fdb7a82f612412c7f0eaf08631f1477367e64641bdR101-R177)

### Minor Changes:

* [`src/templates/common/contentFooter/index.tsx`](diffhunk://#diff-0a6e6061c4cd09c4c1592fcbd3f5ca9047711609e4f1b19e5398aac4116ddb4fR80): Added the `disable-analytics` attribute to the "Give feedback" button.
## Ticket

Closes [19771](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/19771)

## Developer Task

## Testing Steps

 - download an extension that can observe GA4 events like [this one for chrome](https://chromewebstore.google.com/detail/debugger-for-google-analy/nomnoimacbncclfbnfaingniikblfbji?hl=en-US)
 - visit https://www.va.gov/boston-health-care/news-releases/
 - interact with the Feedback button at the bottom of the page (open, close, submit) and note the GA events
 - visit https://pr959-mlloqgh2pg7kaxzayeythni2ydhhqgyp.tugboat.vfs.va.gov/boston-health-care/news-releases/
 - interact with the Feedback button in the same way you did on the live page, and observe that the GA events match

---

### Reviewing a PR

## Standard Checks

### DIFF
--- a/src/templates/common/contentFooter/index.tsx
+++ b/src/templates/common/contentFooter/index.tsx
@@ -77,6 +77,7 @@ export function ContentFooter({
           <va-button
             label="Give feedback"
             id="mdFormButton"
+            disable-analytics
             onClick={() => {
               const isProduction =
                 process.env.NEXT_PUBLIC_BUILD_TYPE === BUILD_TYPES.PROD
--- a/src/templates/common/medallia/index.test.tsx
+++ b/src/templates/common/medallia/index.test.tsx
@@ -78,7 +78,7 @@ const mockEvent: MedalliaEvent = {
 }
 
 describe('MedalliaAssets Component', () => {
-  beforeEach(() => {
+  afterEach(() => {
     jest.clearAllMocks()
   })
   test('renders Script component with correct props', () => {
@@ -130,8 +130,6 @@ describe('MedalliaAssets Component', () => {
       'survey-status': 'Test Form',
       'survey-details': { feedbackUUID: 'abc-123', content: 'Test Content' },
     })
-
-    addEventListenerSpy.mockRestore()
   })
   test('adds listener and calls recordEvent when triggered for Medallia event with label', () => {
     const mockEventDetail = {
@@ -167,7 +165,5 @@ describe('MedalliaAssets Component', () => {
       'survey-status': 'Invite',
       'survey-details': {},
     })
-
-    addEventListenerSpy.mockRestore()
   })
 })
--- a/src/templates/common/medallia/index.test.tsx
+++ b/src/templates/common/medallia/index.test.tsx
@@ -82,11 +82,6 @@ describe('MedalliaAssets Component', () => {
     jest.clearAllMocks()
   })
   test('renders Script component with correct props', () => {
-    // Mocking useEffect
-    jest.mock('react', () => ({
-      ...jest.requireActual('react'),
-      useEffect: jest.fn(),
-    }))
     process.env.NEXT_PUBLIC_BUILD_TYPE = 'local'
 
     render(<MedalliaAssets />)
--- a/src/templates/common/medallia/index.test.tsx
+++ b/src/templates/common/medallia/index.test.tsx
@@ -1,11 +1,6 @@
 import { render, screen } from '@testing-library/react'
 import { MedalliaAssets } from './'
-
-// Mocking useEffect
-jest.mock('react', () => ({
-  ...jest.requireActual('react'),
-  useEffect: jest.fn(),
-}))
+import { recordEvent } from '@/lib/analytics/recordEvent'
 
 // Mock Script
 jest.mock('next/script', () => ({ id, src }) => (
@@ -20,8 +15,78 @@ jest.mock('@/lib/utils/medallia', () => ({
   setWindowVaSurvey: jest.fn(),
 }))
 
+// Mock recordEvent
+jest.mock('@/lib/analytics/recordEvent', () => ({
+  recordEvent: jest.fn(),
+}))
+
+interface MedalliaEvent extends CustomEvent {
+  detail: {
+    Form_Type?: string
+    Form_ID?: string
+    Feedback_UUID?: string
+    Content?: string
+  }
+}
+
+const mockEventDetail: MedalliaEvent = {
+  detail: {},
+  initCustomEvent: function (
+    type: string,
+    bubbles?: boolean,
+    cancelable?: boolean,
+    detail?: unknown
+  ): void {
+    throw new Error('Function not implemented.')
+  },
+  bubbles: false,
+  cancelBubble: false,
+  cancelable: false,
+  composed: false,
+  currentTarget: undefined,
+  defaultPrevented: false,
+  eventPhase: 0,
+  isTrusted: false,
+  returnValue: false,
+  srcElement: undefined,
+  target: undefined,
+  timeStamp: 0,
+  type: '',
+  composedPath: function (): EventTarget[] {
+    throw new Error('Function not implemented.')
+  },
+  initEvent: function (
+    type: string,
+    bubbles?: boolean,
+    cancelable?: boolean
+  ): void {
+    throw new Error('Function not implemented.')
+  },
+  preventDefault: function (): void {
+    throw new Error('Function not implemented.')
+  },
+  stopImmediatePropagation: function (): void {
+    throw new Error('Function not implemented.')
+  },
+  stopPropagation: function (): void {
+    throw new Error('Function not implemented.')
+  },
+  NONE: 0,
+  CAPTURING_PHASE: 1,
+  AT_TARGET: 2,
+  BUBBLING_PHASE: 3,
+}
+
 describe('MedalliaAssets Component', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
   test('renders Script component with correct props', () => {
+    // Mocking useEffect
+    jest.mock('react', () => ({
+      ...jest.requireActual('react'),
+      useEffect: jest.fn(),
+    }))
     process.env.NEXT_PUBLIC_BUILD_TYPE = 'local'
 
     render(<MedalliaAssets />)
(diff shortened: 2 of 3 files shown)


src/templates/common/contentFooter/index.tsx
```diff
@@ -77,6 +77,7 @@ export function ContentFooter({
           <va-button
             label="Give feedback"
             id="mdFormButton"
+            disable-analytics
```
cvalarida, line 80: 🤔 Why disable analytics on this button?

src/templates/common/medallia/index.test.tsx
```diff
@@ -1,11 +1,6 @@
 import { render, screen } from '@testing-library/react'
 import { MedalliaAssets } from './'
-
-// Mocking useEffect
-jest.mock('react', () => ({
-  ...jest.requireActual('react'),
-  useEffect: jest.fn(),
-}))
```
cvalarida, line 8: 🫠 🙈

src/templates/common/medallia/index.tsx
cvalarida, line 1: I know this is outside the scope of this PR, so these are non-blocking questions for sure...

Do you have any idea what the story is with that `scriptId`? It seems pretty...opaque. 😕 

Also, are we confident that using `NEXT_PUBLIC_BUILD_TYPE` is the correct way to determine whether Medallia should load? 🤔 What about `APP_ENV`?

src/templates/common/medallia/index.test.tsx
```diff
@@ -20,8 +15,78 @@ jest.mock('@/lib/utils/medallia', () => ({
...
+}))
+
+interface MedalliaEvent extends CustomEvent {
+  detail: {
+    Form_Type?: string
+    Form_ID?: string
+    Feedback_UUID?: string
+    Content?: string
+  }
+}
```
cvalarida, line 30: 🤔 Would this benefit from pulling out into another module, or do you suppose this is the only place we'd need to define the shape of a Medallia event? (Something something useful for defining mock data for testing, but the implementation details are hidden from actual usage..?)

src/templates/common/medallia/index.test.tsx
```diff
@@ -20,8 +15,78 @@ jest.mock('@/lib/utils/medallia', () => ({
...
+  defaultPrevented: false,
+  eventPhase: 0,
+  isTrusted: false,
+  returnValue: false,
+  srcElement: undefined,
+  target: undefined,
+  timeStamp: 0,
+  type: '',
+  composedPath: function (): EventTarget[] {
+    throw new Error('Function not implemented.')
```
cvalarida, line 56: 😂 That works, I guess.

src/templates/common/medallia/index.test.tsx
```diff
@@ -20,8 +15,78 @@ jest.mock('@/lib/utils/medallia', () => ({
...
+  stopImmediatePropagation: function (): void {
+    throw new Error('Function not implemented.')
+  },
+  stopPropagation: function (): void {
+    throw new Error('Function not implemented.')
+  },
+  NONE: 0,
+  CAPTURING_PHASE: 1,
+  AT_TARGET: 2,
+  BUBBLING_PHASE: 3,
```
cvalarida, line 77: Are these meaningful? Do we have to test against different values for any reason, or can we just rely on Medallia to do Medallia things and handle all this appropriately?

src/templates/common/medallia/index.test.tsx
```diff
@@ -20,8 +15,78 @@ jest.mock('@/lib/utils/medallia', () => ({
...
 describe('MedalliaAssets Component', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
   test('renders Script component with correct props', () => {
+    // Mocking useEffect
+    jest.mock('react', () => ({
+      ...jest.requireActual('react'),
+      useEffect: jest.fn(),
+    }))
```
cvalarida, line 89: 😕 Can we just remove this bit if we're redefining `process.env.NEXT_PUBLIC_BUILD_TYPE`? Or does redefining that environment variable not propagate through to the rendered component?

...and...why do you not have to mock it for your newly-added test below? Seems like we don't actually need this bit...or am I missing something? 🤔

src/templates/common/medallia/index.test.tsx
```diff
@@ -33,4 +98,81 @@ describe('MedalliaAssets Component', () => {
...
+    )
+
+    // Simulate the event listener being triggered
+    const eventHandler = addEventListenerSpy.mock.calls.find(
+      (call): call is [string, (event: CustomEvent) => void] =>
+        call[0] === 'MDigital_Feedback_Submit'
+    )?.[1]
+    if (eventHandler) {
+      eventHandler(mockEventDetail)
+    }
```
cvalarida, line 128: Whoa, this is fancy. Just to make sure I understand:
1. When we render the component, it fires off that new `useEffect()`
2. This `useEffect()` adds some event listeners
3. We're finding a specific event listener by hunting through the calls to our `addEventListener` spy for a specific one, then calling it with a mocked event, thereby simulating sending that actual event, which would normally trigger the listener if we _actually_ added it to `window`'s event listeners

Not sure I fully understand the need for that type guard, but I assume it's useful because of Reasons™. 😅

src/templates/common/medallia/index.tsx
```diff
@@ -30,6 +31,95 @@ export function MedalliaAssets() {
...
+      label: 'Invite',
+    })
+    mEvent({
+      name: 'Invite_Skipped',
+      action: 'survey-invitation-skip',
+      label: 'Invite',
+    })
+    return () => {
+      controller.abort()
+    }
```
cvalarida, line 120: When this component is unmounted, it'll abort any in-flight events, which is great, but the event listeners will technically still be attached to `window`.

For now, that's fine since a full page load will nuke `window` and all its contents. This will happen every time we navigate to a new page because we're serving static pages from S3. After we move to a persistent NextJS server, though, I think the NextJS router won't require a full page reload when navigating between pages _(that's maybe not true if we're only using `<a>` elements and not `<Link>` elements, mind you—I'm not super confident there)_. If that's true, then we'll need to clean up these event handlers here too or we risk adding them multiple times when navigating between pages.

My paranoia wants me to ask you to remove the event listeners in this cleanup step too. 😅

cvalarida (commented):
Looks great! That didn't, of course, stop me from leaving a bunch of comments. 😅 

I think the only one I'd say is stopping me from smashing that approve (radio) button is the event listener cleanup bit at the end.

src/templates/common/contentFooter/index.tsx
brianseek: This is the standard practice when we are doing custom analytics on a button click. So the `survey-show-form-call` is the replacement for this default event.

laflannery: Weirdly enough - we are missing a standard `event.cta-button-click` that exists on Prod but not on next build. I'm not convinced this matters now though because we have everything else. Let me track this down for a few mins

src/templates/common/medallia/index.tsx
brianseek: I believe that the scriptId is so that Medallia is loading the dev version of these forms on non-prod sites. As far as using `NEXT_PUBLIC_BUILD_TYPE` I'm not sure. My gut says that it's probably correct because Tim did the change, but that is not a wise path since we can all make mistakes on things. I guess the answer is that I didn't really thing about it since this stuff seems to have been working thus far for next-build.

src/templates/common/medallia/index.test.tsx
brianseek: Good question, I would bet some decent money that we won't do this anywhere else ever again. So I'm 50/50 on it.

src/templates/common/medallia/index.test.tsx
brianseek: my fiend vscode `add missing properties` added all these missing properties for me, so I can't take credit for that 😄

src/templates/common/medallia/index.test.tsx
brianseek: This was another thing that was added with `add missing properties` we don't interact with that part of the event at all. I'm guessing that these are default values. I wouldn't have added them at all if typescript linter didn't cry about it.

src/templates/common/medallia/index.test.tsx
brianseek: That is a good point, I was trying to leave un changed things intact. I'll test without it all together. My gut says its not needed.

src/templates/common/medallia/index.test.tsx
brianseek: Yes you are 100% correct in your understanding. The type guard is in place because the typescript linter wanted it 🤷  It all looked much simpler when I first added it.

src/templates/common/medallia/index.tsx
brianseek: according to this https://kettanaito.com/blog/dont-sleep-on-abort-controller having the signal abort controller added to all the event listeners allows you to remove them all with the abort in the return.

I went looking for a way to do that because manually adding all of those removes made me cry.

src/templates/common/medallia/index.test.tsx
```diff
@@ -33,4 +98,81 @@ describe('MedalliaAssets Component', () => {
...
+    // Assert that recordEvent is called with the correct data
+    expect(recordEvent).toHaveBeenCalledWith({
+      event: 'survey-submit',
+      'survey-tool': 'Medallia',
+      'survey-form-id': '12345',
+      'survey-status': 'Test Form',
+      'survey-details': { feedbackUUID: 'abc-123', content: 'Test Content' },
+    })
+
+    addEventListenerSpy.mockRestore()
```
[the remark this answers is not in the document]
brianseek: If I added this to an afterEach, we would still have that issue, right? What is the way to properly clean up in the case of an error?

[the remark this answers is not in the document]
cvalarida: No, `afterEach` is the "right" way to do it. It'll get called even if the test fails.

src/templates/common/contentFooter/index.tsx
laflannery: @brianseek I think this is fine to leave disabled, like you said it's standard practice because it tends to prevent duplicate events being fired. If we find out later that it should be enable for whatever reason we can always change it

laflannery:
This all looks good to me, I compared the events being fired on button click, modal close (using the X and the buttons), and modal submit and they are all the same as what is currently on Prod now

cvalarida (dismissed):
That abort controller removing the listeners is pretty slick. Thanks for tracking that down!

I'd still love a comment explaining that and a link to MDN as proof, but I'm approving it anyhow. Great work!

src/templates/common/contentFooter/index.tsx
brianseek: if it exists on prod, that just means they are implementing that incorrectly

brianseek:
> That abort controller removing the listeners is pretty slick. Thanks for tracking that down!
> 
> I'd still love a comment explaining that and a link to MDN as proof, but I'm approving it anyhow. Great work!

thanks and added comment!

cvalarida (approved):

PR #965, opened 2025-04-03
VADS grid migration script

cvalarida:
# Description

Adds a script to migrate to v3 grid.

## Generated description

This pull request introduces a new script `migrate-grid.mjs` to automate the migration of CSS class names from the old grid system to the new one, along with a dry-run feature for testing changes without modifying files. Below are the key changes:

**New Script Addition:**

* [`scripts/migrate-grid.mjs`](diffhunk://#diff-c4661b05cdb67e59b36fbf247a2998bc15ce8a516790cf6abbd82eacc1eaf742R1-R137): Added a new script to handle the migration of CSS class names with specific mappings, prefix replacements, and breakpoint replacements. This script also includes functionality to warn about deprecated Foundation classes and remove duplicate class names.

**Migration Functionality:**

* Implemented exact, prefix, and breakpoint replacements for CSS class names to ensure smooth transition to the new grid system.
* Added a warning system to identify and notify about deprecated Foundation classes without replacing them.

**File Processing:**

* Included a dry-run option to preview changes without modifying the actual files, and provided detailed output for proposed changes during the dry run.
* Ensured the script checks for file existence before processing and handles errors appropriately.

### DIFF
--- a/scripts/migrate-grid.mjs
+++ b/scripts/migrate-grid.mjs
@@ -1,0 +1,137 @@
+#!/usr/bin/env node
+
+/* eslint-disable no-console */
+
+import fs from 'fs'
+import chalk from 'chalk'
+
+const args = process.argv.slice(2)
+const dryRun = args.includes('--dry-run')
+const files = args.filter((arg) => !arg.startsWith('--'))
+
+if (files.length === 0) {
+  console.error('❌ No input files provided.')
+  process.exit(1)
+}
+
+// --- Mapping Rules ---
+
+const exactReplacements = {
+  'vads-l-grid-container': 'vads-grid-container',
+  'vads-l-row': 'vads-grid-row',
+  'usa-grid-full': 'vads-grid-container', // Needs to come before usa-grid because replacing usa-grid first will mess this one up
+  'usa-grid': 'vads-grid-container',
+  'usa-width-one-half': 'vads-grid-col-6',
+  'usa-width-one-third': 'vads-grid-col-4',
+  'usa-width-two-thirds': 'vads-grid-col-8',
+  // Add more as needed
+}
+
+const prefixReplacements = {
+  'vads-l-col--': 'vads-grid-col-',
+}
+
+const breakpointReplacements = {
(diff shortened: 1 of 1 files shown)


scripts/migrate-grid.mjs
cvalarida, line 1: `.mjs` so we can `import` chalk. 🤷

scripts/migrate-grid.mjs
```diff
@@ -0,0 +1,137 @@
...
+  console.error('❌ No input files provided.')
+  process.exit(1)
+}
+
+// --- Mapping Rules ---
+
+const exactReplacements = {
+  'vads-l-grid-container': 'vads-grid-container',
+  'vads-l-row': 'vads-grid-row',
+  'usa-grid-full': 'vads-grid-container', // Needs to come before usa-grid because replacing usa-grid first will mess this one up
```
cvalarida, line 22: There's probably a better way to fix this, but it was easier this way. 🤷

cvalarida (commented):
Output looks like this.
image

image

brianseek (approved):
Neat!

scripts/migrate-grid.mjs
```diff
@@ -0,0 +1,137 @@
...
+  'small-screen:vads-l-': 'mobile-lg:vads-grid-',
+  'medium-screen:vads-l-': 'tablet:vads-grid-',
+  'small-desktop-screen:vads-l-': 'desktop:vads-grid-',
+  'large-screen:vads-l-': 'desktop-lg:vads-grid-',
+}
+
+const foundationRegexes = [
+  /\b(row|column|columns)\b/g,
+  /\bsmall-[\w-]+\b/g,
+  /\bmedium-[\w-]+\b/g,
```
micahchiang, line 45: Only thought here is whether or not it'd be better to just look for numeric values instead of alphanumeric. Quick search in vets-website shows we might have several instances of other random classes that are prefixed with `medium-` but end with things other than numbers. Would using this pattern pick those things up?

micahchiang (approved):

PR #970, opened 2025-04-04
Script to move a product to a new directory

cvalarida:
# Description

Makes code ownership cleaner and groups all the things for a particular "product" into a single place.

## Generated description

This pull request includes significant changes to the script for moving products and updates to import paths in the `src/data/queries/index.ts` file. The most important changes are the addition of a new script for moving products and the adjustment of import paths to use absolute paths.

Changes to the script:

* [`scripts/move-product.sh`](diffhunk://#diff-f5023e4dfd8fde8564994591e48c73d57745d2d53479ac6ee9526999b3d07827R1-R209): Added a new script to facilitate moving products within the project. This script includes CLI arguments for dry-run mode and product name, color-coded logging, and functions to update import paths and move relevant files.

Updates to import paths:

* [`src/data/queries/index.ts`](diffhunk://#diff-4736203072bd77be64e5041b8727a153d8cb72e7ffec7c999324edb7a43c44d2L2-R50): Updated import paths to use absolute paths starting with `@/data/queries/` instead of relative paths. This change affects multiple imports within the file.


scripts/move-product.sh
```diff
@@ -0,0 +1,209 @@
...
+
+success() {
+  echo -e "${GREEN}✅ $1${RESET}"
+}
+
+error() {
+  echo -e "${RED}❌ $1${RESET}"
+}
+
+update_all_imports() {
```
cvalarida, line 97: It's a little deceiving, actually, but `update_some_imports()` felt a little lackluster. 😜 

Mostly, though, the trouble comes with relative imports. I could try to get clever in here, but it's really not a big lift to manually change these. The tool runs `tsc --noEmit` too, so it'll show which ones still need to be updated.

scripts/move-product.sh
```diff
@@ -0,0 +1,209 @@
...
+  CODEOWNERS_LINE="src/products/${PRODUCT_NAME}/ @department-of-veterans-affairs/ap-team @department-of-veterans-affairs/va-platform-cop-frontend"
+  log "Adding CODEOWNERS entry: $(color "$CODEOWNERS_LINE" green)"
+  run_or_echo "echo \"$CODEOWNERS_LINE\" >> '$CODEOWNERS_FILE'"
+  $DRY_RUN || success "CODEOWNERS updated."
+else
+  warn "CODEOWNERS file not found at '$(color "$CODEOWNERS_FILE" gray)'"
+fi
+
+echo ""
+success "Done moving product '$(color "$PRODUCT_NAME" cyan)$(color "'!" green)"
```
cvalarida, line 208: Yeah, alright, so it's a little optimistic. 😅 🤷

src/data/queries/index.ts
cvalarida, line 1: All these imports are the same, but more easily rewritten via the script. Feels a little silly to use the path alias instead of the relative path, but 🤷

scripts/move-product.sh
```diff
@@ -0,0 +1,209 @@
...
+    i=$((i + 1))
+
+    log "Replacing imports from $(color "$search" yellow) → $(color "$replace" green)"
+
+    # Escape slashes in search and replace patterns
+    search_escaped=$(printf '%s\n' "$search" | sed 's/[\/&]/\\&/g')
+    replace_escaped=$(printf '%s\n' "$replace" | sed 's/[\/&]/\\&/g')
+
+    # Find and replace with word boundaries
+    grep -rl --include="*.ts*" --include="*.js*" "$search" ./src | while read -r file; do
```
cvalarida, line 131: Technically, things could get a little silly if we've got a JSON file with a `@/templates/layouts/event` or whatever. 😅 🤷

scripts/move-product.sh
```diff
@@ -0,0 +1,209 @@
...
+      log "↪ Updating file: $(color "$file" cyan)"
+      run_or_echo "sed -i '' -E \"s|$search_escaped([^a-zA-Z])|$replace_escaped\1|g\" \"$file\""
```
cvalarida, line 133: The extra little `[^a-zA-Z]` is to make sure it matches only `event` imports instead of `eventListing` when the product name is `event`.

scripts/move-product.sh
```diff
@@ -0,0 +1,209 @@
...
+    done
+  done
+}
+
+# === Main ===
```
cvalarida, line 138: This bit gets a little wordy. Open to suggestions on how to clean it up. Mostly, I just let ChatGPT do its thing. 🤷 Again, it's temporary, so being a little wordy is fine.

brianseek (approved):
This all looks fine to me. I don't fully understand it, maybe we can run through it tomorrow, but I don't want to hold you up on things. I can confirm that it doesn't break anything 😀

brianseek (approved):

PR #974, opened 2025-04-10
VAMC Facility JSON-LD bit

cvalarida:
# Description

Adds the Place JSON-LD thing.

It needed some extra data I wasn't grabbing yet, so along the way, I had to change the static type for VAMC facility, grab that data, and put it in the `formatter()` output. This had some knock-on effects for other types and tests.

I also couldn't get the same page as I had in the mock data for the content type, so I replaced it entirely when I updated it to add all the new fields that weren't there before. 😬 Thus the huge line count, because mock data and snapshot updates. 🫠 

I _also_ discovered that the way we're constructing the JSON-LD data for `Events` probably needs to change. I mostly copied that bit over to VAMC facilities, but when I compared the output with what's in prod right now, I found a few things:
- The open and close hours were numbers instead of a Time ([as defined by the JSON schema specification for Place](https://schema.org/Place))
- The `dayOfWeek` is probably wrong (using a 0-indexed number instead of the name of the day)

Additionally, there are a few changes from prod that are either fine or better. The diff (lines that are the same are collapsed for brevity):
image

- The `hasMap` URL is HTML-encoded (I tested it and it works the same, as you'd expect)
- The `image` property is now a URL instead of an array with a single URL
   - [According to the spec](https://schema.org/image), this is correct; prod has it wrong
- The `geo.latitude` and `geo.longitude` are numbers instead of strings
   - [According to the spec](https://schema.org/GeoCoordinates), this can be either Text or Number, so we should be fine

## Generated description

This pull request introduces several changes to the `healthCareLocalFacility` module, including enhancements to data formatting, new utility functions, and the addition of structured data for schema.org. The key changes are grouped by theme below:

### Data Formatting Enhancements:
* Added new fields to the `formatter` in `healthCareLocalFacility.ts` to include `address`, `phoneNumber`, `officeHours`, `image`, `facilityLocatorApiId`, and `geoLocation`. [[1]](diffhunk://#diff-35c09c75baec6b07a198e9d78d40d7db3a47ac9fc1c9a1e68a3566c1faf7303fR81-R82) [[2]](diffhunk://#diff-35c09c75baec6b07a198e9d78d40d7db3a47ac9fc1c9a1e68a3566c1faf7303fR91-R94)
* Updated `params` in `healthCareLocalFacility.ts` to include `field_media` and `field_media.image`.

### Utility Functions:
* Added `dayOfWeek` function and corresponding tests to convert numeric day representation to day names. [[1]](diffhunk://#diff-937aff64e402ec119f4a6c6ee146648de17f5fcf694d213d1a48ce8de72e1207R1-R16) [[2]](diffhunk://#diff-4befb3546f63b22112c9c449ec2ebf8ed5c0e2407a644e582447202638d9249aR1-R37)
* Added `numberToTimeString` function and corresponding tests to format numeric time representation to a string. [[1]](diffhunk://#diff-8c7bd13d66801a21a866e278c63ff1ccf3c88b93f6fb48c105bd271225554f3dR1-R18) [[2]](diffhunk://#diff-191f5b38536d50354770b12651126a5129e79de37ebde737f5f62cf15c6e6da9R1-R39)

### Structured Data and Schema.org:
* Integrated structured data for schema.org in the `HealthCareLocalFacility` component, including address, phone number, office hours, and geo-location. [[1]](diffhunk://#diff-dfb29969d1983625591fe3e87de074afa2660839bba7565b48c78b1d25b3aeefR27-R73) [[2]](diffhunk://#diff-dfb29969d1983625591fe3e87de074afa2660839bba7565b48c78b1d25b3aeefL72-R120)

### Test and Mock Data Updates:
* Updated `index.test.tsx` for `HealthCareLocalFacility` to use mock data and the new formatter.
* Added mock data for `field_geolocation` in `event/index.test.tsx`.

### Type Definitions:
* Updated type definitions in `node.ts` and `healthCareLocalFacility.ts` to include new fields such as `field_geolocation`, `field_office_hours`, `address`, `phoneNumber`, `image`, and `facilityLocatorApiId`. [[1]](diffhunk://#diff-a5035ec5110b73d89f63d6fbe12f3bd0af37cf28d17289acfd0bdb8c0932ccbbR147-R154) [[2]](diffhunk://#diff-f1c2e1972ee32d794f599ac7233134a307e92da47943ff8019fbe3442c585589R1-R10) [[3]](diffhunk://#diff-f1c2e1972ee32d794f599ac7233134a307e92da47943ff8019fbe3442c585589R19-R24)

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/20774

### DIFF
--- a/src/templates/layouts/healthCareLocalFacility/index.test.tsx
+++ b/src/templates/layouts/healthCareLocalFacility/index.test.tsx
@@ -1,37 +1,37 @@
 import { render, screen } from '@testing-library/react'
+import drupalMockData from '@/mocks/healthCareLocalFacility.mock.json'
 import { HealthCareLocalFacility } from './index'
 import { HealthCareLocalFacility as FormattedHealthCareLocalFacility } from '@/types/formatted/healthCareLocalFacility'
+import { formatter } from '@/data/queries/healthCareLocalFacility'
+import { DrupalMenuLinkContent } from 'next-drupal'
 
-const mockData: FormattedHealthCareLocalFacility = {
-  id: '123',
-  title: 'Test facility',
-  type: 'node--health_care_local_facility',
-  published: true,
-  lastUpdated: '',
-  introText: 'Test intro text',
-  operatingStatusFacility: 'normal',
-  menu: {
-    rootPath: '/test-nav-path',
-    data: {
-      name: 'test-nav',
-      description: 'test-nav-description',
-      links: [
-        {
-          url: {
-            path: '/test-nav-path',
-          },
-          description: 'test-nav-description',
-          expanded: true,
-          label: 'test-nav-label',
-          links: [],
-        },
-      ],
-    },
-  },
-  path: '/test-nav-path',
-  vamcEhrSystem: 'vista',
-  administration: { entityId: 1234 },
+const menuItem: DrupalMenuLinkContent = {
+  title: 'Foo',
+  type: 'meh',
+  url: '/nowhere/in-particular',
+  id: 'foo',
+  description: 'bar',
+  enabled: true,
+  expanded: true,
+  menu_name: 'baz',
+  meta: {},
+  options: {},
+  parent: null,
+  provider: null,
+  route: null,
+  weight: '0',
 }
+
+const mockMenu = {
+  items: [menuItem],
+  tree: [menuItem],
+}
+
+const mockData = formatter({
+  entity: drupalMockData,
+  menu: mockMenu,
+  lovell: { isLovellVariantPage: false, variant: 'va' },
+})
 
 describe('HealthCareLocalFacility with valid data', () => {
   test('renders HealthCareLocalFacility component with basic data', () => {
--- a/src/templates/layouts/healthCareLocalFacility/index.tsx
+++ b/src/templates/layouts/healthCareLocalFacility/index.tsx
@@ -7,6 +7,8 @@ import { LocationServices } from './LocationServices'
 import { HealthServices } from './HealthServices'
 import { TopTasks } from '@/templates/components/topTasks'
 import { OperatingStatusFlags } from './OperatingStatus'
+import { numberToTimeString } from '@/lib/utils/numberToTimeString'
+import { dayOfWeek } from '@/lib/utils/dayOfWeek'
 
 // Allows additions to window object without overwriting global type
 interface customWindow extends Window {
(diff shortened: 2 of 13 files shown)


src/templates/layouts/healthCareLocalFacility/index.test.tsx
```diff
@@ -1,38 +1,41 @@
...
+}
+
+const mockData = formatter({
+  // @ts-expect-error drupalMockData technically has numbers instead of strings
+  // for some of the IDs, but this is a known problem. See
+  // https://github.com/chapter-three/next-drupal/issues/686#issuecomment-2083175598
+  entity: drupalMockData,
+  menu: mockMenu,
+  lovell: { isLovellVariantPage: false, variant: 'va' },
+})
```
cvalarida, line 37: I'm not in love with pulling in the formatter to translate the mock data from Drupal into the formatted data that the template expects, but it's not...the worst? 🤷

The alternative is either hand-crafting the formatted data like I was before, but now the data is way bigger, or run the mock data through the formatter once and capture the output in a JSON file to be imported here. 🤷 

I'm open to suggestions.

src/templates/layouts/healthCareLocalFacility/index.test.tsx
```diff
@@ -1,38 +1,41 @@
...
+const mockMenu = {
+  items: [menuItem],
+  tree: [menuItem],
+}
+
+const mockData = formatter({
+  // @ts-expect-error drupalMockData technically has numbers instead of strings
+  // for some of the IDs, but this is a known problem. See
+  // https://github.com/chapter-three/next-drupal/issues/686#issuecomment-2083175598
+  entity: drupalMockData,
```
cvalarida, line 34: This. 😮‍💨 

This hurts me. Alas, the type inside `next-drupal` is wrong, and I couldn't overwrite it without breaking other things. So here we are, closing our eyes to it. 🙈

src/templates/layouts/healthCareLocalFacility/index.tsx
```diff
@@ -22,13 +24,53 @@ export function HealthCareLocalFacility({
...
+      addressRegion: address.administrative_area,
+      postalCode: address.postal_code,
+    },
+    name: title,
+    telephone: phoneNumber,
+    openingHoursSpecification: officeHours.map((hours) => ({
+      '@type': 'OpeningHoursSpecification',
+      dayOfWeek: `https://schema.org/${dayOfWeek(hours.day)}`,
+      opens: numberToTimeString(hours.starthours),
+      closes: numberToTimeString(hours.endhours),
```
cvalarida, line 58: `Event` will probably need to do this as well so it matches the schema defined for `'@type': 'Place'`.

src/types/formatted/healthCareLocalFacility.ts
```diff
@@ -10,4 +16,10 @@ export type HealthCareLocalFacility = PublishedEntity & {
   path: string
   administration?: { entityId: number }
   vamcEhrSystem: VamcEhr['field_region_page']['field_vamc_ehr_system']
+  officeHours: FieldOfficeHours[]
+  address: FieldAddress
+  phoneNumber: string
+  image: MediaImage
+  facilityLocatorApiId: string
+  geoLocation: FieldGeoLocation
```
cvalarida, line 24: It _is_ admittedly strange to "translate" some of these properties into a formatted structure but not translate others. Buuuuuuut... 🤷 Formatting everything is a slog and doesn't actually give us any value.

src/templates/layouts/healthCareLocalFacility/index.test.tsx
brianseek: 🥹 At least its skippable and documented!

src/templates/layouts/healthCareLocalFacility/index.test.tsx
brianseek: I don't hate it. I like that changing the formatter doesn't mean running around and changing the crafted mock data.

src/templates/layouts/healthCareLocalFacility/index.tsx
```diff
@@ -22,13 +24,53 @@ export function HealthCareLocalFacility({
...
+    hasMap: `https://maps.google.com?saddr=Current+Location&daddr=${encodeURIComponent(
+      `${address.address_line1}, ${address.locality}, ${address.postal_code}`
+    )}`,
+    // Shouldn't need all these optional chains, but because we're not
+    // validating data during runtime, just in case...
+    image: image?.links?.['2_1_large']?.href,
+    branchCode: facilityLocatorApiId,
+    geo: {
+      '@type': 'GeoCoordinates',
+      latitude: geoLocation.lat,
```
brianseek, line 69: So it looks like geoLocation can be null some times, that is why the tugboat is blowing up.
see: https://www.va.gov/lovell-federal-health-care-tricare/locations/naval-great-lakes-burkey-mall-pharmacy/
the geolocation is null here.

brianseek: Screenshot 2025-04-11 at 2 30 55 PM

brianseek: Screenshot 2025-04-11 at 2 31 37 PM

brianseek (changes requested):
This all looks great to me! With one note on potentially missing data that is breaking your tugboat

brianseek (approved):
Nice work!

PR #986, opened 2025-04-22
SSG fast fail

cvalarida:
# Description

Makes SSG builds fail fast so we don't accidentally push out content builds that are missing pages because of reasons.

In particular, I wanted it to fail the build immediately so we don't wait around to build the remaining however-many-thousand pages when we _know_ we don't want to keep that build because it's missing at least one page.

## How to test

You can test it in a couple ways, but the main thing is to make sure that building a page throws an error.

1. `yarn export`
2. Throw an error during the static build

Options to throw that error include:
- Turning off your SOCKS proxy
   - This was annoying for me since mine kept turning itself back on...which is a weird thing to complain about. :sweat_smile: 
- Reverting [this commit](https://github.com/department-of-veterans-affairs/next-build/pull/986/commits/4c9f7a5f989eecd9f00eabb72560083d1ac880aa) locally before running yarn export
   - It'll start building a few pages, then throw an error. Note that the pagesBuilt variable is scoped per worker, so the built pages count will actually get up to ~20 before it throws an error, even though the pagesBuilt > 2 makes it looks like it should happen way earlier.

## How it works

I tried a bunch of ways to get this working, but this was what I settled on. There's a whole story here. Buckle up; I'll try to tell it properly.

What we need is for the build to stop _as soon_ as it cannot build even just one page. It needs to stop and exit with a non-zero exit code so CI doesn't think the command completed successfully. If it exits with a 0, the content deploy job will do the S3 sync, even if there are only like 3 pages. This means we accidentally delete a ton of pages. That's...y'know, pretty bad. 😅 

To get the `next build` command to exit with a non-zero exit code, I tried all the following:
- Throw an error
  - The build just laughs and keeps on going
  - Looking through Next.js's source, it looks like because they intend for us to use a fallback page, but we don't want this
- `process.exit(1)`
  - This actually just exits the child worker that's processing that particular page, not the parent process that's responsible for spawning all the workers that process the pages
  - It's a Next.js thing we have no control over without forking the repo
-  `process.kill(process.ppid, 'SIGINT')` to kill the parent process
  - My goal with this one was to simulate a `^C` keystroke from the CLI
  - While this technically works, it the command terminates with an exit code 0
  - I tried `SIGINT`, `SIGTERM`, and `SIGKILL`, but they all resulted in an exit code 0
- I even grabbed the grandparent PID by executing a `ps` shell command, then killed that process
  - This resulted in a ton of orphaned child (worker) processes logging to `stderr` after I killed their grandparent, which was just tragic
  - I actually couldn't get them to stop and had to close the terminal window; they were very distraught 😢 
- Listening to `cmd.stderr.on('data', (data) => { ... })` for a specific error message that I'd log
  - This actually worked really well, but I hated it because...it's pretty gross 😬 

So in the end, the way I made it work was by passing data between the process spawner in `processEnv()` and the child processes in the form of writing files to disk. Very old school. Also pretty gross, but less "magic."

## Screenshots

Behold the exit code 1!
image

## Generated description

This pull request introduces changes to improve the handling of static site generation (SSG) failures during the `next build` process. The changes include creating a mechanism to track the `next build` process ID and terminate it in case of errors to prevent partial builds.

### Enhancements to SSG error handling:

* **Tracking `next build` process ID**: Added logic in `packages/env-loader/src/index.ts` to create a `buildID` file that stores the process ID of the `next build` command. The file is created when the build starts and deleted when the process exits. (`[[1]](diffhunk://#diff-ca053d64b4b75a4548854337d7eb6bb889b1f9f0f9336e56085fca3511f2c9cdR6-R9)`, `[[2]](diffhunk://#diff-ca053d64b4b75a4548854337d7eb6bb889b1f9f0f9336e56085fca3511f2c9cdR96-R105)`)

* **Terminating the build on SSG errors**: Updated `src/pages/[[...slug]].tsx` to read the `buildID` file and terminate the `next build` process if an error occurs during static site generation. This ensures that partial builds are avoided. (`[src/pages/[[...slug]].tsxR278-R296](diffhunk://#diff-5f218029560343be88582c198bb3b04e4c18259a5e716a88250afabe4466f957R278-R296)`)

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/20817


cvalarida (commented):
I think the thing I like least about this solution is that it relies on us using `processEnv()`. I think it's not the end of the world because we already _do_ rely on it, but the more custom tooling we've got on top of `next build`, the harder it is to onboard new folks (among other things, of course).

That said, I dug around in Next.js's docs and source code, and there just isn't a way to get `next build` to exit with a non-zero exit code if any page fails to build. The alternative is forking Next.js and modifying it, so...this feels like a better solution?

Maybe we try to get this capability integrated upstream anyhow with a CLI option or feature flag, but 🤷

brianseek:
This worked testing locally pretty well:
Screenshot 2025-04-23 at 3 56 22 PM

I am curious about the tugboat error though. I don't recall ever seeing that. Maybe something to look into there.

brianseek:
also, when running what should be a non-erroring export, I see this at the end:
Screenshot 2025-04-23 at 4 09 29 PM

cvalarida:
> I am curious about the tugboat error though. I don't recall ever seeing that. Maybe something to look into there.

I'm guessing that's because this was during the tugboat upgrade. Re-running now to confirm.

brianseek:
successful local build looks good now
Screenshot 2025-04-24 at 12 01 02 PM

cvalarida:
I've added colored logs.

At the start of the process
image

Simulated error
image

^C to SIGINT the process
image

brianseek (approved):
Looks good to me. Like the chalk addition!

PR #989, opened 2025-04-22
Cvalarida/vamc image and map

cvalarida:
# Description

Adds the image and (presumably) map to the page.

Along the way, I had to fiddle with a few other things to get it to display and behave as expected, but the changes were minimal.

## Screenshots

Locally, it can't access the facility locator API, so it doesn't actually render a map.
image

We can see by the spinner that the widget is loading correctly, though, and when we examine the DOM, we see the attributes are set correctly as well.

image

## Generated description

This pull request introduces several updates to improve the functionality and structure of the `MediaImage` and `ImageAndStaticMap` components, as well as their integration into the `HealthCareLocalFacility` layout. The main changes include adding a validation check for `alt` text in the `MediaImage` component, removing unused analytics functionality from `ImageAndStaticMap`, and integrating the `ImageAndStaticMap` component into the `HealthCareLocalFacility` layout.

### MediaImage Component Enhancements:
* Added a validation in `MediaImage` to return `null` if the `alt` text is missing or empty, ensuring accessibility compliance. (`src/templates/common/mediaImage/index.tsx`, [src/templates/common/mediaImage/index.tsxR66-R69](diffhunk://#diff-bfccd708583b87c70bd90c950ec2db5c6764c99f77ff28876fd328c969c4be23R66-R69))
* Updated the `MediaImage` test suite to include a test case verifying that the component returns `null` when the `alt` text is empty. (`src/templates/common/mediaImage/index.test.tsx`, [src/templates/common/mediaImage/index.test.tsxR99-R105](diffhunk://#diff-baaccf8461bb720a65af31245bb5b30bb4b1aca535701b16a17e5e5fb4e52d50R99-R105))

### ImageAndStaticMap Component Simplifications:
* Removed the `recordEvent` analytics functionality and associated `onClick` handler from `ImageAndStaticMap`, simplifying the component's behavior. (`src/templates/components/imageAndStaticMap/index.tsx`, [[1]](diffhunk://#diff-ead5fe56a6dba7dc893c7621f779f77a25d2461a7682200a9cf30438c319078bL2) [[2]](diffhunk://#diff-ead5fe56a6dba7dc893c7621f779f77a25d2461a7682200a9cf30438c319078bL17-R16)
* Deleted the corresponding test for the removed `recordEvent` functionality. (`src/templates/components/imageAndStaticMap/index.test.tsx`, [[1]](diffhunk://#diff-7c10f45a360345ddbf25ca9464357cc019bc1eda67721869430456571fd49a3fL10-L12) [[2]](diffhunk://#diff-7c10f45a360345ddbf25ca9464357cc019bc1eda67721869430456571fd49a3fL32-L42)

### HealthCareLocalFacility Layout Integration:
* Integrated the `ImageAndStaticMap` component into the `HealthCareLocalFacility` layout to replace a placeholder, enhancing the layout with dynamic content. (`src/templates/layouts/healthCareLocalFacility/index.tsx`, [[1]](diffhunk://#diff-dfb29969d1983625591fe3e87de074afa2660839bba7565b48c78b1d25b3aeefR16) [[2]](diffhunk://#diff-dfb29969d1983625591fe3e87de074afa2660839bba7565b48c78b1d25b3aeefL151-R155)
* Adjusted layout styles to improve responsiveness and spacing in the `HealthCareLocalFacility` component. (`src/templates/layouts/healthCareLocalFacility/index.tsx`, [src/templates/layouts/healthCareLocalFacility/index.tsxL112-R113](diffhunk://#diff-dfb29969d1983625591fe3e87de074afa2660839bba7565b48c78b1d25b3aeefL112-R113))

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/20786

### DIFF
--- a/src/templates/common/mediaImage/index.tsx
+++ b/src/templates/common/mediaImage/index.tsx
@@ -63,6 +63,10 @@ export const MediaImage = (
     return null
   }
 
+  if (!props.alt) {
+    return null
+  }
+
   return (
     <Image
       id={props.id}
--- a/src/templates/components/imageAndStaticMap/index.tsx
+++ b/src/templates/components/imageAndStaticMap/index.tsx
@@ -1,5 +1,4 @@
 import { MediaImage } from '@/templates/common/mediaImage'
-import { recordEvent } from '@/lib/analytics/recordEvent'
 
 import { MediaImage as FormattedMediaImage } from '@/types/formatted/media'
 
@@ -14,18 +13,7 @@ export const ImageAndStaticMap: React.FC<ImageAndStaticMapProps> = ({
 }) => {
   return (
     <div className="usa-width-one-third inline-table-helper vads-u-order--first mobile-lg:vads-u-order--initial vads-u-margin-bottom--2 vads-u-margin-left--auto facility">
-      <div
-        onClick={() =>
-          recordEvent({ event: 'image-click', 'facility-name': facilityId })
-        }
-        style={{ cursor: 'pointer' }}
-      >
-        <MediaImage
-          {...image}
-          imageStyle="2_1_large"
-          className="facility-img"
-        />
-      </div>
+      <MediaImage {...image} imageStyle="2_1_large" className="facility-img" />
       <div data-widget-type="facility-map" data-facility={facilityId}>
         {/* TODO: Create Facility Map component for display here */}
       </div>
--- a/src/templates/layouts/healthCareLocalFacility/index.tsx
+++ b/src/templates/layouts/healthCareLocalFacility/index.tsx
@@ -13,6 +13,7 @@ import { OperatingStatusFlags } from './OperatingStatus'
 import { Address } from './Address'
 import { Phone } from './Phone'
 import { Hours } from '@/templates/components/hours'
+import { ImageAndStaticMap } from '@/templates/components/imageAndStaticMap'
 
 // Allows additions to window object without overwriting global type
 interface customWindow extends Window {
@@ -109,7 +110,7 @@ export function HealthCareLocalFacility({
             >
               Location and contact information
             </h2>
-            <div className="region-list usa-grid usa-grid-full vads-u-display--flex vads-u-flex-direction--column small-screen:vads-u-flex-direction--row facility vads-u-margin-bottom--4">
+            <div className="region-list usa-grid usa-grid-full vads-u-display--flex vads-u-flex-direction--column mobile-lg:vads-u-flex-direction--row facility vads-u-margin-bottom--2p5 vads-u-margin-bottom--4">
               <div className="usa-width-two-thirds">
                 <div>
                   <OperatingStatusFlags
@@ -148,7 +149,10 @@ export function HealthCareLocalFacility({
                   </section>
                 </div>
               </div>
-              <div>TODO: Image and static map</div>
+              <ImageAndStaticMap
+                image={image}
+                facilityId={facilityLocatorApiId}
+              />
             </div>
             <LocationServices />
             <div>TODO: List of links section</div>
--- a/src/templates/common/mediaImage/index.test.tsx
+++ b/src/templates/common/mediaImage/index.test.tsx
@@ -96,4 +96,11 @@ describe('MediaImage component returns null', () => {
     )
     expect(container.firstChild).toBeNull()
   })
+
+  test('returns null when alt text is missing or empty', () => {
+    const { container } = render(
+      <MediaImage {...mediaImage} alt="" imageStyle="2_1_large" />
+    )
+    expect(container.firstChild).toBeNull()
+  })
 })
--- a/src/templates/components/imageAndStaticMap/index.test.tsx
+++ b/src/templates/components/imageAndStaticMap/index.test.tsx
@@ -7,9 +7,6 @@ jest.mock('@/templates/common/mediaImage', () => ({
   // eslint-disable-next-line @next/next/no-img-element
   MediaImage: ({ alt }) => <img alt={alt} />,
 }))
-jest.mock('@/lib/analytics/recordEvent', () => ({
-  recordEvent: jest.fn(),
-}))
 
 describe('ImageAndStaticMap Component', () => {
   const mockImage = {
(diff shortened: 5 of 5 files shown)


src/templates/common/mediaImage/index.tsx
```diff
@@ -63,6 +63,10 @@ export const MediaImage = (
     return null
   }
 
+  if (!props.alt) {
+    return null
+  }
+
```
cvalarida, line 69: This is to mirror [what's happening on content-build](https://github.com/department-of-veterans-affairs/content-build/blob/main/src/site/components/image.drupal.liquid#L1-L1).

Alt text is a required field in Drupal, so this shouldn't happen, but a bit o' defensive programming here makes certain that we're not displaying images without alt text.

src/templates/components/imageAndStaticMap/index.tsx
```diff
@@ -14,18 +13,7 @@ export const ImageAndStaticMap: React.FC<ImageAndStaticMapProps> = ({
 }) => {
   return (
     <div className="usa-width-one-third inline-table-helper vads-u-order--first mobile-lg:vads-u-order--initial vads-u-margin-bottom--2 vads-u-margin-left--auto facility">
-      <div
-        onClick={() =>
-          recordEvent({ event: 'image-click', 'facility-name': facilityId })
-        }
-        style={{ cursor: 'pointer' }}
-      >
```
cvalarida, line 22: I'm not sure why we were recording click events to a non-interactive image. This comes from `content-build`, but I think from a misunderstanding of what that image template was actually doing. That image template was also used to render images inside `<a>` tags. The click event would pass through both the `<a>`'s `onClick` handler _and_ the ``'s `onClick` handler. This...isn't needed right here. 🤷

src/templates/layouts/healthCareLocalFacility/index.tsx
```diff
@@ -109,7 +110,7 @@ export function HealthCareLocalFacility({
             >
               Location and contact information
             </h2>
-            <div className="region-list usa-grid usa-grid-full vads-u-display--flex vads-u-flex-direction--column small-screen:vads-u-flex-direction--row facility vads-u-margin-bottom--4">
+            <div className="region-list usa-grid usa-grid-full vads-u-display--flex vads-u-flex-direction--column mobile-lg:vads-u-flex-direction--row facility vads-u-margin-bottom--2p5 vads-u-margin-bottom--4">
```
cvalarida, line 113: Just re-copied this from `content-build` because it wasn't displaying to the side like I expected.

laflannery:
@cvalarida A few notes:

- Because we are properly rendering the Directions link and using the correct font size, its slightly larger and therefore running into the Facility image at certain mobile sizes. Can you add a small amount of right padding: `vads-u-padding-right--0p5`
Screenshot 2025-04-23 at 8 32 20 AM

- The image is using the incorrect ratio and therefore is being cut off. The is a screen shot of mobile and it was easier to see the difference but even on desktop notice that on Prod you can see the grass in the bottom left but you can't on next:
Screenshot 2025-04-23 at 8 35 47 AM

- The map doesn't load at all on Tugboat, not even the spinner. The map and the analytics are important and I'm nervous to launch without seeing these and confirming everything. I don't necessarily mind approving this individual PR but do you know of any way we could get this to show? Or is there a way we could deploy to dev?

src/templates/components/imageAndStaticMap/index.tsx
brianseek: Nice! I like that we are clearing out a lot of this tech debt stuff

brianseek (approved):
This seems good to me. I spent way too much time trying to get my local vets-api running again so that I could test the fetch locally, not sure what is wrong with it. I think if that is a sticking point for Laura, we could test it that way. 

We could also make a change to rev proxy to allow tugboat to access vets-api. [This ticket](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/20912). We could do that before switching these types over to next.

*edit: one thing to fix.

src/templates/components/imageAndStaticMap/index.tsx
```diff
@@ -14,18 +13,7 @@ export const ImageAndStaticMap: React.FC<ImageAndStaticMapProps> = ({
...
-        <MediaImage
-          {...image}
-          imageStyle="2_1_large"
-          className="facility-img"
-        />
-      </div>
+      <MediaImage {...image} imageStyle="2_1_large" className="facility-img" />
```
brianseek, line 16: Approved too quick...Laura called this out, but minor thing that you can just merge after fixing, this should be `3_2_medium_thumbnail`

cvalarida:
> This seems good to me. I spent way too much time trying to get my local vets-api running again so that I could test the fetch locally, not sure what is wrong with it. I think if that is a sticking point for Laura, we could test it that way.

I think fundamentally, the problem is that our dev server isn't set up to forward the traffic through to `vets-api`. If you look in the network tab, the request it makes is to http://localhost:3000/facilities_api/v2/va/vha_523BZ/, which will go through the next-build dev server, and since there's no `src/pages/facilities_api`, it'll go through `[[..slug]].tsx`, which will in turn make the request of Drupal.

src/templates/components/imageAndStaticMap/index.tsx
```diff
@@ -14,18 +13,11 @@ export const ImageAndStaticMap: React.FC<ImageAndStaticMapProps> = ({
...
-      >
-        <MediaImage
-          {...image}
-          imageStyle="2_1_large"
-          className="facility-img"
-        />
-      </div>
+      <MediaImage
+        {...image}
+        imageStyle="3_2_medium_thumbnail"
```
cvalarida, line 18: Changing this from `2_1_large` to `3_2_thumbnail` will also affect vet centers and vet center outstations. @laflannery can you double-check that those are still okay? (I'm not sure how to get to those pages right now. 🙈 )

If we need, we can always prop thread this to change it up per content type that needs to display this image and map combo.

laflannery: @cvalarida All facility images should have the same ratio, we have very similar layout for these facility. I confirmed Vet center facility specifically though, [here's an example](https://www.va.gov/boston-vet-center/)) but there are no Outstations because these were never finished. However, Outstations were intended to essentially be a copy of the main Vet center pages so those images should be the same as well.

laflannery (approved):
Looks good! I would still like to see if we can figure out a way that isn't a massive lift for us to test the map before launch but we can definitely merge this for now

PR #1002, opened 2025-04-30
Unrevert press release fix

cvalarida:
# Description

Re-fixes the Lovell thing for press release listing pages.

The main change here is that we're also spreading the `webpackConfig.optimization`, so when we go to make a prod release, it's still minified _and_ it doesn't choke up by not finding the minimizer plugin.


next.config.js
```diff
@@ -29,6 +29,15 @@ const nextConfig = {
     // this could be anything, use latest git hash if it exists
     return process.env.GIT_HASH ?? 'vagovprod'
   },
+  webpack(webpackConfig) {
+    return {
+      ...webpackConfig,
+      optimization: {
+        ...webpackConfig.optimization,
```
brianseek, line 36: 🙃

timcosgrove (approved):

PR #1005, opened 2025-05-01
VAMC facility related links

cvalarida:
# Description

Adds the related links section. It was probably more painful than it needed to be because of the stuff that was in there already but unused. Leveraging the existing component and incorrect static types...I had to fix some stuff along the way.

## Screenshots

image

**Note:** The DOM is different. @laflannery The `<ul>` -> `<li>` -> `<p>` -> `<va-link>` was what was already in this component in Next Build. Should I continue to use that? I was thinking yes because of the increased font size thing. Also using an unordered list seemed better than a div, but please sanity check me on that. 😅 

Increased font size:
image

## Generated description

This pull request introduces updates to the `HealthCareLocalFacility` query, enhancing its functionality to include related links, updating snapshots for testing, and refining data formatting. Key changes include adding support for related links in the query, updating snapshot tests to reflect the new data structure, and modifying URLs and metadata for consistency.

### Query Enhancements:
* Added `field_region_page.field_related_links.field_va_paragraphs` to the query's `addInclude` parameters to fetch related links.
* Introduced a `relatedLinks` field in the query formatter to include a section title and a list of links derived from `ParagraphLinkTeaser` entities.

### Snapshot Updates:
* Updated snapshot tests in `healthCareLocalFacility.test.tsx.snap` to reflect changes in the query output, including new `relatedLinks` data and updated metadata. [[1]](diffhunk://#diff-209da7eaba46ff4f3c51597305f96e3c943c67839359c0e0533a1e1b099d5376L387-R458) [[2]](diffhunk://#diff-209da7eaba46ff4f3c51597305f96e3c943c67839359c0e0533a1e1b099d5376L406-R474)
* Modified URLs in snapshots to replace development URLs with staging URLs for consistency. [[1]](diffhunk://#diff-209da7eaba46ff4f3c51597305f96e3c943c67839359c0e0533a1e1b099d5376L22-R27) [[2]](diffhunk://#diff-209da7eaba46ff4f3c51597305f96e3c943c67839359c0e0533a1e1b099d5376L417-R490)

### Metadata Adjustments:
* Adjusted metadata fields such as `langcode`, `content_translation_changed`, and `field_phone_label` to ensure accurate representation of the data. [[1]](diffhunk://#diff-209da7eaba46ff4f3c51597305f96e3c943c67839359c0e0533a1e1b099d5376L51-R86) [[2]](diffhunk://#diff-209da7eaba46ff4f3c51597305f96e3c943c67839359c0e0533a1e1b099d5376L446-R549)
* Updated image metadata, including `twitter:image` and `twitter:image:alt`, for better social media integration. [[1]](diffhunk://#diff-209da7eaba46ff4f3c51597305f96e3c943c67839359c0e0533a1e1b099d5376L322-R361) [[2]](diffhunk://#diff-209da7eaba46ff4f3c51597305f96e3c943c67839359c0e0533a1e1b099d5376L728-R835)

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/20792

### DIFF
--- a/src/data/queries/healthCareLocalFacility.ts
+++ b/src/data/queries/healthCareLocalFacility.ts
@@ -100,9 +100,9 @@ export const formatter: QueryFormatter<
     relatedLinks: {
       sectionTitle: entity.field_region_page.title
         ? `Other services at ${entity.field_region_page.title}`
-        : entity.field_region_page.field_related_links.field_va_paragraphs
+        : entity.field_region_page.field_related_links?.field_va_paragraphs
             .field_title,
-      links: entity.field_region_page.field_related_links.field_va_paragraphs
+      links: entity.field_region_page.field_related_links?.field_va_paragraphs
         .slice(0, 8)
         // Adding the type annotation because TS doesn't apparently pick up on
         // this since we've done an Omit<> on the parent type.
--- a/src/data/queries/healthCareLocalFacility.ts
+++ b/src/data/queries/healthCareLocalFacility.ts
@@ -13,6 +13,7 @@ import { Menu } from '@/types/drupal/menu'
 import { buildSideNavDataFromMenu } from '@/lib/drupal/facilitySideNav'
 import { getLovellVariantOfMenu } from '@/lib/drupal/lovell/utils'
 import { formatter as formatImage } from '@/data/queries/mediaImage'
+import { ParagraphLinkTeaser } from '@/types/drupal/paragraph'
 
 // Define the query params for fetching node--health_care_local_facility.
 export const params: QueryParams<null> = () => {
@@ -103,7 +104,9 @@ export const formatter: QueryFormatter<
             .field_title,
       links: entity.field_region_page.field_related_links.field_va_paragraphs
         .slice(0, 8)
-        .map((linkTeaser) => ({
+        // Adding the type annotation because TS doesn't apparently pick up on
+        // this since we've done an Omit<> on the parent type.
+        .map((linkTeaser: ParagraphLinkTeaser) => ({
           title: linkTeaser.field_link.title,
           uri: linkTeaser.field_link.url,
           // summary: ''
--- a/src/data/queries/healthCareLocalFacility.ts
+++ b/src/data/queries/healthCareLocalFacility.ts
@@ -37,7 +37,7 @@ export type HealthCareLocalFacilityDataOpts = {
  * We're adding `lovell` from the context here to conditionally re-shape
  * the menu for Lovell facilities.
  */
-type LocalFacilityData = {
+export type LocalFacilityData = {
   entity: NodeHealthCareLocalFacility
   menu: Menu | null
   lovell?: ExpandedStaticPropsContext['lovell']
--- a/src/data/queries/healthCareLocalFacility.ts
+++ b/src/data/queries/healthCareLocalFacility.ts
@@ -17,7 +17,7 @@ import { formatter as formatImage } from '@/data/queries/mediaImage'
 // Define the query params for fetching node--health_care_local_facility.
 export const params: QueryParams<null> = () => {
   return new DrupalJsonApiParams().addInclude([
-    'field_region_page',
+    'field_region_page.field_related_links.field_va_paragraphs',
     'field_media',
     'field_media.image',
     'field_administration',
@@ -96,5 +96,19 @@ export const formatter: QueryFormatter<
     image: formatImage(entity.field_media),
     facilityLocatorApiId: entity.field_facility_locator_api_id,
     geoLocation: entity.field_geolocation,
+    relatedLinks: {
+      sectionTitle: entity.field_region_page.title
+        ? `Other services at ${entity.field_region_page.field_related_links.field_title}`
+        : entity.field_region_page.field_related_links.field_va_paragraphs
+            .field_title,
+      links:
+        entity.field_region_page.field_related_links.field_va_paragraphs.map(
+          (linkTeaser) => ({
+            title: linkTeaser.field_link.title,
+            uri: linkTeaser.field_link.url,
+            // summary: ''
+          })
+        ),
+    },
   }
 }
--- a/src/data/queries/healthCareLocalFacility.ts
+++ b/src/data/queries/healthCareLocalFacility.ts
@@ -100,17 +100,18 @@ export const formatter: QueryFormatter<
     relatedLinks: {
       sectionTitle: entity.field_region_page.title
         ? `Other services at ${entity.field_region_page.title}`
-        : entity.field_region_page.field_related_links?.field_va_paragraphs
-            .field_title,
-      links: entity.field_region_page.field_related_links?.field_va_paragraphs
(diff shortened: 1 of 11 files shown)


src/data/queries/tests/__snapshots__/healthCareLocalFacility.test.tsx.snap
cvalarida, line 1: Added the related links to the mock data, which resulted in updated snapshots.

src/mocks/healthCareLocalFacility.mock.json
cvalarida, line 1: Added related links, which gets _very_ verbose in this diff because of all the "filler" fields. Also, I'm pulling this from  a tugboat instance now, so it changed a lot of the other fields as well, which is kinda noisy.

src/data/queries/tests/storyListing.test.tsx
```diff
@@ -141,9 +141,8 @@ describe('data function for StoryListing', () => {
           field_media: null,
           field_related_links: {
             id: '1',
-            field_link: null,
-            field_links: [],
-            field_section_header: null,
+            field_va_paragraphs: [],
+            field_title: null,
```
cvalarida, line 145: The `field_office` uses the `node--health_care_region_page`, which maps to the `NodeHealthCareRegionPage` static type. This static type was using the wrong static type for its `field_related_links`, so this mock data was the wrong shape.

This is what we get for not pulling mock data from an authoritative source. 🤷

src/data/queries/healthCareLocalFacility.ts
cvalarida, line 1: This feels a little messy, but I think we either handle the messiness here, in the template, or in the `<RelatedLinks>` component. This felt like the right place, but I'm definitely open to feedback.

It's entirely possible that it's messy here because I was trying to shoehorn this into that existing—but otherwise unused—`<RelatedLinks>` component. I wouldn't be against just tossing in the `field_related_links` object and the `field_region_page.title` to that component and letting it handle this logic we've got in here now. 🤷

src/templates/common/relatedLinks/index.tsx
cvalarida, line 1: This whole component was already built, but entirely unused, so I decided to go ahead and style it for this particular use case. We may regret that later or not, I have no idea. 🤷

src/types/drupal/node.ts
```diff
@@ -200,7 +201,7 @@ export interface NodeHealthCareLocalHealthService extends DrupalNode {
 export interface NodeHealthCareRegionPage extends DrupalNode {
   field_appointments_online: boolean
   field_media: DrupalMediaImage
-  field_related_links: ParagraphListOfLinks
+  field_related_links: ParagraphListOfLinkTeasers
```
cvalarida, line 204: This was wrong. 🤷

src/types/formatted/healthCareLocalFacility.ts
```diff
@@ -25,4 +26,5 @@ export type HealthCareLocalFacility = PublishedEntity & {
   geoLocation: FieldGeoLocation
   fieldTelephone: ParagraphPhoneNumber | null
   vaHealthConnectPhoneNumber: string | null
+  relatedLinks: FormattedRelatedLinks
```
cvalarida, line 29: I don't know that I love this, but 🤷

cvalarida (commented):
There's a whole lotta "meh" and "🤷." 🙈

laflannery:
@cvalarida Tugboat is not building, there are errors on the Facility pages. I did try to rebuild just to check, but unfortunately I got the same errors

laflannery:
Yes @cvalarida using the semantic elements is definitely better, thank you! My only comment is that we can remove the margins on the `<li>` and `<p>` tags because they are essentially cancelling each other out:

- We are saying add margin to the `<li>` and then remove it from the `<p>`
- However, if we left everything default, it's essentially the same but reversed (the margin on the `<p>` is slightly larger but that's ok)

src/templates/common/relatedLinks/index.tsx
```diff
@@ -11,41 +9,40 @@ export const RelatedLinks = ({
...
 
-  let link
-  const renderLink = (uri, title, summary) => (
+  const renderLink = (uri: string, title: string, summary?: string) => (
     <>
       <p className="vads-u-margin--0">
-        <strong>
-          <va-link href={uri} text={title} />
-        </strong>
+        <va-link disable-analytics href={uri} text={title} />
```
brianseek, line 15: why the disable-analytics?

laflannery: I was wondering the same thing and I just chased that down. We are currently disabling DST analytics on these links in Prod so that's correct. But what I'm not sure about is is we are correctly capturing the custom event. [This is the ticket from sitewide](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/18501) about this work, I'm not sure I can see the custom event in that PR though, is this something we have @cvalarida ?

cvalarida: Genuinely have no idea. It's like that in prod. 🤷

cvalarida:
> Yes @cvalarida using the semantic elements is definitely better, thank you! My only comment is that we can remove the margins on the `<li>` and `<p>` tags because they are essentially cancelling each other out:
> 
> * We are saying add margin to the `<li>` and then remove it from the `<p>`
> * However, if we left everything default, it's essentially the same but reversed (the margin on the `<p>` is slightly larger but that's ok)

The only trouble with that is the final spacing at the very end of the list. The way we've got it now, we can check to see if the `<li>` is the last in the list and not add that final bottom margin. The other way, we cannot.

There are workarounds, but they involve adding custom CSS, which I'd prefer to not if we can manage it.

| Before | After |
|---|---|
| image | image |

laflannery:
Ah, got it no worries, I'm already over it lol

laflannery (approved):

src/data/queries/healthCareLocalFacility.ts
```diff
@@ -96,5 +97,21 @@ export const formatter: QueryFormatter<
...
+        entity.field_region_page.field_related_links?.field_va_paragraphs
+          .slice(0, 8)
+          // Adding the type annotation because TS doesn't apparently pick up on
+          // this since we've done an Omit<> on the parent type.
+          .map((linkTeaser: ParagraphLinkTeaser) => ({
+            title: linkTeaser.field_link.title,
+            uri: linkTeaser.field_link.url,
+            // summary: ''
+          })) ?? null,
+    },
```
brianseek, line 115: I feel like the sectionTitle and links business should have a test

brianseek (changes requested):
This looks great! I think we are missing some test coverage in the query processing though

cvalarida:
> This looks great! I think we are missing some test coverage in the query processing though

Great call! I added tests and discovered I got a thing wrong. 😅 Fixed it up and it should be golden now. 🎉

brianseek (approved):
nice! :shipit:

PR #1010, opened 2025-05-02
VAMC facility: Add the satisfaction scores widget

cvalarida:
# Description

Adds the widget to fetch the information from `vets-api` and display the patient satisfaction scores.

You can find the [original here in `content-build`](https://github.com/department-of-veterans-affairs/content-build/blob/main/src/site/layouts/health_care_local_facility.drupal.liquid#L214-L218). [That filter comes from here](https://github.com/department-of-veterans-affairs/content-build/blob/main/src/site/filters/liquid.js#L426-L431).

Alas, I don't know how to test it since I can't connect to `vets-api`. We may need to make a passthrough route (or routes) to fetch data from `vets-api` whenever we need to do that. 🤔  It'd only be a dev thing, I think...

## Generated description

This pull request updates the `HealthCareLocalFacility` component to dynamically render a patient satisfaction scores section based on the facility's API ID.

### Updates to `HealthCareLocalFacility`:

* Added a conditional rendering block to include a `div` for patient satisfaction scores if the `facilityLocatorApiId` contains `'vha_'`. This `div` uses `data-widget-type` and `data-facility` attributes to integrate the scores widget. (`src/templates/layouts/healthCareLocalFacility/index.tsx`, [src/templates/layouts/healthCareLocalFacility/index.tsxL166-R171](diffhunk://#diff-dfb29969d1983625591fe3e87de074afa2660839bba7565b48c78b1d25b3aeefL166-R171))

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/20796


laflannery:
I am not sure the api loads in Tugboat which is problematic for testing lol. I do get the loading indicator so that something:

Screenshot 2025-05-05 at 9 36 56 AM

One thing - can you confirm you put the ID on the heading so it shows in the "On this page"? Then I had already spoken to Tim about the map and there is a way we can deploy a branch to dev so we can merge and I can test there.

cvalarida:
> One thing - can you confirm you put the ID on the heading so it shows in the "On this page"?

Alas, I cannot. Or rather, not in this PR at least. It'd be a change to the widget in `vets-website` that gets loaded up automagically. It makes the API call and renders the output based on the result. That also means that prod doesn't have the ID and it isn't showing up in those ToCs.

If you like, we can definitely make that happen, but in a PR to `vets-website`, not `next-build`.

laflannery:
@cvalarida This heading definitely does show currently in Prod: https://www.va.gov/boston-health-care/locations/jamaica-plain-va-medical-center/#our-patient-satisfaction-scores

Screenshot 2025-05-05 at 1 35 18 PM

However, if this is not manually added this is probably a case where the change in dynamic ID functionality may not work here and we need a different solution?

moanu404 (approved):
🚀

cvalarida:
> @cvalarida This heading definitely does show currently in Prod: https://www.va.gov/boston-health-care/locations/jamaica-plain-va-medical-center/#our-patient-satisfaction-scores
> 
> Screenshot 2025-05-05 at 1 35 18 PM
> 
> However, if this is not manually added this is probably a case where the change in dynamic ID functionality may not work here and we need a different solution?

:awkward-turtle:

This is what I see for [Brockton](https://www.va.gov/boston-health-care/locations/brockton-va-medical-center/):
image

cvalarida:
Probably we've got a race condition. If the script to load the satisfaction scores finishes first (only time I can imagine it'd do that is if the API response was cached locally? 🤔), then it might render before the script to populate that ToC.

cvalarida:
image

I don't see it in the list for the ToC; can't get that satisfaction scores widget to render after the ToC. 😕 I'm really interested how you managed that now. 😅

laflannery:
Merging this and working on that later works for me. Also side note, I talked to Tim about how to test and we will end up deploying to dev so that I can test this and the map and anything else we can't see on Tugboat once we are all done with everything. He thinks he has some docs written up somewhere

PR #1011, opened 2025-05-03
VAMC facility social links

cvalarida:
# Description

Adds the social links to the facility page.

I discovered a shared `SocialLinks` component, but when I tried to use it, the thing looked like it was doing way more than I needed it to do, and it wasn't styled correctly (indeed, the DOM wasn't using a `<section>` and had the `<h2>` in the `<ul>`... 😵), so I decided to start from scratch to do exactly what I needed.

## Generated description

This pull request introduces a new `FacilitySocialLinks` component to display social media and subscription links for healthcare facilities. It includes updates to the query, types, templates, and tests to integrate this feature.

### New Feature: Facility Social Links Component

* **New `FacilitySocialLinks` Component**: Added a reusable component to display social media links (e.g., Facebook, Twitter) and subscription options (e.g., emergency notifications, news updates) for healthcare facilities. The component dynamically renders links based on provided props. (`src/templates/layouts/healthCareLocalFacility/FacilitySocialLinks.tsx`)
* **Integration into Facility Page**: Updated the `HealthCareLocalFacility` layout to include the `FacilitySocialLinks` component, replacing the placeholder "Social links section" with dynamic rendering. (`src/templates/layouts/healthCareLocalFacility/index.tsx`) [[1]](diffhunk://#diff-dfb29969d1983625591fe3e87de074afa2660839bba7565b48c78b1d25b3aeefR18) [[2]](diffhunk://#diff-dfb29969d1983625591fe3e87de074afa2660839bba7565b48c78b1d25b3aeefR44) [[3]](diffhunk://#diff-dfb29969d1983625591fe3e87de074afa2660839bba7565b48c78b1d25b3aeefL167-R169)

### Query and Data Model Updates

* **Query Enhancements**: Modified the `QueryFormatter` in `healthCareLocalFacility.ts` to include `socialLinks` data, such as social media URLs and subscription IDs. (`src/data/queries/healthCareLocalFacility.ts`)
* **Type Updates**: Extended the `HealthCareLocalFacility` type to include a new `socialLinks` field, which uses the `FacilitySocialLinksProps` interface. (`src/types/formatted/healthCareLocalFacility.ts`) [[1]](diffhunk://#diff-f1c2e1972ee32d794f599ac7233134a307e92da47943ff8019fbe3442c585589L13-R13) [[2]](diffhunk://#diff-f1c2e1972ee32d794f599ac7233134a307e92da47943ff8019fbe3442c585589R38)

### Testing Improvements

* **Snapshot Updates**: Updated snapshots in `healthCareLocalFacility.test.tsx.snap` to include `socialLinks` data, ensuring proper formatting and rendering. (`src/data/queries/tests/__snapshots__/healthCareLocalFacility.test.tsx.snap`) [[1]](diffhunk://#diff-209da7eaba46ff4f3c51597305f96e3c943c67839359c0e0533a1e1b099d5376R487-R529) [[2]](diffhunk://#diff-209da7eaba46ff4f3c51597305f96e3c943c67839359c0e0533a1e1b099d5376R1032-R1074)
* **Component Tests**: Added unit tests for `FacilitySocialLinks` to validate rendering of various link types, including social media, news, and emergency notifications. (`src/templates/layouts/healthCareLocalFacility/FacilitySocialLinks.test.tsx`)

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/20797


src/templates/common/socialLinks/index.tsx
cvalarida, line 1: I started using this component, but it proved too much of a hassle, so I ended up making my own. Decided to leave these comments here in case they were helpful, since I was trying to puzzle through this myself. 🤷

src/templates/layouts/healthCareLocalFacility/FacilitySocialLinks.test.tsx
cvalarida, line 1: ChatGPT-generated tests. I think they're pretty comprehensive, but I didn't spend a ton of time on them. If I'm missing something, please let me know!

src/templates/layouts/healthCareLocalFacility/FacilitySocialLinks.tsx
```diff
@@ -0,0 +1,187 @@
...
+  fieldFacebook,
+  fieldTwitter,
+  fieldFlickr,
+  fieldInstagram,
+  fieldYoutube,
+}: FacilitySocialLinksProps) => {
+  const showGovDelivery =
+    !!fieldGovdeliveryIdEmerg ||
+    !!fieldGovdeliveryIdNews ||
+    !!fieldOperatingStatus
```
cvalarida, line 38: This is different from the [original template](https://github.com/department-of-veterans-affairs/content-build/blob/main/src/site/facilities/facility_social_links.drupal.liquid#L5-L5), which will only render the first section—the GovDelivery section—of these links if it has the `Emerge` or `News` fields, but _not_ if those are empty _and_ the operating status is not.

I'm _assuming_ this is an oversight, but @laflannery please keep me honest. Should we show the link to the operating status if the `news and announcements` or `emergency notifications` links aren't present?

src/templates/layouts/healthCareLocalFacility/FacilitySocialLinks.tsx
```diff
@@ -0,0 +1,187 @@
...
+  const showSocialLinks =
+    !!fieldFacebook ||
+    !!fieldTwitter ||
+    !!fieldFlickr ||
+    !!fieldInstagram ||
+    !!fieldYoutube
+
+  if (!showGovDelivery && !showSocialLinks) {
+    return null
+  }
```
cvalarida, line 49: This is also new. I'm assuming we shouldn't show a "Get updates from..." section with no links in it. 😅

src/templates/layouts/healthCareLocalFacility/FacilitySocialLinks.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+  return (
```
cvalarida, line 51: ChatGPT-translated from the Liquid template. Only thing I had to change was the operating status URL bit because it's found in `fieldOperatingStatus.url` now, not `fieldOperatingStatus.url.path` like it is in `content-build`.

src/types/drupal/field_type.d.ts
```diff
@@ -31,8 +31,7 @@ export interface FieldLink {
   uri: string //e.g. `entity:node/2424`
   url?: string //e.g. `/outreach-and-events/events/2424`
   title: string
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  options: any[]
+  options: unknown[]
```
cvalarida, line 34: Not actually related to my changes, but I found this while spelunking and decided it should be `unknown` instead of `any`. 🤷

src/types/formatted/healthCareLocalFacility.ts
```diff
@@ -10,7 +10,7 @@ import { PublishedEntity } from './publishedEntity'
 import { FormattedRelatedLinks } from './relatedLinks'
 import { SideNavMenu } from './sideNav'
 import { MediaImage } from '@/types/formatted/media'
-import { Wysiwyg } from './wysiwyg'
+import { FacilitySocialLinksProps } from '@/templates/layouts/healthCareLocalFacility/FacilitySocialLinks'
```
cvalarida, line 13: I don't love pulling this type in from the template, but it _does_ make it explicit that the data is _meant_ to be fed into the component. 🤷 If you've got strong opinions, let me know and I can refactor.

src/templates/layouts/healthCareLocalFacility/FacilitySocialLinks.tsx
laflannery: @cvalarida I honestly wasn't sure so I went to check the current logic. Both GovDelivery fields are required in Drupal so they shouldn't ever be empty:
Screenshot 2025-05-05 at 9 22 57 AM

So that being said, I don't really think we need lots of fancy logic around a condition that really shouldn't ever happen

laflannery:
@cvalarida Just one minor update

- "Get updates..." needs an ID so it will show in the On the page component
(I also 100% forgot about this for the "Other Services" so I will make a ticket for this)

src/templates/layouts/healthCareLocalFacility/FacilitySocialLinks.tsx
cvalarida: It was already there, so I'ma just leave it. I'd much rather be paranoid that the requirement may be loosened or bypassed later than have stuff break—either loudly or quietly.

laflannery (approved):

brianseek (approved):
Looks good to me! One thing that I would mention is that this is not using the new grid stuff. It feels like for newly added files, we shouldn't be adding old grid in since it will then get removed. But I also understand that there is maybe a more broad strategy for converting stuff from a page type POV...so 🤷

PR #1016, opened 2025-05-06
Lovell news releases left nav fix

moanu404:
# Description

Fixes the tricare pages to now have tricare links - it uses the same format as other sidebars now use, it checks based dynamically on the props to see if lovell or tricare is detected, and handles the menu accordingly.

## Generated description

This pull request refactors the `PressReleaseListing` query to improve handling of Lovell facility variants (TRICARE and VA) and enhances test coverage for the formatter logic. Key changes include introducing utilities for variant detection, updating the `buildSideNavDataFromMenu` function to support variants, and adding comprehensive tests for these updates.

### Refactoring and Feature Enhancements:
* Updated `src/data/queries/pressReleaseListing.ts` to include utilities for detecting Lovell TRICARE and VA paths (`isLovellTricarePath`, `isLovellVaPath`) and refactored the `formatter` to handle menu variants dynamically based on the detected path. (`[[1]](diffhunk://#diff-32f669bc8c8d7816ed403b5c2823741c5ba666e5990ceaec57185aa02239a7eaR11-R12)`, `[[2]](diffhunk://#diff-32f669bc8c8d7816ed403b5c2823741c5ba666e5990ceaec57185aa02239a7eaL105-R118)`)

### Test Coverage Improvements:
* Added imports and mock implementations in `src/data/queries/tests/pressReleaseListing.test.tsx` to support testing the new formatter logic. (`[[1]](diffhunk://#diff-bcc35aa6d85a14dd9c2c5d6ce6176537f556c9f235f12dda32f8ced7012ceec8R9)`, `[[2]](diffhunk://#diff-bcc35aa6d85a14dd9c2c5d6ce6176537f556c9f235f12dda32f8ced7012ceec8R43-R59)`)
* Introduced new tests to verify TRICARE and VA variant detection, as well as fallback behavior for non-Lovell paths, ensuring the formatter correctly handles different scenarios. (`[src/data/queries/tests/pressReleaseListing.test.tsxR99-R274](diffhunk://#diff-bcc35aa6d85a14dd9c2c5d6ce6176537f556c9f235f12dda32f8ced7012ceec8R99-R274)`)
## Ticket

Closes _link to [21247](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21247)

## Developer Task

```md
- [ ] PR submitted against the `main` branch of `next-build`.
- [ ] Link to the issue that this PR addresses (if applicable).
- [ ] Define all changes in your PR and note any changes that could potentially be breaking changes.
[...]

PR #1017, opened 2025-05-06
Add the beginnings of a VAMC System page

pwolfert:
# Description

This PR creates the basic scaffolding for the `VAMC System` page (`node--health_care_region_page`) as well as one of its first features: The side nav.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21258

## Testing Steps

Routes to test (production and localhost):

https://www.va.gov/washington-dc-health-care/
https://www.va.gov/boston-health-care
https://www.va.gov/loma-linda-health-care

http://localhost:3999/washington-dc-health-care/
http://localhost:3999/boston-health-care
http://localhost:3999/loma-linda-health-care

## QA steps

For now we just need to make sure the title and nav bar show up and that the header and footer are correct.


playwright/tests/vamcSystem.spec.js
```diff
@@ -0,0 +1,23 @@
+import { test, expect } from '../utils/next-test'
+
+// This test should be modified to test page output and function.
+// test.describe('Vamc_system', () => {
+test.skip('Vamc_system', () => {
```
pwolfert, line 5: Should I bother with the playwright tests yet?

src/templates/layouts/healthCareLocalFacility/OperatingStatus.tsx
```diff
@@ -1,5 +1,9 @@
 import { FacilityOperatingStatusFlags } from '@/types/drupal/node'
-import { HealthCareLocalFacility as FormattedHealthCareLocalFacility } from '@/types/formatted/healthCareLocalFacility'
+
+interface OperatingStatusFlagsProps {
+  operatingStatusFacility: FacilityOperatingStatusFlags
+  basePath: string
+}
```
pwolfert, line 6: I changed this component because I'm going to start using it in my template. That probably means it should move out of this folder, but I didn't think we were ready for that. I actually had more code related to the usage of this, but I accidentally clobbered it beyond recovery, like a noob.

src/templates/layouts/healthCareLocalFacility/OperatingStatus.tsx
```diff
@@ -10,14 +14,9 @@ import { HealthCareLocalFacility as FormattedHealthCareLocalFacility } from '@/t
...
-}: Pick<
-  FormattedHealthCareLocalFacility,
-  'operatingStatusFacility' | 'menu'
->) => {
-  const linkBasePath = menu.data.links[0].url.path
-
-  if (!linkBasePath || operatingStatusFacility === 'normal') {
+  basePath,
+}: OperatingStatusFlagsProps) => {
+  if (!basePath || operatingStatusFacility === 'normal') {
```
pwolfert, line 19: I changed this component because I'm going to start using it in my template. That probably means it should move out of this folder, but I didn't think we were ready for that. I actually had more code related to the usage of this, but I accidentally clobbered it beyond recovery, like a noob.

src/templates/layouts/vamcSystem/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+    window.sideNav = menu
+  })
+  
+  // TODO: It looks like the reason the image is taller than in the original page is
+  // because the image file is actually different. It has a taller intrinsic height.
+  // Here is an example from `/washington-dc-health-care/`:
+  // Old:
+  // https://s3-us-gov-west-1.amazonaws.com/content.www.va.gov/img/styles/7_2_medium_thumbnail/public/2021-08/Washington%20VA%20Medical%20Center.jpg
+  // New:
+  // https://dsva-vagov-staging-cms-files.s3.us-gov-west-1.amazonaws.com/styles/2_1_large/public/2021-08/Washington%20VA%20Medical%20Center.jpg
```
pwolfert, line 50: I'm not ready for criticisms to be lobbied against this image. I'll fix it in another pull request

pwolfert: And if anyone has ideas about my comments here, I'd welcome any insights

playwright/tests/vamcSystem.spec.js
cvalarida: 🤷 Probably would be good to eventually, but I haven't done this yet either. You definitely don't need to worry about it for now.

src/pages/api/drupal-query.ts
```diff
@@ -60,7 +60,27 @@ export default async function handler(
...
+      if (
+        item.type === 'node--health_care_region_page' &&
+        'field_clinical_health_services' in item
+      ) {
+        return {
+          ...item,
+          field_clinical_health_services:
+            item.field_clinical_health_services.map((service) => ({
+              ...service,
+              field_region_page: '[Circular Reference]',
```
cvalarida, line 76: Interesting. If you're using this to get your mock data, does this not give you a type error when you import and use the mock data?

src/pages/api/drupal-query.ts
```diff
@@ -60,7 +60,27 @@ export default async function handler(
       withAuth,
     })
 
-    return res.status(200).json(data)
+    // Avoid circular references in the health_care_region_page node. Every item in its
+    // `field_clinical_health_services` has a 'field_region_page', and we need to set it
+    // to "["Circular Reference]". Circular references will cause the json output to fail
+    const safeData = data.map((item) => {
```
cvalarida, line 66: To reiterate what we talked about while pairing the other day, this `drupal-query` API route is only used for the API explorer, which is only used for—uh, well, exploration. 😅 This isn't production facing, so throwing a hack in here for reference is kind of ugly, but not the end of the world.

Mind calling all that out in this comment block for posterity?

src/templates/layouts/vamcSystem/index.tsx
cvalarida: Ah, yeah. There's Some Stuff™ that happens with images to generate different styles. You've just gotta use the correct one. I forget what the details off the top of my head, but you can find an example here (similar to what you've got below): https://github.com/department-of-veterans-affairs/next-build/blob/main/src/templates/components/imageAndStaticMap/index.tsx#L16-L20

src/templates/layouts/vamcSystem/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+        <div className="usa-width-three-fourths">
+          <article className="usa-content va-l-facility-detail vads-u-padding-bottom--0">
+            {/* <LovellSwitcher
+              currentVariant={lovellVariant}
+              switchPath={lovellSwitchPath}
+            /> */}
+            {title && <h1>{title}</h1>}
+            {hasValidImage && (
+              <div className="duotone darken lighten medium-screen:vads-u-margin-bottom--0p5">
+                <MediaImage {...image} imageStyle="2_1_large" />
```
cvalarida, line 68: Does it need to be `7_2_medium_thumbnail`?
```suggestion
                <MediaImage {...image} imageStyle="7_2_medium_thumbnail" />
```

src/templates/layouts/vamcSystem/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+                className="vads-u-font-size--md vads-u-display--block vads-u-width--full"
+                href={`${path}/locations`}
+                text="See all locations"
+              />
+            </section>
+            {/* Manage your health online section */}
+            {administration?.id !== 1039 && (
+              <section>
+                <h2>
+                  {administration?.id === 1040
```
cvalarida, line 101: Oof. Magic numbers. Can we define these as `consts` with meaningful names?

@laflannery or @omahane Do you know offhand what these magic numbers are?

src/mocks/vamcSystem.mock.json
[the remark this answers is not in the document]
pwolfert: It's a lot. I did make it a little smaller. I could maybe go through and delete all the fields that we'll never need. Like the tags?

src/pages/api/drupal-query.ts
pwolfert: That's a good question. I didn't get past the type error of having numeric ids vs string ids, so I don't know.

pwolfert: I guess I could set them to null, but I figured these would help act like comments

src/pages/api/drupal-query.ts
pwolfert: https://github.com/department-of-veterans-affairs/next-build/pull/1017/commits/f95ebd42c4cf1280a08cb702744fa1e450769872

src/templates/layouts/vamcSystem/index.tsx
pwolfert: That was it, thank you

src/templates/layouts/vamcSystem/index.tsx
pwolfert: Meaningful names would be nice. Good idea

laflannery: Those are the entity IDs for [Lovell TRICARE](https://prod.cms.va.gov/section/vha/vamc-facilities/visn-12/lovell-tricare) and [Lovell VA](https://prod.cms.va.gov/section/vha/vamc-facilities/visn-12/lovell-va). You can see them in the Section Taxonomy if you edit the term. There is also [a comment in the content-build file](https://github.com/department-of-veterans-affairs/content-build/blob/d25325957f43309d349d61274537ecd0458683da/src/site/layouts/health_care_region_page.drupal.liquid#L55) that might be useful to include here to remind us later that this is to hide this section from TRICARE

laflannery:
@pwolfert These pages aren't working on Tugboat, besides the environment variable I unfortunately don't know much else as to why this wouldn't be showing. Chris or Brian would probably be able to help though

Screenshot 2025-05-07 at 8 56 57 AM

src/templates/layouts/vamcSystem/index.tsx
omahane: Without looking at the code, I was almost certain these were Lovell. Thanks for confirming, @laflannery .

src/pages/api/drupal-query.ts
cvalarida: image

This is what's up. I get it now.

The static type you've defined for the `NodeVamcSystem` doesn't include the `field_clinical_health_services` property yet, so TS couldn't care less about anything in here. When you get around to adding that field for the section of the page that uses it, I expect TS will start getting angry at your unit tests that pull in this mock data.

@laflannery or @omahane: Will there ever be a situation where `thisVamcSystemEntity.field_clinical_health_services.field_region_page` will ever be _not_ a circular reference?

I'm thinking we can just set these to `null` in our mock data because in our `formatter()`, there's never a time where we'll add them to the formatted result.

🤔 In fact...we may need to set them to `null` in `formatter()` when the time comes for this same JSON serialization issue. 🤔 Something something SSR...I'm not sure...

src/templates/layouts/vamcSystem/index.tsx
pwolfert: Thank you!

pwolfert: https://github.com/department-of-veterans-affairs/next-build/pull/1017/commits/78ac4f4278c4910cee133be9e67e5aa9fdb55d61

laflannery:
@pwolfert This is loading now, 2 comments:
- The header image is decorative, not contextual. Which means we should have a null alt -`alt=""` (if you look at prod it has the `alt=""`, `role="presentation"` and `aria-hidden="true"`. All of that is overkill, just null alt is sufficient.
- There is an issue with the TRICARE sidenav not pulling the correct URL for some pages (it goes to VA instead of TRICARE). Anu has fixed this multiple times, [here is the most recent PR](https://github.com/department-of-veterans-affairs/next-build/pull/1016)
  - If you would prefer the sidebar fix go into a separate ticket that's fine as well, just let me know.

src/templates/layouts/vamcSystem/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+        <div className="usa-width-three-fourths">
+          <article className="usa-content va-l-facility-detail vads-u-padding-bottom--0">
+            {/* <LovellSwitcher
+              currentVariant={lovellVariant}
+              switchPath={lovellSwitchPath}
+            /> */}
+            {title && <h1>{title}</h1>}
+            {hasValidImage && (
+              <div className="duotone darken lighten medium-screen:vads-u-margin-bottom--0p5">
+                <MediaImage {...image} imageStyle={imageStyle} />
```
pwolfert, line 65: @laflannery, it looks like the `alt` text is coming from the Drupal data for the image. We use this `<MediaImage>` component and pattern in several different components/pages, which means we might need to watch out for that in other places. I'll go ahead and override the `alt` attribute that we get from Drupal with `""` here.

pwolfert: Oh actually, I lied; I can't just change it here. I would also need to [updated this logic](https://github.com/department-of-veterans-affairs/next-build/blob/pwolfert/vamc-system/src/templates/common/mediaImage/index.tsx#L67) so we explicitly check for an `alt` value of `null` or `undefined` in the `MediaImage` component and allow for explicitly setting `alt=""`. It makes sense that we wouldn't want to allow the attribute to be completely omitted from the element, but we should start allowing for the case where it's only presentational.

pwolfert: CC @cvalarida

pwolfert:
My latest commit introduces changes the way we process Lovell menu data everywhere. There's a lot to explain, but I'll need to do it tomorrow or later tonight.

pwolfert:
The [latest changes](https://github.com/department-of-veterans-affairs/next-build/pull/1017/commits/1eaffd87dc77e6dd2322c4a1dd9e8fdea4da89c5) address Lovell side-nav loading issues that [Laura previously found](https://github.com/department-of-veterans-affairs/next-build/pull/1017#issuecomment-2859932183).

## Summary of changes

1. Further mutation of the side-nav data has been [moved out of the Lovell-specific get-static-props function](src/lib/drupal/lovell/staticProps.ts) and [into the buildSideNavDataFromMenu function](src/lib/drupal/facilitySideNav.ts) where similar mutations were already occuring.
    - I believe that consolidating this business logic into here will make it easier to find everything and reason about all the transformations that are happening.
    - I found [the original PR where this logic was added two years ago](https://github.com/department-of-veterans-affairs/next-build/pull/226). There is some discussion about the best place to put it, but we have the benefit of hindsight.
2. The `buildSideNavDataFromMenu` function now uses the context it already had in the `entityPath` to [determine what Lovell-variant logic to apply](https://github.com/department-of-veterans-affairs/next-build/pull/1017/commits/1eaffd87dc77e6dd2322c4a1dd9e8fdea4da89c5#diff-c931cad709f1a1e5fd69c51a9a3714ed71cc1f6c10bf9fcd4fa573fa2f430bbcR64).
    - This keeps our formatter functions concise and ensures we apply consistent logic. I don't think I'm jumping too quickly into DRYing code here. It seems to me that this logic is the same for all of these Lovell paths.
3. I've updated the following data query formatter functions to use the new pattern: `healthCareLocalFacility.ts`, `pressReleaseListing.ts`, `storyListing.ts`, `vamcSystem.ts`, `eventListing/query.ts`.
    - This increases the QA burden for this PR, but I think it's worth it. CC @laflannery

src/products/eventListing/query.test.ts
```diff
@@ -144,235 +157,7 @@ describe('EventListing formatter', () => {
...
-          severity: 0,
-          items: [],
-        },
-      ],
-    },
-    totalItems: 0,
-    totalPages: 1,
-  }
-
-  const mockDataWithoutEvents = {
```
pwolfert, line 174: This object was unused, and the one below it was unused too

src/lib/drupal/facilitySideNav.test.ts
```diff
@@ -59,109 +58,51 @@ describe('SideNav Data Building Function', () => {
...
-            name: 'entity.node.canonical',
-            parameters: {},
-          },
-          type: 'menu_link_content',
-          meta: {},
-          parent: '',
-        },
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      ] as any[], // Use 'any[]' to bypass detailed type requirements
+    function getLinkByLabel(sideNavData: SideNavMenu, label: string) {
```
cvalarida, line 61: Clever. I like it.

src/templates/common/mediaImage/index.tsx
```diff
@@ -63,7 +63,8 @@ export const MediaImage = (
     return null
   }
 
-  if (!props.alt) {
+  // Allow explicitly setting alt text to "" but not omitting the attribute entirely
+  if (props.alt == null) {
```
cvalarida, line 67: Is there any way for `props.alt` to be `undefined`?

cvalarida (approved):
That's a whole lotta mock data. 😳 Not sure that it's worth it to pare it down, but holy cow.

Also, I'd like to see a little narrower PRs in the future ideally. This one not only scaffolded the content type (which was a lot by itself), but also refactored the menus. While this was a good change, I think that slowed you down quite a bit. 😬 

At any rate, it's looking great! I see a first pass of the VAMC system page and the dealing with the menus is way nicer now. 🎉 

Let's goooo! :shipit:

src/templates/common/mediaImage/index.tsx
pwolfert: From my conversation with Laura, I think the original intent of the code here was to make sure we're not missing the alt text entirely—that we've taken it into consideration even if we set it to an empty string. Since the alt prop would normally come from the `FormattedMediaImage` and that comes from Drupal, I think it's possible that it could be empty...but maybe not `null`. I guess I really don't know.

PR #1044, opened 2025-05-21
Add fetch-entity CLI

cvalarida:
# Description

Adds a utility script to fetch and hydrate entities from Drupal's JSON:API from the command line.

The goal here is to use this for grabbing mock data more easily. I built this because the mock data I was trying to grab had circular references, and every other way I tried to plug that into a mock data file kept failing during serialization. I used this to output Mostly JavaScript™ by the power of `util.inspect()`. From there, I copy-pasta'd that into a `.mock.ts` file, removed the `[Circular *1]` and `<Ref *1>` pointers, and plugged that new variable into the `mockData.field_local_health_care_service_` property.

It's not the most elegant solution, but it works. 😬 

## Screenshots

**Help**
image

**Colored output**
image

**JSON output**
image

## Ticket

Relates to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/20793

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,8 @@
     "test:playwright:a11y:local": "USE_PROXY=false PW_WIDTH_VALUE=768 PW_HEIGHT_VALUE=720 PW_BROWSER_VALUE=chromium yarn playwright test --project=a11y",
     "test:types": "yarn build:env-loader && APP_ENV=test tsc --noEmit",
     "storybook": "storybook dev -p 6006",
-    "storybook:build": "storybook build -o out/storybook-static --quiet"
+    "storybook:build": "storybook build -o out/storybook-static --quiet",
+    "fetch-entity": "NODE_OPTIONS='--disable-warning=ExperimentalWarning' node scripts/fetch-entity/src/index.ts"
   },
   "dependencies": {
     "@actions/core": "^1.10.1",
--- a/scripts/fetch-entity/src/drupalClient.ts
+++ b/scripts/fetch-entity/src/drupalClient.ts
@@ -1,0 +1,44 @@
+/**
+ * This is essentially a copy of what's in the drupalClient.ts module, but
+ * copying it over here for use in the script because importing it from
+ * next-build/src/lib/drupal/drupalClient.ts is a pain.
+ */
+
+import { DrupalClient } from 'next-drupal'
+
+/* eslint-disable import/no-extraneous-dependencies */
+import { SocksProxyAgent } from 'socks-proxy-agent'
+import fetch from 'node-fetch'
+
+const syswideCas = await import('syswide-cas')
+syswideCas.addCAs('certs/VA-Internal-S2-RCA-combined.pem')
+syswideCas.addCAs('certs/rootCA.pem')
+
+const baseUrl =
+  process.env.NEXT_PUBLIC_DRUPAL_BASE_URL || 'https://va-gov-cms.ddev.site'
+
+const debug = process.env.DEBUG === 'true'
+
+const proxyFetch = (input: RequestInfo, init: RequestInit = {}) => {
+  const agent = new SocksProxyAgent('socks5h://127.0.0.1:2001')
+  return fetch(input, { ...init, agent })
+}
+
+export const drupalClient = new DrupalClient(baseUrl, {
+  fetcher: proxyFetch,
+  useDefaultResourceTypeEntry: true,
+  throwJsonApiErrors: false,
+  auth: {
+    // @ts-expect-error It's possible for the env vars to not be set properly
+    clientId: process.env.DRUPAL_CLIENT_ID,
+    // @ts-expect-error It's possible for the env vars to not be set properly
+    clientSecret: process.env.DRUPAL_CLIENT_SECRET,
+  },
+  previewSecret: process.env.DRUPAL_PREVIEW_SECRET,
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -39,7 +39,8 @@
     "test:playwright:a11y:local": "USE_PROXY=false PW_WIDTH_VALUE=768 PW_HEIGHT_VALUE=720 PW_BROWSER_VALUE=chromium yarn playwright test --project=a11y",
     "test:types": "yarn build:env-loader && APP_ENV=test tsc --noEmit",
     "storybook": "storybook dev -p 6006",
-    "storybook:build": "storybook build -o out/storybook-static --quiet"
+    "storybook:build": "storybook build -o out/storybook-static --quiet",
+    "fetch-entity": "NODE_OPTIONS='--disable-warning=ExperimentalWarning' node scripts/fetch-entity/src/index.ts"
```
cvalarida, line 43: We can run `.ts` files in Node 23.6+ without the `--experimental-strip-types` flag, but it'll still warn you about it. 🤷 

Our `.nvmrc` says 18.17.0, but apparently I've been running on 23.11.0 accidentally for...a while? With no ill effect, so....... 🤷

scripts/fetch-entity/src/drupalClient.ts
cvalarida, line 1: I fought with this for _way_ too long before just copying it over. If you wanna try to figure out how to use the original, be my guest, but it was a pain. 😬 

I also simplified it somewhat since we'll only be running it locally. 🤷

scripts/fetch-entity/src/drupalClient.ts
```diff
@@ -0,0 +1,44 @@
...
+
+const baseUrl =
+  process.env.NEXT_PUBLIC_DRUPAL_BASE_URL || 'https://va-gov-cms.ddev.site'
+
+const debug = process.env.DEBUG === 'true'
+
+const proxyFetch = (input: RequestInfo, init: RequestInit = {}) => {
+  const agent = new SocksProxyAgent('socks5h://127.0.0.1:2001')
+  return fetch(input, { ...init, agent })
+}
```
cvalarida, line 23: Deliberately not using our `proxy-fetch` package.
1. It was too painful to try to include it; module resolution nightmare
2. We don't actually need any of the bells and whistles with `cross-fetch` or `p-retry` 🤷

scripts/fetch-entity/src/drupalClient.ts
```diff
@@ -0,0 +1,44 @@
...
+    clientSecret: process.env.DRUPAL_CLIENT_SECRET,
+  },
+  previewSecret: process.env.DRUPAL_PREVIEW_SECRET,
+  // Add request header to tell the CMS to return public-facing URLs for files.
+  headers: {
+    'Content-Type': 'application/vnd.api+json',
+    Accept: 'application/vnd.api+json',
+    'File-Public-Base-Url-Check': 'true',
+  },
+})
```
cvalarida, line 42: Some of this may not be necessary, but 🙈

scripts/fetch-entity/src/getEnvVars.ts
cvalarida, line 1: Same story with this one; including it from `packages/env-loader` was a module resolution nightmare, so I just copied it over and simplified.

scripts/fetch-entity/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../../tsconfig.json",
+  "include": ["./src"],
+  "compilerOptions": {
+    "strict": true
```
cvalarida, line 5: We may start incrementally adding strict mode by virtue of lots of little `tsconfig.json`s scattered around the repo. 🤔 That's a problem for future us, though.

brianseek (changes requested):
This is really neat! But it doesn't work with the node version set in next-build (18)
`/versions/node/v18.19.0/bin/node: --disable-warning= is not allowed in NODE_OPTIONS`

If I switch to v23.11.0 that you are using, it works, but then yarn also wanted to upgrade. I think we should update the project to v23 or newer first, or re-work this to work with v18.

cvalarida:
> This is really neat! But it doesn't work with the node version set in next-build (18) `/versions/node/v18.19.0/bin/node: --disable-warning= is not allowed in NODE_OPTIONS`
> 
> If I switch to v23.11.0 that you are using, it works, but then yarn also wanted to upgrade. I think we should update the project to v23 or newer first, or re-work this to work with v18.

So...that's fair. Buuuuuuuut if I do that, I wouldn't be able to work in the circular reference bit from [this PR](https://github.com/department-of-veterans-affairs/next-build/pull/1050) because it uses TypeScript. 😅 

As long as the dependencies are self-contained (either node modules or local-to-this-script files), it's trivial to make the change to use `// @ts-check` and `/** @type <whatever> */` per TS's docs (this [intro doc](https://www.typescriptlang.org/docs/handbook/intro-to-js-ts.html) and [this how-to doc](https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html)). In fact, I just had ChatGPT make that conversion right quick. No problem. But then I couldn't serialize data with circular references using the object graph deflator I wrote up, and that's gonna be important later.

Instead, I just added a shell script that explicitly runs this TypeScript module using a version of Node that _will_ work. It won't affect any of the rest of the repo this way. The alternatives were...less pleasant. 😅 😬

PR #1046, opened 2025-05-22
Simplify `field_administration` usage across content types

pwolfert:
# Description

[Context here](https://github.com/department-of-veterans-affairs/next-build/pull/1040#discussion_r2103388274)

We were inconsistently referencing and using `field_administration` data in our various content types and components, and this aims to make it consistent and simplify it.

## Ticket

Related to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21250

### DIFF
--- a/src/products/event/template.tsx
+++ b/src/products/event/template.tsx
@@ -28,7 +28,8 @@ export const Event = ({
   cost,
   registrationRequired,
   socialLinks,
-  administration,
+  administrationId,
+  administrationName,
   link,
   additionalInfo,
   eventCTA,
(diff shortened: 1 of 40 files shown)


src/products/event/template.tsx
```diff
@@ -330,17 +331,17 @@ export const Event = ({
...
             <p>
               <va-link
                 href={listing}
                 onClick={() =>
                   recordEvent({ event: 'nav-secondary-button-click' })
                 }
                 id="see-more-events"
-                text={`Browse the ${administration.name} events calendar`}
+                text={`Browse the ${administrationName} events calendar`}
               ></va-link>
```
pwolfert, line 345: The solution for simplifying was based on my earlier evaluation that only the id was ever used in the site, but I was wrong. All the changes were based on that disproven assumption. But now the work day is over, so I'm going to leave this here.

pwolfert: Because of this, I created [an alternate path](https://github.com/department-of-veterans-affairs/next-build/pull/1048), which is what @cvalarida had suggested in the first place.

PR #1048, opened 2025-05-23
Unify `field_administration` handling across content types

pwolfert:
# Description

This is an alternative to https://github.com/department-of-veterans-affairs/next-build/pull/1046. Because [the event template uses the administration name](https://github.com/department-of-veterans-affairs/next-build/pull/1046#discussion_r2103568654), it may not be the best to simplify everything down to just the `administrationId`.

## Ticket

Relates to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21250

### DIFF
--- a/src/types/formatted/administration.ts
+++ b/src/types/formatted/administration.ts
@@ -3,7 +3,7 @@ export interface Administration {
    * This is the internal Drupal id of the administration term. See the administration
    * formatter function comments for more details.
    */
-  entityId: number
+  entityId: number | null
   /**
    * The human-readable name of the administration term.
    */


src/data/queries/administration.ts
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Formats a Drupal administration field into our standardized Administration interface.
+ *
+ * In our Drupal content, the `field_administration` field points to a taxonomy Term
+ * that is currently called "Section", but "Section" is only a human label added to
+ * it in the Drupal interface, and that name has evolved over time. It was originally,
+ * "administration" (and thus the name of the field). Since the only machine name of
+ * it is `field_administration`, we'll continue to call it "administration" in our code.
+ */
+export const formatAdministration = (
```
cvalarida, line 13: I don't know if I care yet, but this _does_ deviate from the pattern. In `src/data/queries/*.ts`, we export `formatter()` and then `import { formatter as formatAdministration } from '...'` rather than defining it in here as `formatAdministration`.

For consistency, we should probably keep it the same, but I'm not strongly opinionated about this.

src/types/formatted/administration.ts
```diff
@@ -1,4 +1,4 @@
-export type Administration = {
-  id: number
-  name: string
+export interface Administration {
+  entityId: number
```
cvalarida, line 2:
```suggestion
  entityId: number | null
```

It _shouldn't_ be, but our formatter is returning `null` for this in some cases. 🙈

cvalarida: I also recommend adding a `/**` comment on the property to describe _which_ ID this is, since there's some confusion around that. 😅 It should be clear when you read the `formatAdministration` function, but this static type definition and that function are far enough removed from each other to make it a cognitive leap to even _know_ you should look there, y'know?

cvalarida (approved):
Looks good! That's quite a few changes, but they're all one-liners.

I'm not in love with adding `name` where we didn't need it before, but that's a small price to pay for consistency, so I'm here for it. 👍

src/data/queries/administration.ts
```diff
@@ -0,0 +1,18 @@
...
+  administration: FieldAdministration | null | undefined
+): Administration => ({
+  entityId: administration?.drupal_internal__tid || null, // "t" is for taxonomy term
+  name: administration?.name || null,
```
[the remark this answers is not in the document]
pwolfert: Good idea 👍

src/data/queries/administration.ts
pwolfert: 72183573

PR #1050, opened 2025-05-24
Circular reference flattener

cvalarida:
# Description

This PR creates an object graph flattener & re-hydrator.

## The problem
We were running into here was that some entities from Drupal have circular references that are really hard to avoid. (And I anticipate scenarios where we _shouldn't_ avoid them, though none come to mind right now.) When the objects are held in memory, there's no problem with that. Because we're using SSR, however, we need to serialize the `resource` to send it from the server to the client for rehydration client-side.

> [!Note]
> I'm not 100% on _why_ we need need this serialization during static site generation (SSG), but it's a Next.js thing. Our SSG (`yarn export`) runs into this problem when we have circular references in our `resource` from `getStaticProps()`. 🤷 

## The solution
Flatten the object graph to avoid circular dependencies so we can serialize the data from `getStaticProps()` to `<ResourcePage>`.

> [!Important]
> **Bonus points:** We'll be able to use this to serialize and store mock data with circular references too, then rehydrate it in our test file.

## Generated description

This pull request introduces a utility for serializing and deserializing object graphs with circular or repeated references, integrates it into the `ResourcePage` component, and adds comprehensive test coverage. The main changes include the implementation of the utility, its integration into the static props pipeline, and the addition of tests to ensure correctness.

### Implementation of object graph serialization/deserialization utility:
* [`src/lib/utils/object-graph.ts`](diffhunk://#diff-db5f5ba4a73be6724f3e400ef0c98d72c6719ef49711d2997e8eeb9907b89e3cR1-R172): Added `flattenObjectGraph` and `inflateObjectGraph` functions to serialize and deserialize object graphs with circular or repeated references. Includes support for deduplication, circular references, and preserving object identity.

### Test coverage for the utility:
* [`src/lib/utils/object-graph.test.ts`](diffhunk://#diff-c66f0b04db1fc4132f005cde5b3409d5fd0b9ccbc695b94cae9ef0111cbee036R1-R146): Added comprehensive tests for the `flattenObjectGraph` and `inflateObjectGraph` functions, covering scenarios like nested objects, arrays, circular references, shared objects, and edge cases like `null` and `undefined`.

### Integration into `ResourcePage` component:
* `src/pages/[[...slug]].tsx`:
  - Imported the new utility functions and types. ([src/pages/[[...slug]].tsxR28-R32](diffhunk://#diff-5f218029560343be88582c198bb3b04e4c18259a5e716a88250afabe4466f957R28-R32))
  - Updated the `ResourcePage` component to accept a serialized resource (`serializedResource`) and deserialize it using `inflateObjectGraph`. ([src/pages/[[...slug]].tsxL98-R117](diffhunk://#diff-5f218029560343be88582c198bb3b04e4c18259a5e716a88250afabe4466f957L98-R117))
  - Updated `getStaticProps` to serialize the `resource` object using `flattenObjectGraph` before passing it to the component. ([src/pages/[[...slug]].tsxL318-R327](diffhunk://#diff-5f218029560343be88582c198bb3b04e4c18259a5e716a88250afabe4466f957L318-R327))

## Ticket

References https://github.com/department-of-veterans-affairs/va.gov-cms/issues/20793


src/lib/utils/object-graph.test.ts
cvalarida, line 1: I tried to make sure this was pretty thorough, but these are ChatGPT-generated test cases. If you think of a test case that's not caught in here, please please please leave a comment! 🙏 

This functionality needs to be _super_ stable since it'll affect _every_ page.

src/lib/utils/object-graph.ts
```diff
@@ -0,0 +1,172 @@
...
+ */
+export interface Reference {
+  __refId: string
+}
+
+/**
+ * Serializes an object containing circular and/or repeated references into a
+ * JSON-safe format by extracting duplicates into an `include` map.
+ */
+export function flattenObjectGraph<T>(input: T): FlattenedGraph<T> {
```
cvalarida, line 25: This runs at `O(n)` time and space complexity (where `n` is the number of nodes in the object graph)...but realistically, that's still `Θ(2n)` time and `Θ(3n)` space complexity (if I'm not mistaken...which I very well may be, given that we're storing by reference, not duplicate objects 🫠 ), which isn't ideal. I wasn't sure how to get that down.

Some of our object graphs are thousands of nodes (which is so silly), so if you can figure out how to optimize this, please let me know!

I don't _think_ this will cause meaningful performance degradation, but I also haven't benchmarked it. 🙈 I have an idea how I might do that, though... just gotta get [my entity-fetching CLI PR merged first](https://github.com/department-of-veterans-affairs/next-build/pull/1044). 😜

src/lib/utils/object-graph.ts
cvalarida, line 1: **Note:** Much of this is by the power of ChatGPT with my guidance on some implementation details. If you see anything weird, it's more than fair game to call it out.

src/lib/utils/object-graph.test.ts
```diff
@@ -0,0 +1,146 @@
...
+  it('handles circular references', () => {
+    const obj: Record<string, unknown> = { name: 'Selfie' }
+    obj.self = obj
+
+    const result = flattenObjectGraph(obj)
+    const uuids = Object.keys(result.include)
+    expect(uuids.length).toBe(1)
+
+    const roundTrip = inflateObjectGraph(result)
+    expect(roundTrip.self).toBe(roundTrip)
```
pwolfert, line 48: Clever assertion

src/lib/utils/object-graph.test.ts
```diff
@@ -0,0 +1,146 @@
...
+      c: { d: undefined, e: null },
+    }
+
+    const result = flattenObjectGraph(input)
+    const roundTrip = inflateObjectGraph(result)
+
+    expect(roundTrip.a).toBeNull()
+    expect(roundTrip.b).toBeUndefined()
+    expect(roundTrip.c.e).toBeNull()
+    expect('d' in roundTrip.c).toBe(true) // optional: assert 'undefined' is retained
```
pwolfert, line 144: Should we be making the same two assertions for `b` and `d`? Test that `d` is undefined but that `b` is a key in the object?

And is "skips undefined" an accurate description of what's happening? From reading the tests, it seems that `{ b: undefined }` gets translated into exactly the same thing in the round-trip object (value of undefined but the key exists).

src/lib/utils/object-graph.ts
```diff
@@ -0,0 +1,172 @@
...
+  const serializing = new Set<string>()
+
+  /**
+   * First pass: detects duplicates and assigns UUIDs to all objects. Only marks
+   * objects as `isDuplicate` if seen more than once.
+   */
+  function findDuplicates(value: unknown) {
+    if (typeof value !== 'object' || value === null) return
+
+    const existing = seen.get(value)
```
pwolfert, line 44: Oh interesting, the object itself is the key. I didn't know you could do that with JavaScript Maps!

pwolfert: I see it in the MDN docs now. I'm always serializing objects into string keys by hand 😂😭

src/lib/utils/object-graph.ts
```diff
@@ -0,0 +1,172 @@
...
+  // Guards against infinite recursion by tracking active __refIds during
+  // serialization
+  const serializing = new Set<string>()
+
+  /**
+   * First pass: detects duplicates and assigns UUIDs to all objects. Only marks
+   * objects as `isDuplicate` if seen more than once.
+   */
+  function findDuplicates(value: unknown) {
+    if (typeof value !== 'object' || value === null) return
```
pwolfert, line 42: I forgot that `typeof null` is `object`. You tricky JavaScript

src/lib/utils/object-graph.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    const existing = seen.get(value)
+    if (existing) {
+      existing.isDuplicate = true
+      return
```
pwolfert, line 47: Maybe it's worth commenting that without this return statement we could infinitely recurse.

src/lib/utils/object-graph.ts
```diff
@@ -0,0 +1,172 @@
...
+   * `include` and referenced by ID.
+   */
+  function walk(value: unknown): unknown {
+    if (typeof value !== 'object' || value === null) return value
+
+    const entry = seen.get(value)
+    if (entry?.isDuplicate) {
+      if (!(entry.__refId in include)) {
+        // If we're already serializing this __refId (due to circularity), just
+        // return the ref
```
pwolfert, line 73: I didn't understand the order of events here when I read this comment and didn't realize what was happening until I saw the `remove` at the end. It might help me understand better if it said "If we're _currently_ serializing..."

src/lib/utils/object-graph.ts
```diff
@@ -0,0 +1,172 @@
...
+   * Recursively resolves `{ __refId }` references into actual object instances.
+   * Uses a cache to prevent reprocessing and to rebuild cycles correctly.
+   */
+  function resolve(value: unknown): unknown {
+    if (Array.isArray(value)) {
+      return value.map(resolve)
+    }
+
+    if (typeof value === 'object' && value !== null) {
+      if ('__refId' in value && typeof value.__refId === 'string') {
```
pwolfert, line 138: I realize this function is undoing everything, but is there a technical reason why this function is structured in the reverse of the first two recursive functions (where we started with scalar values at the top)?

src/lib/utils/object-graph.ts
```diff
@@ -0,0 +1,172 @@
...
+        if (refData === undefined) {
+          throw new Error(`Reference id ${__refId} not found in include map`)
+        }
+
+        // Create placeholder to support circular references
+        const shell: Record<string, unknown> = {}
+        cache.set(__refId, shell)
+
+        const resolved = resolve(refData)
+        Object.assign(shell, resolved)
```
pwolfert, line 156: It would require too much brain power and time right now for me to reason through the sequence of events here, but I trust you and the automated tests. Instinctively this feels right. 🤯

pwolfert (approved):
I haven't actually executed it locally, but as far as I can tell the code looks perfect. I love that the serialization and deserialization is so localized to the page renderer! Looks so good 😍

src/lib/utils/object-graph.ts
cvalarida: Ah, yeah. Good call. 👍

src/lib/utils/object-graph.ts
cvalarida: Nope! I just changed it for consistency. 👍

src/lib/utils/object-graph.test.ts
```diff
@@ -0,0 +1,146 @@
+import { flattenObjectGraph, inflateObjectGraph } from './object-graph'
+
+describe('data-serializer', () => {
+  it('serializes and deserializes a simple object', () => {
```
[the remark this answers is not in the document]
cvalarida: Yeah, good point. As long as we're being nitpicky, I don't think it technically even _serializes_ at all. 😅 It does deduplicate object references, though, so 🤷 

I just changed it to "handles objects with only primitives" to be accurate and precise. 😜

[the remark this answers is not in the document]
brianseek: yeah that's fair. works for me!

src/lib/utils/object-graph.test.ts
cvalarida: Oh, y'know what? "Skips undefined" is 100% wrong. Good catch! If we tried to serialize the output of `flattenObjectGraph` into JSON when the input contained `undefined` values, the JSON would just drop the properties altogether. That's not what this does...and honestly, I think that's fine? 🤔

Unless you've got an argument to change the behavior to _actually_ drop properties whose values are `undefined`, I'll just change the description of the test.

src/lib/utils/object-graph.ts
```diff
@@ -0,0 +1,171 @@
...
+        return cache.get(__refId)
+      }
+
+      const refData = include[__refId]
+      if (refData === undefined) {
+        throw new Error(`Reference id ${__refId} not found in include map`)
+      }
+
+      // Create placeholder to support circular references
+      const shell: Record<string, unknown> = {}
```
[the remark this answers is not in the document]
cvalarida: Okay, yeah. I wrote this test:
```ts
  it('handles circular array references', () => {
    const array: unknown[] = ['b']
    const input = {
      a: array,
    }
    array.push(array)

    const result = deflateObjectGraph(input)
    const roundTrip = inflateObjectGraph(result)

    expect(roundTrip.a[1]).toBe(roundTrip.a)
    expect(Array.isArray(roundTrip.a)).toBe(true)
  })
```
And it fails because `roundTrip.a` isn't an array. Good catch! Look at that, Copilot coming in clutch. 😂

It shouldn't ever happen, but I'll go ahead and be paranoid about it. 🤓 Changes incoming!

PR #1059, opened 2025-05-28
Add repomix config

cvalarida:
# Description

Adds a [repomix](https://github.com/yamadashy/repomix) config for plugging into the likes of ChatGPT.

It's a low-hanging fruit for incorporating some AI into template migration and can be useful for things like describing what you need and seeing if there's already an existing component that does what you want. I intend to use it to help out with template migration.

### DIFF
--- a/.repomixignore
+++ b/.repomixignore
@@ -1,0 +1,21 @@
+# Jest snapshots
+*.snap
+
+# Storybook stories
+*.stories.*
+
+# Mock data
+*.mock.*
+src/products/**/mock.js*
+
+# Add the path to the mock file you want to include here.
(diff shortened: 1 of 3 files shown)


.repomixignore
```diff
@@ -0,0 +1,21 @@
+# Jest snapshots
+*.snap
+
+# Storybook stories
+*.stories.*
+
+# Mock data
+*.mock.*
+src/products/**/mock.js*
+
```
brianseek, line 10: Should `src/mocks/*` also get added here?

cvalarida: Everything in there would get covered by `*.mock.*`, but no harm in including it. 👍

PR #1063, opened 2025-05-30
Add stories section to VAMC System page

pwolfert:
# Description

- Updates the query and types to get the featured stories
- Renders up to two of them in the VAMC System page

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21254

## Screenshots

Screenshot 2025-05-30 at 12 17 27 PM

### DIFF
--- a/src/data/queries/tests/__snapshots__/vamcSystemVaPolice.test.tsx.snap
+++ b/src/data/queries/tests/__snapshots__/vamcSystemVaPolice.test.tsx.snap
@@ -113,6 +113,13 @@ exports[`VamcSystemVaPolice formatData outputs formatted data 1`] = `
   ],
   "moderationState": "published",
   "path": "/central-arkansas-health-care/va-police",
+  "phoneNumber": {
+    "extension": "",
+    "id": "2bc3cc07-58a2-4d7f-b639-a90e0b890d7b",
+    "number": "501-257-6550",
+    "phoneType": "tel",
+    "type": "paragraph--phone_number",
+  },
   "policeOverview": {
     "html": "<p>VA police officers help make VA medical centers and other VA health facilities safe for Veterans and their family members.</p>
 ",
@@ -120,6 +127,7 @@ exports[`VamcSystemVaPolice formatData outputs formatted data 1`] = `
     "type": "paragraph--wysiwyg",
   },
   "published": true,
+  "system": "VA Central Arkansas health care",
   "title": "VA police",
   "type": "node--vamc_system_va_police",
 }
--- a/src/data/queries/vamcSystemVaPolice.ts
+++ b/src/data/queries/vamcSystemVaPolice.ts
@@ -1,6 +1,5 @@
 import { QueryData, QueryFormatter, QueryParams } from 'next-drupal-query'
 import { DrupalJsonApiParams } from 'drupal-jsonapi-params'
-import { drupalClient } from '@/lib/drupal/drupalClient'
 import { NodeVamcSystemVaPolice } from '@/types/drupal/node'
 import { VamcSystemVaPolice } from '@/types/formatted/vamcSystemVaPolice'
 import { RESOURCE_TYPES } from '@/lib/constants/resourceTypes'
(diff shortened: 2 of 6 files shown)


src/data/queries/newsStoryTeaser.ts
```diff
@@ -22,7 +22,7 @@ export const formatter: QueryFormatter<NodeNewsStory, NewsStoryTeaser> = (
     published: entity.status,
     headingLevel: 'h2', //@todo fix headingLevel,
     title: entity.title,
-    image: queries.formatData('media--image', entity.field_media), //cropType: '2_1_large'
+    image: formatImage(entity.field_media), //cropType: '2_1_large'
```
pwolfert, line 25: This line was failing my unit tests due to the following error:

```
 FAIL  src/templates/layouts/vamcSystem/index.test.tsx
  ● Test suite failed to run

    You should not call getQueryData on the client. This is a server-only call.

      23 |     headingLevel: 'h2', //@todo fix headingLevel,
      24 |     title: entity.title,
    > 25 |     image: queries.formatData('media--image', entity.field_media), //cropType: '2_1_large'
         |                    ^
      26 |     link: `${entity.path.alias}`,
      27 |     introText: entity.field_intro_text,
      28 |     lastUpdated: entity.field_last_saved_by_an_editor || entity.created,

      at Object.formatData (node_modules/next-drupal-query/src/index.ts:337:13)
      at formatData (src/data/queries/newsStoryTeaser.ts:25:20)
          at Array.map (<anonymous>)
      at map (src/data/queries/vamcSystem.ts:133:38)
      at Object.<anonymous> (src/templates/layouts/vamcSystem/index.test.tsx:34:27)
```

laflannery:
@pwolfert A few comments:

1. The "See all stories" link needs to be wrapped in a `<p>` tag:
Screenshot 2025-06-02 at 8 36 50 AM
2. Also on this link there is a `classname` attribute. `classname` does not work here, we need to use `class` - you can see that at the end with `class="hydrated"`. However, in this case we don't need any of these classes so everything here can be removed.
Screenshot 2025-06-02 at 8 54 07 AM
3. The Stories titles need to be `<h3>` not `<h2>`. Also, there are some vads classes that are changing the font sizes here. I took a look and these aren't necessary, we can remove these:
Screenshot 2025-06-02 at 8 41 07 AM
4. The mobile & tablet display isn't quite matching Production:
   1. The Stories section is going to a 1 column mobile view too soon. On Prod currently, this doesn't go to the image on top until 479px. If you change where you have `tablet` to `mobile-lg` this should resolve this.
   1. However, on tablet breakpoints (768px - 480px) the ratio of the columns is 50/50 rather than what is previously was on desktop. 
   2. Using `tablet:vads-grid-col-8 mobile-lg:vads-grid-col-6` seems like it should account for the display on all breakpoints. 



Also a note for future reference and my own sake - The ratio for the images here is 2:1, which is a change from Prod (Prod currently uses 3:2). However, this is a change we are keeping because the stories listing using 2:1 ratio and Drupal crop uses 2:1 so this is an intended change and is better for UX

pwolfert:
oh okay, the heading was in the shared component, then, which I'm not going to use anymore. Another confusing thing about that drupal template was that there's an `isStoriesPage` variable, but that template is only used on the system page. Must have been unified at one point in the past and then split, and this is vestigial.

laflannery:
@pwolfert for the most part this looks good, thanks for updated it. Just 2 comments

1. "Stories" needs to be reverted back to an `<h2>`
2. This will eventually need to be updated to use v3 grid styles. Is that something you would like to do now or would you rather just merge this and get it done and then update the grid pieces later in a separate ticket?

src/data/queries/newsStoryTeaser.ts
brianseek: @omahane was having the [same issue ](https://github.com/department-of-veterans-affairs/next-build/pull/1064#discussion_r2123877339) and it stumped me, but @cvalarida knew what was up. I like your solution here. Makes sense

brianseek (approved):
Looks good to me. Works great locally

src/data/queries/newsStoryTeaser.ts
pwolfert: I don't know if it's the right solution. I just didn't want to battle the tests when I saw that this function worked as an alternative 🤷‍♂️

PR #1067, opened 2025-06-02
Add "Get updates" section to VAMC System page

pwolfert:
# Description

- Added social links section to VAMC System page
- Added the corresponding formatter code
- Added unit tests
- Fixed some type errors (see comment below)

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21256

## Screenshots

Right now I'm using the same component as we use on the local facility page, and the two columns are smushed together. I don't know if it began this way or was a side-effect of a grid migration (@cvalarida, @moanu404, @laflannery).

Screenshot 2025-05-30 at 10 12 07 PM


src/templates/layouts/healthCareLocalFacility/FacilitySocialLinks.tsx
```diff
@@ -1,13 +1,8 @@
...
-interface SocialLink {
-  uri: string
-  title: string
-}
-
-interface OperatingStatus {
-  url: string
-}
+type SocialLink = Pick<FieldLink, 'uri' | 'title'>
+type OperatingStatus = Pick<FieldLink, 'url'>
```
pwolfert, line 5: Updated these types because TypeScript was made about my formatter function for `FieldLink` not having nullable `url` values. I looked at the local facility formatter, and it looks like the reason it didn't catch it is because the `field_operating_status` is `any` in that formatter.

field_operating_status is any type

laflannery:
@pwolfert A few comments regarding the new v3 grid markup:
- At 768px, this needs to change to a single column display. With v3, I believe this would be 640px rather than 768px however currently all breakpoints are remaining 2 columns which is incorrect
- There is know space between the columns when it is a 2 columns layout. We should be able to add this by adding `vads-grid-gap` at the row level. I reviewed and while this makes the gap slightly larger than currently it matches the wrap of the links better than `vads-grid-gap-sm`
- On Prod, there is a CSS style that is removing the bottom margin from the last child link, I'm not seeing this on Tugboat, do you know why that would be the case?
Screenshot 2025-06-02 at 1 37 21 PM

pwolfert:
Regarding the `.social-links` rule, it looks like we've been moving bits of an old stylesheet from `content-build` called [sass/style.scss](https://github.com/department-of-veterans-affairs/content-build/blob/main/src/site/assets/sass/style.scss) as needed, but we haven't just copied all of it. Our version is [styles/global.css](https://github.com/department-of-veterans-affairs/next-build/blob/pwolfert/vamc-system-social/src/assets/styles/globals.css). Rather than porting it over, I could attempt to do the same thing without it using utility classes and JavaScript. Seems like we're trying to keep our custom CSS to a minimum.

laflannery:
Yes correct, we want everything inheriting that 1.06rem size, that was a change lat year.

laflannery:
@pwolfert This looks much better! Just a few comments:

- I think we should be using `tablet` instead of `medium-screen` at line 129
- I'm seeing `vads-u-margin-bottom--0` instead of `vads-u-margin-bottom--2` in consistently on some of the links. For example:
  - "Emergency notification" link (the second link in the first column) on Boston: https://pr1067-fputvsxxkbbr70f8xgmzvxc8sfz8zp5l.tugboat.vfs.va.gov/boston-health-care/ and Loma Linda: https://pr1067-fputvsxxkbbr70f8xgmzvxc8sfz8zp5l.tugboat.vfs.va.gov/loma-linda-health-care/
  - "News and announcement" link (the first link in the first column) on Huntington: https://pr1067-fputvsxxkbbr70f8xgmzvxc8sfz8zp5l.tugboat.vfs.va.gov/huntington-health-care/ and Tuscaloosa: https://pr1067-fputvsxxkbbr70f8xgmzvxc8sfz8zp5l.tugboat.vfs.va.gov/tuscaloosa-health-care/
- I also found another grid-gap value `vads-grid-gap-md` I think this would be better. The `sm` seems very small, especially compared to prod and on some of VAMC system and facilty pages with longer names.

pwolfert:
@laflannery, fixed those items!

For the margin classes issue, I forgot to update my reference to a variable when I moved some code around, but it's fixed, and I've added a unit test to catch regressions.

PR #1073, opened 2025-06-05
Add events section to the VAMC system page

pwolfert:
# Description

- Adds the events section
- Migrates to v3 grid (with decision to not align the columns between when/where rows and instead match the events page)
- Adds appropriate queries
- Creates a new event teaser component

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21255

## Screenshots

Screenshot 2025-06-10 at 5 49 04 PM


laflannery:
@pwolfert The Tugboat is failing, are you able to take a look?

laflannery:
Comments:

- Can you change the wrapping `<div>` to a `<p>` and also on the "See all events" link so that the text sizes properly? It's looks like this when I have large font:
Screenshot 2025-06-05 at 1 09 24 PM
- I think the date querying might still be a little off, I've got a couple examples:
   - Central Texas is showing a past event: https://pr1073-rkx8cnbixq0evftqeez5jlwgayspysqu.tugboat.vfs.va.gov/central-texas-health-care/
      - It should be showing the event that is on Prod currently: https://www.va.gov/central-texas-health-care/
   - Greater LA is showing an event that isn't the next most upcoming: https://pr1073-rkx8cnbixq0evftqeez5jlwgayspysqu.tugboat.vfs.va.gov/greater-los-angeles-health-care/
      - The Women Veterans Social Group is on June 10
      - The Bunker Social event on Prod (https://www.va.gov/greater-los-angeles-health-care/) is on June 9 which is before the event showing on Tugboat

pwolfert:
I'm not sure what's happening with those, but I'll try to investigate. I was hopefuly that it was working because I see the same event for https://www.va.gov/providence-health-care/ as I do locally

laflannery:
@pwolfert On tugboat I saw the correct event on Providence as well. Locally, are you seeing the correct events (the ones on Prod) or the incorrect ones (the ones on Tugboat) for the Systems I pointed out above? Because if you are seeing those correct on local, maybe it is a tugboat issue

pwolfert:
I'm seeing the same thing locally as tugboat, and I'm thinking that data is out of date. At least in the case of [Los Angeles](https://pr1073-rkx8cnbixq0evftqeez5jlwgayspysqu.tugboat.vfs.va.gov/greater-los-angeles-health-care/), it's not finding any featured events, so it's showing the next non-featured event.

pwolfert:
Okay, it works like production now!

Evidence

See [this commit message](https://github.com/department-of-veterans-affairs/next-build/pull/1073/commits/d23814827323fc4fa58446fff8a379d237205a73) for an explanation of what changed

laflannery:
Couple things for the "See all events" link:
- It needs to be wrapped in a `<p>` tag so it can resize
- There is also a `classname` attribute that is incorrect (it should be just `class`) however we don't need any of these classes so everything here can just get removed entirely
Screenshot 2025-06-06 at 3 13 22 PM

Also, I'm stating this for future reference:
- This contains a change from production. Currently if there are no featured or upcoming events, we still show the "Events" heading and the "See all events" link. However, we are no longer do that here. I confirmed with Michelle Middaugh and it does make sense because we are no longer showing past events so there is not a lot of benefit to including this section without upcoming events anymore.

laflannery (approved):
Visual QA looks good!

src/templates/layouts/vamcSystem/index.tsx
```diff
@@ -48,6 +50,13 @@ export function VamcSystem({
   const imageStyle = '7_2_medium_thumbnail'
   const hasValidImage = image?.links?.[imageStyle]?.href
 
+  // console.log('featuredEvents', featuredEvents)
+  // console.log('otherEvents', otherEvents)
+  // if (typeof window !== 'undefined') {
```
moanu404, line 55: should this be removed?

src/products/event/query-utils.ts
```diff
@@ -0,0 +1,44 @@
+import { NodeEvent } from '@/types/drupal/node'
+
+/**
+ * Recurring events include entries in their field_datetime_range_timezone for each
+ * occurrence of the event. This function returns the first occurrence of each event
+ * after the timestamp. For non-recurring events, it won't touch the event but will
+ * filter out any events before the timestamp. Remaining events will then be sorted from
+ * earliest to latest.
```
moanu404, line 8: thanks for the detailed description!

src/products/event/query-utils.test.ts
```diff
@@ -0,0 +1,232 @@
...
+      expect(originalEvents[0].field_datetime_range_timezone).toHaveLength(
+        originalLength
+      )
+      expect(originalEvents[0].field_datetime_range_timezone[0].value).toBe(
+        pastDateTime1
+      )
+    })
+  })
+
+  describe('complex scenarios', () => {
```
moanu404, line 197: this test file is lovely - with all of the different scenarios tested

src/templates/components/eventTeaser/index.test.tsx
```diff
@@ -0,0 +1,223 @@
...
+            },
+            title: 'Sample Medical Center',
+          },
+        },
+        fieldLocationHumanreadable: '123 Main St, City, State 12345',
+      }
+
+      const { container } = render(<EventTeaser {...mockEventWithFacility} />)
+
+      expect(screen.getByText('Where')).toBeInTheDocument()
```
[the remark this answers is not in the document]
laflannery: "Where" is not a heading and shouldn't be, it's just bold text so unfortunately I don't think we can look for much more than the text here. However, I do want to point out that "Where" won't always show for every event, [here's an example](https://pr1073-rkx8cnbixq0evftqeez5jlwgayspysqu.tugboat.vfs.va.gov/greater-los-angeles-health-care/). Is that important for what this is testing?

src/templates/layouts/vamcSystem/index.tsx
pwolfert: Oh that was actually me wanting to be able to access the data in the browser console but not blow up my terminal console during SSR

src/products/event/query-utils.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * This is based on [this function from vets-website that drives the event listing page](https://github.com/department-of-veterans-affairs/vets-website/blob/main/src/applications/static-pages/events/helpers/index.js#L78).
```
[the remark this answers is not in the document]
pwolfert: Good catch

src/data/queries/tests/__snapshots__/vamcSystem.test.tsx.snap
```diff
@@ -22,6 +22,8 @@ exports[`VamcSystem formatData outputs formatted data 1`] = `
   ],
   "entityId": 8525,
   "entityPath": "/new-york-harbor-health-care",
+  "fallbackEvent": null,
+  "featuredEvents": [],
```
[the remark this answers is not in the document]
pwolfert: Huh, you're right. It's not supposed to be empty. 🤔

PR #1075, opened 2025-06-05
VAMC Facility health services scaffold

cvalarida:
# Description

This gets us _close-ish_ to the health services, but the actual contents of the components are created by ChatGPT and need to be fixed up. Consider this an extensive scaffolding PR. 😅 

I'm gonna follow this up with a few PRs to:
- Update the mock data
  - We're not filtering by _published_ health services, so the mock data we've got now isn't helpful; all the health services are unpublished
- Add tests
- Fix whatever rendering bugs we discover (there are a few that are immediately obvious, and I'm sure we'll discover more along the way)

# Ticket

Related to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/20793

# Screenshots

The list renders!
image

The insides get a little...weird...
image


src/data/queries/healthCareLocalFacility.ts
```diff
@@ -167,6 +170,73 @@ export const formatter: QueryFormatter<
...
+            serviceTaxonomy?.field_tricare_description ?? null,
+          description: serviceTaxonomy?.description ?? null,
+          entityId: serviceTaxonomy.id,
+          entityBundle: healthService.type.split('--')[1],
+          fieldBody:
+            healthService.field_regional_health_service.field_body.processed,
+          locations: healthService.field_service_location.map((location) => {
+            return {
+              fieldReferralRequired: healthService.field_referral_required,
+              fieldTelephone: formatPhone(entity.field_telephone),
```
cvalarida, line 203: There's no reason this should live in the `locations` object when it's not actually a property of the location. This data should just be passed into the component as a separate prop.

But for now, I'm just reaching for 1:1 parity with the `content-build` template. Once we get it rendering like it should and have sufficient tests in place, we can refactor to our hearts' content.

src/data/queries/healthCareLocalFacility.ts
```diff
@@ -167,6 +170,73 @@ export const formatter: QueryFormatter<
...
+              fieldPhoneNumber: entity.field_phone_number,
+              isMentalHealthService: serviceTaxonomy.name
+                .toLowerCase()
+                .includes('mental health'),
+              single: {
```
cvalarida, line 208: This also includes using awful names like `single`. 🫠  This will change.

src/templates/layouts/healthCareLocalFacility/HealthServices/ServiceLocation.tsx
cvalarida, line 1: It _is_ worth noting that some of the features on the `content-build` template aren't reproduced here because they're _only_ used for VBA or VHA non-clinical locations. For now, I intend to keep these WET and just copy-pasta them for each different type of facility, but we may decide that's a Bad Idea™ later and add in the extra functionality that isn't used for VAMC. 🤷 

But I figured it's already complicated enough as it is right now. This was an attempt to simplify it just a little bit.

src/types/drupal/paragraph.ts
```diff
@@ -212,13 +212,37 @@ export interface ParagraphRichTextCharLimit1000 extends DrupalParagraph {
 }
 
 export interface ParagraphServiceLocation extends DrupalParagraph {
```
cvalarida, line 214: We should probably rename this...I'm guessing? I don't actually know what all uses it yet, though, so it may not be VAMC-specific. 🤷

src/types/formatted/healthCareLocalFacility.ts
```diff
@@ -43,5 +44,101 @@ export type HealthCareLocalFacility = PublishedEntity & {
   socialLinks: FacilitySocialLinksProps
   lovellVariant?: LovellChildVariant
   lovellSwitchPath?: string
-  healthServices: NodeHealthCareLocalHealthService[] // TODO: Format these
+  healthServices: FormattedVAMCFacilityHealthServices[]
+}
+
+/** Represents the "single" object containing service-related fields. */
+interface VamcFacilityServiceLocation {
```
cvalarida, line 51: Inverse of the above: We may want to rename this to be more generic if we discover it's useful for VHA and VBA facilities.

laflannery:
@cvalarida This all looks pretty good, I think the only comment for now that I think makes sense for this PR is that the `<va-card>` components have `classname` attributes that do not apply, we need to use `class` We are also missing a margin which adds space when there are multiple cards added within one services, such as https://pr1075-hlabd7sgxilymvfbp7qacuxvviemfr3r.tugboat.vfs.va.gov/lovell-federal-health-care-va/locations/captain-james-a-lovell-federal-health-care-center/#addiction-and-substance-use-care

Screenshot 2025-06-06 at 8 08 35 AM

(Obviously if this doesn't make sense to update now ignore me)

src/templates/layouts/healthCareLocalFacility/HealthServices/ServiceLocation.tsx
```diff
@@ -0,0 +1,238 @@
...
+                icon="location_city"
+                size="3"
+              ></va-icon>{' '}
+              {{
+                yes_appointment_only: 'Visit our office, by appointment only',
+                yes_walk_in_visits_only:
+                  'Visit our office, walk-in visits only',
+                yes_with_or_without_appointment:
+                  'Visit our office, with or without an appointment',
+              }[single.fieldOfficeVisits as string] || ''}
```
brianseek, line 60: This is neat.

src/types/drupal/paragraph.ts
brianseek: According to Drupal these all use it

VAMC Facility Health Service, VAMC Facility Non-clinical Service, Service Region, VBA Facility Service

src/templates/layouts/healthCareLocalFacility/HealthServices/HealthServices.tsx
```diff
@@ -0,0 +1,87 @@
...
+              <div id={`${service.entityBundle}-${service.entityId}`}>
+                {service.fieldCommonlyTreatedCondition && (
+                  <p className="vads-u-margin-bottom--2">
+                    Common conditions: {service.fieldCommonlyTreatedCondition}
+                  </p>
+                )}
+
+                {service.fieldTricareDescription ? (
+                  <div>{service.fieldTricareDescription}</div>
+                ) : service.description ? (
```
brianseek, line 49: `service.description` is an object because it is a wysiwyg field. This should either be selecting the `value` or `processed` values. I think that depends on the value of `format` but I'm not sure, I think we need to look at how other wysiwyg fields are rendered.

brianseek (approved):
I'm approving considering that this is scaffolding. That said, there is one minor issue that I pointed out, but that can get fixed in one of the follow up PRs

src/data/queries/healthCareLocalFacility.ts
```diff
@@ -167,6 +170,73 @@ export const formatter: QueryFormatter<
             getOppositeChildVariant(lovell?.variant)
           )
         : null,
-    healthServices: entity.field_local_health_care_service_, // TODO: Format these
+    healthServices: entity.field_local_health_care_service_
```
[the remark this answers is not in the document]
timcosgrove: Life is beautifully messily joyously complex. 

Just kidding, I feel your pain.

src/templates/layouts/healthCareLocalFacility/HealthServices/HealthServices.tsx
timcosgrove: If it's a wysiwyg field we always use processed; On the other hand, if this is rendering, `description` may not actually be a wysiwyg field.

src/templates/layouts/healthCareLocalFacility/HealthServices/ServiceLocation.tsx
```diff
@@ -0,0 +1,238 @@
+import { ServiceLocationTemplateData } from '@/types/formatted/healthCareLocalFacility'
+import { ServiceAddress } from './ServiceAddress'
+import { PhoneNumber } from '@/templates/common/phoneNumber'
+
+export const ServiceLocation = ({
+  fieldReferralRequired,
+  fieldTelephone,
+  fieldPhoneNumber,
```
timcosgrove, line 8: 🤦🏻

src/types/drupal/field_type.d.ts
```diff
@@ -34,9 +34,13 @@ export interface FieldLink {
   options: unknown[]
 }
 export interface FieldOfficeHours {
+  /** Numeric representation of the day of the week (0 is Sunday...I think...) */
```
timcosgrove, line 37: Thank you for all these.

src/types/formatted/healthCareLocalFacility.ts
```diff
@@ -43,5 +44,101 @@ export type HealthCareLocalFacility = PublishedEntity & {
...
+   * "1" for yes
+   * "2", "not_applicable", or "unknown" to omit the referral section
+   */
+  fieldReferralRequired?:
+    | '0'
+    | '1'
+    | '2'
+    | 'not_applicable'
+    | 'unknown'
+    | string
```
timcosgrove, line 96: If we allow `string` here do we need any of the others?

src/templates/layouts/healthCareLocalFacility/HealthServices/ServiceLocation.tsx
cvalarida: 🫠

src/types/formatted/healthCareLocalFacility.ts
cvalarida: TL;DR: Without `| string`, TS gets angry, but we don't expect it to be anything other than the listed string literals.

So the way I typically use this (and should probably document in a comment) is by making the string literals in the union the _expected_ options, but because we're getting data from Drupal, we have no static type guarantee that the string we receive is _absolutely_ going to be one of these. Once we get runtime data validation through the likes of Zod, we'll be able to guarantee that the strings are one of these options.

That said, since we _are_ doing `as Whatever` type assertions when we grab the data from either the mock data or the query to Drupal, this is maybe unnecessary. 🤷 🙈

PR #1084, opened 2025-06-11
Migrate VAMC system page layout to v3 grid

pwolfert:
# Description

I copied the implementation from the [new leadership listing page](https://github.com/department-of-veterans-affairs/next-build/pull/1072/files#diff-19b239ea5d9ee0ffd930cba72dfa39ba2e499646f9c2ddd439babb72aa7e6078). The main difference I'm seeing with the new layout and the old (and it's the same on that leadership listing page) is that the left edge of the nav panel now aligns with the left edge of the breadcrumbs. This is how it works in production, except that both of those things are aligned with the edge of the horizontal rule in the header instead of the header's content. I think this was a purposeful change in direction we made a few weeks back, though I don't remember who decided on it. See my screenshots below.

Note that it's easier to see the real code changes [when whitespace changes are hidden](https://github.com/department-of-veterans-affairs/next-build/pull/1084/files?diff=unified&w=1).

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21432

## Testing Steps

To see the leadership page for a layout comparison, go to http://localhost:3999/greater-los-angeles-health-care/about-us/leadership/.

To see a VAMC System page, go to http://localhost:3999/providence-health-care/

## Screenshots

Before:
Screenshot 2025-06-10 at 6 41 30 PM

After:
Screenshot 2025-06-10 at 6 41 10 PM

Annotated "after":
Screenshot 2025-06-10 at 6 41 10 PM copy

I think possibly some more refinement is needed, but we should probably make that effort for all pages at once so everything's consistent.

Note also that the screenshots in https://github.com/department-of-veterans-affairs/next-build/pull/1072 might be out of date, because the alignment for that page looks more like my screenshots when I go to it now.

### DIFF
--- a/src/templates/components/eventTeaser/index.tsx
+++ b/src/templates/components/eventTeaser/index.tsx
@@ -25,9 +25,9 @@ export const EventTeaser = ({
     <div
       data-template="teasers/event"
       data-featured={fieldFeatured}
-      className="vads-u-margin-bottom--3 medium-screen:vads-u-margin-bottom--4"
+      className="vads-u-margin-bottom--3 tablet:vads-u-margin-bottom--4"
     >
-      <h3 className="vads-u-margin-top--0 vads-u-margin-bottom-1 vads-u-font-size--md medium-screen:vads-u-font-size--lg">
+      <h3 className="vads-u-margin-top--0 vads-u-margin-bottom-1 vads-u-font-size--md tablet:vads-u-font-size--lg">
         <va-link text={title} href={entityUrl.path} />
       </h3>
       <p className="vads-u-margin-bottom--1p5 vads-u-margin-top--0">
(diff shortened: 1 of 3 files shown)


brianseek (changes requested):
related kinda to v3 we have been updating all `medium-screen` breakpoint classes to `tablet`

laflannery:
- On the Top Task links, there is still `usa-grid usa-grid-full`. However, I'm not sure these need to be replaced with v3 class, I think we can probably just remove these entirely
Screenshot 2025-06-11 at 12 05 14 PM

- The Locations section also still has some v1 width classes still present that should get replaced with v3 classes
Screenshot 2025-06-11 at 12 03 46 PM

brianseek (approved):
LGTM!

laflannery:
@pwolfert I'm still seeing quite a few `medium-screen` breakpoints:
- In the Locations section
- In Manage your health online
- In Events
Are you able to update those in this PR?

PR #1115, opened 2025-06-24
Cvalarida/use debug for logging

cvalarida:
# Description

I wasn't seeing the kind of information I wanted to see in GitHub actions, os I spent some time working on the logging situation. Turns out we already had [`debug`](https://www.npmjs.com/package/debug) in our dependencies, so I set it up such that all `console.log`s (and `.error` and `.warn`) instead use `debug`, then added the `DEBUG` env var to the `.env.*` files.

> [!important]
> To see any logs, you'll need to add the `DEBUG` environment variable to your `.env.local`. You can copy it from `.env.example` or use `DEBUG=next-build:*,proxy-fetcher:error` to get basically what was there before.

## Screenshots

image

image
image


envs/.env.dev
```diff
@@ -10,6 +10,11 @@ NEXT_PUBLIC_BUILD_TYPE=vagovdev
 # Location to load vets-website assets
 NEXT_PUBLIC_ASSETS_URL=https://dev-va-gov-assets.s3-us-gov-west-1.amazonaws.com/generated/
 
+# Which loggers to show messages from in the console.
+# https://www.npmjs.com/package/debug
+DEBUG='*,-eslint,-eslint*,-typescript*,-next:jsconfig-paths-plugin,-socks-proxy-agent'
+DEBUG_COLORS=true
```
cvalarida, line 16: This maybe shouldn't be `true`? 🤔 🤷

I've set it to `true` in the other environments because the output should be a TTY, but since most of the work happens in a child worker, by default, it doesn't use colors. See https://www.npmjs.com/package/debug#usage-in-child-processes. I'm not sure how this will affect the output in CI, or what to do about that if it's ugly. 🙈

packages/env-loader/src/cli-options.ts
```diff
@@ -66,8 +66,6 @@ export const getCliOptionsAndArgs = (): CliOptionsAndArgs => {
 
   program
     .passThroughOptions()
-    .option('--DEBUG', 'Verbose logging on build', false)
-    .option('--no-DEBUG', 'No verbose logging on build')
```
cvalarida, line 70: This was used really only to show all `proxy-fetcher` failures, but to do that, it was overriding the `DEBUG` environment variable, which [`debug`](https://www.npmjs.com/package/debug) uses.

packages/proxy-fetcher/src/index.ts
```diff
@@ -82,25 +82,16 @@ export const getFetcher = (
...
-        }
+        const logOrError = attempt <= retryCount ? log : error
+        logOrError(
+          `Failed request (Attempt ${attempt} of ${retryCount + 1}): %o`,
+          {
+            url: response.url,
+            status: response.status,
+            statusText: response.statusText,
+          }
+        )
```
cvalarida, line 94: Looks like this now:
image

If you don't want to see all the logs for all the failed attempts, you can add `-proxy-fetcher:log` to the `DEBUG` env var.

It's kind of a bummer that I get a `+0ms` on that last error instead of `+16s`, but 🤷 .

pwolfert (approved):
Oh wow, I like these logs. I thought the `send` logs were rather verbose, so I just added a `,-send` to the config. I think it would be to have some kind of documentation on how to do that—how to fine-tune which modules we show in the logs—because that's a handy feature.

PR #1117, opened 2025-06-26
Send an event via `recordEvents` like the original

pwolfert:
# Description

The "Other services" links were missing custom analytics events that are present in production. The links in [the original template](https://github.com/department-of-veterans-affairs/content-build/blob/0a5b0ab4db49c212b77ff2e54d8da78b29fcf55d/src/site/paragraphs/facilities/list_of_link_teasers_facility.drupal.liquid#L54) were calling a `recordEvent` function, to which we already have an analog in next-build. I've updated the component for this section to call it with the same parameters.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21522

## Testing Steps

I'm not sure if there's a better way to test the analytics events, but I was adding a console log to [the event function](https://github.com/department-of-veterans-affairs/next-build/blob/pwolfert/related-links-event/src/lib/analytics/index.tsx#L28) and setting my browser console to preserve the log between refreshes. Then you click on one of the links in the "Other services" section of a facility or region page. Here are a links to one of each:

- http://localhost:3999/new-york-harbor-health-care/locations/margaret-cochran-corbin-va-campus/
- http://localhost:3999/new-york-harbor-health-care/

### DIFF
--- a/src/templates/common/relatedLinks/index.test.tsx
+++ b/src/templates/common/relatedLinks/index.test.tsx
@@ -1,6 +1,13 @@
 import { render } from '@testing-library/react'
+import { fireEvent } from '@testing-library/dom'
 import { RelatedLinks } from '.'
 import { FormattedRelatedLinks } from '@/types/formatted/relatedLinks'
+import { recordEvent } from '@/lib/analytics/recordEvent'
+
+// Mock recordEvent
+jest.mock('@/lib/analytics/recordEvent', () => ({
+  recordEvent: jest.fn(),
+}))
 
 describe('RelatedLinks Component', () => {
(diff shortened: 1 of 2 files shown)


laflannery:
@pwolfert In case you are curious or if it's helpful, I use [this extension for testing analytics events](https://chromewebstore.google.com/detail/debugger-for-google-analy/nomnoimacbncclfbnfaingniikblfbji?hl=en-US) and compare the events to what is in Prod, that's how I got the screenshot in the original ticket

pwolfert:
That's helpful, Laura. Thanks!

PR #1122, opened 2025-06-27
Fix screen readers breaking up the "Get updates from..." text

pwolfert:
# Description

React was treating these as distinct text nodes and was even putting an HMTL comment between them. If I concatenate the string before inserting it, React treats it as one thing.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21524


brianseek (approved):
very nice

PR #1137, opened 2025-07-02
Also fetch bifurcated Lovell stories and events for Lovell VAMC system pages

pwolfert:
# Description

Lovell VA and TRICARE system pages were supposed to be showing featured stories that belonged to the VA _federal_ (parent/shared/bifurcated) administration as well as their own variant-specific stories. @omahane found [some functions in the old content-build project](https://github.com/department-of-veterans-affairs/content-build/blob/main/src/site/stages/build/drupal/process-lovell-pages.js#L189) that did similar things, and I noticed that they were treating events similarly. It made sense to me to apply this same logic to the events as well, so I did. I'll include screenshots below showing that we were previously showing a fallback event on the Lovell VA system page but now we get an actual featured event from the parent Lovell page now that we're querying for it.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21521

## Testing Steps

The following pages should now show stories and events that they weren't previously showing: 

- http://localhost:3999/lovell-federal-health-care-va/
- http://localhost:3999/lovell-federal-health-care-tricare/

## Screenshots

Before (one distant-future fallback event, no stories):

Screenshot 2025-07-02 at 12 28 35 PM

After (one bifurcated Lovell event, two bifurcated Lovell featured stories):

Screenshot 2025-07-02 at 12 28 51 PM

### DIFF
--- a/src/data/queries/tests/vamcSystem.test.tsx
+++ b/src/data/queries/tests/vamcSystem.test.tsx
@@ -10,12 +10,37 @@ import mockEventData from '@/products/event/mock.json'
 import { RESOURCE_TYPES } from '@/lib/constants/resourceTypes'
 import { params } from '../vamcSystem'
 import { DrupalJsonApiParams } from 'drupal-jsonapi-params'
+import { LOVELL } from '@/lib/drupal/lovell/constants'
+import { ExpandedStaticPropsContext } from '@/lib/drupal/staticProps'
 
 const mockFeaturedEventData = {
   ...mockEventData,
   // Just to differentiate in the snapshot
   field_featured: true,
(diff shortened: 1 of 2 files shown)


cvalarida (approved):
I think it's good? 🤔 If I'm honest (like I mentioned in my one comment), I'm having a hard time wrapping my brain around it, but if @laflannery says the output is what we want, then the code looks as clean as we're liable to get it, which is about all I can speak to. 😅

laflannery:
Approved! This looks good, I used the fancy new test Tugboat system to run through a bunch of scenarios.

PR #1139, opened 2025-07-03
Fix icon alignment in location services

pwolfert:
# Description

There was some whitespace inconsistency causing a difference in the gap between the icon and the text. This is one way to make it consistent. We could also get rid of the extra space on all of them. However, I don't like how close the icons are when I remove the space. It looks really crowded.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21608

## Testing Steps

Go to http://localhost:3999/central-texas-health-care/locations/olin-e-teague-veterans-center/#advice-nurse and take a look at the three icons for "Visit our office...", "Virtual visits...", and "A referall...". In the fixed version, the space between the icons and text should be consistent across the three rows.

## Screenshots

Before:
Screenshot 2025-07-03 at 2 16 47 PM

After:
Screenshot 2025-07-03 at 2 13 06 PM

With annotation to see the alignment:
alignment

### DIFF
--- a/src/templates/layouts/healthCareLocalFacility/HealthServices/ServiceLocation.tsx
+++ b/src/templates/layouts/healthCareLocalFacility/HealthServices/ServiceLocation.tsx
@@ -98,7 +98,7 @@ export const ServiceLocation = ({
                 icon={fieldReferralRequired === '1' ? 'check_circle' : 'cancel'}
                 size="3"
                 data-testid="referral-icon"
-              ></va-icon>
+              ></va-icon>{' '}
               {fieldReferralRequired === '1'
                 ? 'A referral is required'
                 : 'A referral is not required'}


src/templates/layouts/healthCareLocalFacility/HealthServices/ServiceLocation.tsx
```diff
@@ -98,7 +98,7 @@ export const ServiceLocation = ({
                 icon={fieldReferralRequired === '1' ? 'check_circle' : 'cancel'}
                 size="3"
                 data-testid="referral-icon"
-              ></va-icon>
+              ></va-icon>{' '}
```
brianseek, line 101: 😆

brianseek (approved):
🚀

PR #1140, opened 2025-07-03
Vets Center: Correct defects in "Also called" subtitle

pwolfert:
# Description

These changes correct some small defects in the "Also called" subtitle of the Vets Center pages, which shows the official name of the vets center as opposed to its common name.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21554

## Testing Steps

- Visit http://localhost:3999/st-paul-vet-center/ and check the page title and the `<p>` element below it. Should have an `aria-describedby` on the title that points to the subtitle element. 
- If you go to http://localhost:3999/boston-vet-center/, you should find that the title doesn't have an `aria-describeby` attribute, because there's nothing to point to.

Also check the ticket for details about intended behavior.

## Screenshots

Before:
Screenshot 2025-07-03 at 3 07 04 PM

After:
Screenshot 2025-07-03 at 3 06 51 PM

### DIFF
--- a/src/templates/layouts/vetCenter/index.test.tsx
+++ b/src/templates/layouts/vetCenter/index.test.tsx
@@ -220,4 +220,97 @@ describe('VetCenter with valid data', () => {
     expect(screen.queryByText(/1010 Delafield Road/)).toBeInTheDocument()
     expect(screen.queryByText(/In the spotlight/)).toBeInTheDocument()
   })
+
+  describe('Also called functionality', () => {
+    test('renders "Also called" text when officialName is different from title', () => {
+      const dataWithDifferentOfficialName = {
+        ...mockData,
+        title: 'Pittsburgh Vet Center',
+        officialName: 'Pittsburgh Veterans Center for Readjustment',
+      }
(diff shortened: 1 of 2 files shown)


src/templates/layouts/vetCenter/index.test.tsx
```diff
@@ -220,4 +220,97 @@ describe('VetCenter with valid data', () => {
...
+
+      // Should not render the "Also called" text
+      expect(screen.queryByText(/Also called the/)).not.toBeInTheDocument()
+
+      // h1 should not have aria-describedby attribute
+      const heading = screen.getByRole('heading', { level: 1 })
+      expect(heading).not.toHaveAttribute('aria-describedby')
+    })
+
+    test('handles case where title is null but officialName exists', () => {
```
cvalarida, line 303:
```suggestion
    test('does not render "Also called" when title is null but officialName exists', () => {
```

To keep consistency with the other tests.

cvalarida (approved):

PR #1143, opened 2025-07-04
Vets Center: Use standard `PhoneNumber` component

pwolfert:
# Description

The old code was just using an anchor tag with the `tel:` prefix for the number. We have a more sophisticated component for this now, so we're going to use it. Should satisfy all the requirements of the ticket. (Also changes the label and doesn't use a heading so we're more consistent with the other places where phone numbers are displayed on the site.)

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21558

## Testing Steps

Go to http://localhost:3999/boston-vet-center/ and look at the locations and contact info section.

### Before:

Screenshot 2025-07-04 at 11 34 12 AM

### After:

Screenshot 2025-07-04 at 11 37 48 AM


laflannery (approved):
I think the margin, especially between the address and the phone might need to be adjusted but it's hard to tell until everything is merged together. I can make a new ticket later if that ends up needing to happen

PR #1151, opened 2025-07-07
VADS 3458: remove right margin from feedback button

jamigibbs:
# Description

The `va-button` web component uses USWDS styles which has a right margin. When the button is used in a way that makes it right aligned to content (like the feedback button), this right margin should be removed. For those special scenarios, the web component has a `part` attribute that can be styled through the shadow dom. That is what this PR is doing for the feedback button.

Here is the feedback button in question: 

https://github.com/department-of-veterans-affairs/next-build/blob/ec31c5b604e7d5c2d35330967d66c957fc0dd047/src/templates/common/contentFooter/index.tsx#L47-L61

## Ticket

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3458

## Developer Task

```md
```

## Testing Steps

## QA steps

- Review the feedback button at the bottom of the content on this page: http://localhost:3999/boston-health-care/events/

## Screenshots

**Before**: 

before

**After**:

after

---

### Reviewing a PR

## Standard Checks

## Merging a Layout

When merging a layout, you must ensure that the content type has been turned on for `next-build` in the [.tugboat.env](../envs/.tugboat.env). This method mocks the CMS flag that must be turned on for a layout to be included in the build.

The layout component and matching resource type should be included in the [slug.tsx](../src/pages/[[...slug]].tsx), so that it can reviewed. Including a component in the slug.tsx does not mean a page will be viewable in production only on the tugboat for the branch.

When a layout is merged to main and approved for deployment, the prod CMS will turn the toggle on for the resource type. 

The status of layouts should be kept up to date inside [templates.md](../READMEs/templates.md). This includes QA progress, development progress, etc. A link should be provided for where testing can occur.

### DIFF
--- a/src/assets/styles/globals.css
+++ b/src/assets/styles/globals.css
@@ -20,6 +20,10 @@ body {
   padding-top: 0.78125rem;
 }
 
+va-button#mdFormButton::part(button) {
+  margin-right: 0px;
+}
+
 .feedback-button {
   background-color: var(color-primary);
   font-size: 1rem;


jamigibbs:
> now it seems almost not aligned when the Back to Top button is present

@laflannery Thanks for taking a look! I think that _might_ be an illusion? For comparison, here is a page on VA.gov that already has the button right aligned (this is the non-web compnent version): https://www.va.gov/education/about-gi-bill-benefits/


I took a screenshot of before and after and when I compare them on top of each other, only the feedback button moves over. It helps to visually see it when I hover over the back to top so it has a background color.

https://github.com/user-attachments/assets/11a167ab-9370-4880-a927-ad328daf4248

laflannery:
Gotcha, it definitely still looks weird to me but I think it's one of those things where I've just stared at it for too long at this point :laughing:

src/assets/styles/globals.css
```diff
@@ -20,6 +20,10 @@ body {
   padding-top: 0.78125rem;
 }
 
+va-button#mdFormButton::part(button) {
+  margin-right: 0px;
+}
+
```
cvalarida, line 26: I think this is fine, but is there any way we can make this change outside of `globals.css`? I think the part that I'm confused about is why we need it in the first place.

If handling this upstream is impractical for reasons I don't know, can you add a little comment explaining what this does, what it targets, and why we're handling it here instead of elsewhere?

cvalarida (approved):

src/assets/styles/globals.css
cvalarida: > I think the part that I'm confused about is why we need it in the first place.

To be clear, what I mean here is why we need to target this `#mdFormButton` by ID instead of a more general rule in the design system like `va-button.vads-align-right` or something. Put another way, I think what I'm asking is, "Will this be useful as part of the design system? Are there any other right-aligned buttons anywhere that would need to reproduce this rule?" It _seems_ like making an exception to the DS here is introducing a code smell 🙊 , but I don't have the broader context to say for sure right now.

jamigibbs: >  is there any way we can make this change outside of globals.css

@cvalarida Good questions! I think what you're asking is why it's not being done in the component itself? The `va-button` web component uses [USWDS styles directly](https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/components/va-button/va-button.scss#L4) which is providing us with the right margin upstream. We generally don't prefer to override USWDS styles in the component itself unless we're deviating pretty significantly from them for reasons that are specific to the VA (like adding our own features).

In situations where there are minor styling adjustments needed on a case-by-case basis (like here when the button is fully right aligned to the content), we provide a `part` attribute on elements for teams to handle any minor styling adjustments themselves. This provides flexibility while also keeping the complexity of the component in check and also keeping us from deviating from USWDS at the component's core.

**_But_** if we find that we need to apply this override style in many places, I think it would make sense to update the component itself.

jamigibbs: I double checked vets-website to see if this was being overridden in other places but none of these showed the margin being removed.

Screenshot 2025-07-07 at 2 11 52 PM

jamigibbs: > To be clear, what I mean here is why we need to target this `#mdFormButton` by ID instead of a more general rule in the design system like `va-button.vads-align-right` or something. 

@cvalarida I just saw your second comment after I already posted so I missed a little bit of context after I already replied. 🫣

Yes, I agree it does feel suspicious. Unless there are multiple cases that the USWDS right margin is needing to be overridden, I'm of the opinion that custom style should be handled on an individual bases (not in the component). A utility class like this probably won't work in this case `va-button.vads-align-right` because the shadow dom won't allow styling through the web component element. We would likely need to add an entirely new prop to the component.

But I fully agree that if we are seeing several instances of the right margin needing to be overridden, the component should handle that.

jamigibbs: > So this is the only time that you know of where a va-button needs to be right-aligned, then?

@cvalarida I believe so, yes. It kind of feels like it would be common for RTL content languages so maybe if VA.gov supports Arabic or some other RTL language in the future it would be in lots of places? 🤔  

And sorry, no catalogue unfortunately but ideally what would happen is that it would be a feature that USWDS would be getting requests for and we would be able to consume an update that would handle it from the upstream package. That would appear in their release notes which VADS would account for during the update process.

PR #1153, opened 2025-07-08
Vets Center: Operating status flags

pwolfert:
# Description

Ported [the old operating status template](https://github.com/department-of-veterans-affairs/content-build/blob/main/src/site/includes/operating_status_expandable_alert.liquid) and modeled the new one after [the facility version](https://github.com/department-of-veterans-affairs/next-build/blob/db846cbc427efe5a32f23d99bc455aad8156a12f/src/templates/layouts/healthCareLocalFacility/OperatingStatus.tsx).

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21568

## Testing Steps

- http://localhost:3999/greensboro-vet-center/ has a "Faciility notice" under the "main location" heading
- http://localhost:3999/yakima-valley-vet-center/ has a "Virtual care only" status
- http://localhost:3999/columbus-oh-vet-center/ has a temporary closure

## Screenshots

Screenshot 2025-07-07 at 10 26 05 PM

### DIFF
--- a/src/lib/utils/getHtmlFromDrupalContent.ts
+++ b/src/lib/utils/getHtmlFromDrupalContent.ts
@@ -1,13 +1,24 @@
 import { drupalToVaPath, newlinesToBr } from './helpers'
 import { createPhoneLinks } from './createPhoneLinks'
 
+export type GetHtmlFromDrupalContentOptions = {
+  convertNewlines?: boolean
+}
+
 /**
  * Extract and transform HTML content from Drupal with various filters applied.
  *
  * Return the HTML string after processing.
  */
-export const getHtmlFromDrupalContent = (content: string): string => {
-  return [createPhoneLinks, drupalToVaPath, newlinesToBr].reduce(
+export const getHtmlFromDrupalContent = (
+  content: string,
+  options: GetHtmlFromDrupalContentOptions = {}
+): string => {
+  const transformers = [createPhoneLinks, drupalToVaPath]
+  if (options.convertNewlines) {
+    transformers.push(newlinesToBr)
+  }
+  return transformers.reduce(
     (contentToTransform, transformFn) => transformFn(contentToTransform),
     content
   )
--- a/src/lib/utils/getHtmlFromDrupalContent.ts
+++ b/src/lib/utils/getHtmlFromDrupalContent.ts
@@ -1,0 +1,14 @@
+import { drupalToVaPath, newlinesToBr } from './helpers'
+import { createPhoneLinks } from './createPhoneLinks'
+
+/**
+ * Extract and transform HTML content from Drupal with various filters applied.
+ *
+ * Return the HTML string after processing.
+ */
+export const getHtmlFromDrupalContent = (content: string): string => {
+  return [createPhoneLinks, drupalToVaPath, newlinesToBr].reduce(
+    (contentToTransform, transformFn) => transformFn(contentToTransform),
+    content
+  )
+}
--- a/src/lib/utils/helpers.ts
+++ b/src/lib/utils/helpers.ts
@@ -154,3 +154,10 @@ export const numToWord = (num) => {
       (num % 10 ? '-' + belowTwenty[(num % 10) - 1] : '')
     )
 }
+
+/**
+ * Convert newline characters to HTML <br> tags.
+ */
+export const newlinesToBr = (content: string): string => {
+  return content.replace(/\n/g, '<br>')
+}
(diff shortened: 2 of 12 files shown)


pwolfert:
Hmm, the problem I was experiencing during local development doesn't appear to be happening in my browser when viewing the Tugboat deployment. Here's a video of what's happening locally:

https://github.com/user-attachments/assets/6bcaeebd-1118-4760-8ded-78c874fb13c8

I have found no evidence that anything external to the web component that is under our control is contributing to it. I think it's just a timing issue. When I have dev tools open in Chrome, it works fine. And when I view it in Safari, it works fine. I think maybe it's not a real issue that we can solve, so I'll move on.

laflannery:
@pwolfert Just one comment:
- Phone numbers and links need to be processes so that they are clickable on the FE. Some examples where you can see this:
  - [Naples Vet center](https://pr1153-r6cysl4wmjpzyhk5p2ycnkqij3lhmlvm.tugboat.vfs.va.gov/naples-vet-center/)
  - [Babylon Vet center](https://pr1153-r6cysl4wmjpzyhk5p2ycnkqij3lhmlvm.tugboat.vfs.va.gov/babylon-vet-center/)
  - [Sanford Vet center](https://pr1153-r6cysl4wmjpzyhk5p2ycnkqij3lhmlvm.tugboat.vfs.va.gov/sanford-vet-center/)
Screenshot 2025-07-08 at 8 49 25 AM

pwolfert:
Doh! I remember seeing that in the old template and forgot to go back after getting distracted by the drop-down issue. Thanks for the links to examples!

laflannery (approved):
Looks good!

pwolfert:
Turns out that particular page (that particular vets center) has hydration errors on the main branch as well. And it happens whether I use `dangerouslySetInnerHTML` or not, so it's unrelated to my changes. I'll have to work through it as I continue to work on the Vets Center pages.

src/lib/utils/helpers.ts
```diff
@@ -154,3 +154,10 @@ export const numToWord = (num) => {
       (num % 10 ? '-' + belowTwenty[(num % 10) - 1] : '')
     )
 }
+
+/**
+ * Convert newline characters to HTML <br> tags.
+ */
+export const newlinesToBr = (content: string): string => {
+  return content.replace(/\n/g, '<br>')
```
cvalarida, line 162: Idle curiosity: Should this not be a self-closing tag? I remember `<br>` being weird. 🤔

src/templates/layouts/vetCenter/ExpandableOperatingStatus.tsx
```diff
@@ -0,0 +1,63 @@
...
+ * Returns null if the operating status is 'normal' or invalid.
+ */
+export const ExpandableOperatingStatus = ({
+  operatingStatusFlag,
+  operatingStatusMoreInfo,
+}: ExpandableOperatingStatusProps) => {
+  const statusAlert = statusAlerts[operatingStatusFlag]
+  if (statusAlert) {
+    const { status, text } = statusAlert
+    const style = { maxWidth: '303px' }
```
cvalarida, line 35: Oh curious. Is there a design token we can use for this value, perchance? 🤔

cvalarida (approved):
Looks solid. There's an opportunity for consolidation, but otherwise, it looks good to me.

src/lib/utils/getHtmlFromDrupalContent.ts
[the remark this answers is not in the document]
pwolfert: I wasn't sure if possibly the `<br>` conversion shouldn't be part of the other one, but I can try to trace down the original templates. Or if you or @laflannery think that it's fine to convert newline characters from all WYSIWYG blocks and health services descriptions to HTML line breaks, I can just do it.

src/lib/utils/helpers.ts
pwolfert: You might then ask, as my brain did, whether since we're feeding this HTML to React React will be unhappy. It doesn't seem to be though.

src/templates/layouts/vetCenter/ExpandableOperatingStatus.tsx
pwolfert: Hmm, I was just copying it verbatim, but there's probably a close-enough token value we could use

src/lib/utils/getHtmlFromDrupalContent.ts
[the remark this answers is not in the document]
pwolfert: It looks like the `newline_to_br` filter that I was translating here is only used in the meta tags for the page and two different liquid templates that basically implement the same expandable alert logic. It may not hurt, though, to add that logic to the other places I mentioned above that currently use `getHtmlFromField`

PR #1158, opened 2025-07-10
Vets Center: Clean up locations section

pwolfert:
# Description

Does a lot of cleaning up of the headings in this section and standardizes the phone number display. See ticket for more details.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21589

## Testing Steps

http://localhost:3999/columbus-oh-vet-center/

## Screenshots

Before:

localhost_3999_columbus-oh-vet-center_ (1)

After:

localhost_3999_columbus-oh-vet-center_

Note that the first paragraph of the hours section has a paragraph in it with default styling. This paragraph comes from Drupal content, so I don't have direct control over whether it has utility classes on it. While I could write a CSS rule that overrides its styles, that seemed like it might be overkill. Curious what people think about that. I know we try to avoid creating new custom CSS rules.

### DIFF
--- a/src/templates/layouts/vetCenter/index.tsx
+++ b/src/templates/layouts/vetCenter/index.tsx
@@ -261,7 +261,7 @@ export function VetCenter({
 
           {/* Other locations */}
           <div className="vads-u-margin-bottom--3">
-            <h2>Other locations</h2>
+            <h2 id="other-locations">Other locations</h2>
             <p>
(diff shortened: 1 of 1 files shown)


laflannery:
@pwolfert I forgot one small thing - now that we changed "Other locations" to an `<h2>`, it needs an ID so that the "On this page" component will pick it up and display and anchor link for it. Are you able to add that?

PR #1161, opened 2025-07-11
Vets Center: Banner image from centralized content

pwolfert:
# Description

Adds the Vets Center banner image common to all Vets Center pages. There's a reference to a Drupal media item in the `field_vet_center_banner_image` field that comes from _centralized content_. We use that reference to fetch the media directly as a `media--image` from Drupal and give that to the formatter function.

- Adds banner image fetching to the query module
- Adds and updates the TS types accordingly
- Renders the banner image on the Vets Center page component
- Refactors the query module unit tests to favor testing the full integration of data fetching and formatting rather than doing it separately, which I think is redundant.
  - I don't really think it's that valuable to test the formatter function in isolation when we can test the integrated form. We'd end up having to duplicate some tests anyway if we wanted to test all the same aspects of the data function too. And it's actually pretty easy to mock these drupal query functions anyway. Am I wrong?

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21555

## Testing Steps

[Explain the steps needed for testing](http://localhost:3999/greensboro-vet-center/)

## Screenshots

Screenshot 2025-07-11 at 3 07 07 PM

### DIFF
--- a/src/templates/layouts/vetCenter/index.tsx
+++ b/src/templates/layouts/vetCenter/index.tsx
@@ -190,7 +190,7 @@ export function VetCenter({
             <MediaImage
               {...bannerImage}
               imageStyle="7_2_medium_thumbnail"
-              alt=""
+              className="vads-u-padding-y--1p5"
             />
           )}
           {introText && (


pwolfert:
Something really weird is happening with the CI tests. It's failing on old code. You look in the diff and the problems it's complaining about just aren't there. It's like the CI tests are checking an old commit.

pwolfert:
Tugboat build is failing because of some unexpected data (at least not what _I_ expect): The "South Orange County Vet Center" has a `field_vet_center_banner_image` value of `null`. The other ones that I've looked at have that centralized content banner image.

brianseek (approved):
LGTM!

PR #1162, opened 2025-07-11
Vets Center: Transform HTML content for various content fields

pwolfert:
# Description

There were a lot of different content fields that needed to be transformed so they'd pick up on things like phone number.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21464
Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21563

## Testing Steps

- http://localhost:3999/omaha-vet-center/ - "In the spotlight" section
- http://localhost:3999/boston-vet-center/ - "Where can I get help after hours?"
- http://localhost:3999/greensboro-vet-center/#item-suicide-prevention - "Suicide prevention" (https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21563)

## Screenshots
Screenshot 2025-07-11 at 4 39 16 PM

Screenshot 2025-07-11 at 4 19 22 PM

Screenshot 2025-07-11 at 4 38 27 PM

### DIFF
--- a/src/data/queries/vetCenterHealthServices.ts
+++ b/src/data/queries/vetCenterHealthServices.ts
@@ -36,7 +36,6 @@ export const formatter: QueryFormatter<
         getHtmlFromDrupalContent(
           serviceEntity.field_vet_center_service_descrip
         ) || null,
-      description: getHtmlFromField(serviceEntity.description) || null,
       body: getHtmlFromField(item.field_body) || null,
     }
   })
--- a/src/data/queries/vetCenterHealthServices.ts
+++ b/src/data/queries/vetCenterHealthServices.ts
@@ -5,6 +5,8 @@ import {
 } from '@/types/formatted/vetCenterHealthServices'
 import { QueryFormatter, QueryParams } from 'next-drupal-query'
 import { DrupalJsonApiParams } from 'drupal-jsonapi-params'
+import { getHtmlFromField } from '@/lib/utils/getHtmlFromField'
+import { getHtmlFromDrupalContent } from '@/lib/utils/getHtmlFromDrupalContent'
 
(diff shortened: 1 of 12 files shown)


src/data/queries/vetCenterHealthServices.ts
```diff
@@ -31,9 +33,11 @@ export const formatter: QueryFormatter<
       commonlyTreatedCondition:
         serviceEntity.field_commonly_treated_condition || null,
       vetCenterServiceDescription:
-        serviceEntity.field_vet_center_service_descrip || null,
-      description: serviceEntity.description?.processed || null,
-      body: item.field_body?.processed || null,
+        getHtmlFromDrupalContent(
+          serviceEntity.field_vet_center_service_descrip
+        ) || null,
+      description: getHtmlFromField(serviceEntity.description) || null,
```
pwolfert, line 39: This `description` field does not appear to be used anywhere. This query formatter is specific to the Vets Center. I'd like to know from someone familiar with the Drupal side of things if these fields' value is coming from the same source (`description` vs `field_vet_center_service_descrip `).

laflannery (approved):
LGTM

src/data/queries/vetCenterHealthServices.ts
brianseek: sorry, I take that back. I see this is an included taxonomy term that does have a description value. It is different data than what is in `field_vet_center_service_descrip` and also its an object, so I think it should be `getHtmlFromField(serviceEntity.description?.processed) || null`.

brianseek: oh, I see its not used in the render...yeah lets kill it then.

PR #1177, opened 2025-07-21
Vets Center: Support full range of nested data in FAQ items

pwolfert:
# Description

- The previous version of the formatter used a naive and simplistic approach to surfacing the content for the _answer_ content for the questions in the FAQ section. It assumed there was always a single paragraph block for each answer and that it was a WYSIWYG paragraph. The section in the ticket's screenshot actually has three paragraph blocks for the answer, and the middle one is an Alert type paragraph. In order to fix this problem, however, I needed to normalize the data being fetched for _centralized content_ so the existing _Drupal paragraph_ infrastructure could read that data. Some details:
    - The data is being fetched by [the entity_field_fetch Drupal plugin](https://www.drupal.org/project/entity_field_fetch) (thanks to @timcosgrove for [getting to the bottom of that](https://dsva.slack.com/archives/C01SR56755H/p1752528448420559))
    - The data is being transformed in strange but mostly consistent ways
    - My normalization function transforms it back into something we'd expect to see from the base Drupal API
    - We can then use the existing _paragraph_ formatters and UI components, which are fairly extensive
    - One caveat is that the data that passes through that plugin no longer has UUIDs, which some of the formatters and UI components expect, so I had to specify a fallback in some places.
- Once the right data was available to the frontend, I discovered that the existing component for the expandable-text-type paragraphs used the "wrong" web component—a different one than the old liquid templates (now using [va-additional-info](https://design.va.gov/components/additional-info)). However, changing this **has the potential to break other pages**. I do not really have a way of knowing what other pages use this Drupal paragraph type and therefore which pages will be affected.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21564

## Testing Steps

On http://localhost:3999/secaucus-vet-center/, scroll down to the "Am I eligible for Vet Center services as a Veteran or service member?" accordion item.

Screenshot 2025-07-21 at 2 30 15 PM

Without this fix, the info alert in this content is not shown, and neither is the paragraph after it.

## Screenshots

Before:

Screenshot 2025-07-21 at 2 32 47 PM

After:

Screenshot 2025-07-21 at 2 33 27 PM

### DIFF
--- a/additional.d.ts
+++ b/additional.d.ts
@@ -24,6 +24,7 @@ declare namespace JSX {
   interface IntrinsicElements {
     'va-accordion'
     'va-accordion-item'
+    'va-additional-info'
     'va-alert'
     'va-alert-expandable'
     'va-back-to-top'
--- a/src/data/queries/alert.ts
+++ b/src/data/queries/alert.ts
@@ -9,8 +9,8 @@ export const formatter: QueryFormatter<ParagraphAlert, Alert> = (
 ) => {
   return {
     type: entity.type as Alert['type'],
-    id: entity.id,
-    entityId: entity.drupal_internal__id,
+    id: entity.id ?? null,
+    entityId: entity.drupal_internal__id ?? null,
     alertType: entity.field_alert_type as AlertType,
     heading: entity.field_alert_heading,
     blockReference: queries.formatData(
--- a/src/data/queries/alertBlock.ts
+++ b/src/data/queries/alertBlock.ts
@@ -12,7 +12,6 @@ export const formatter: QueryFormatter<BlockAlert, AlertBlock> = (
   if (!entity) {
     return null
   }
-  console.log("block alert formatter", entity)
 
   return {
     alertType: (entity.field_alert_type === 'information'
--- a/src/data/queries/alertBlock.ts
+++ b/src/data/queries/alertBlock.ts
@@ -12,6 +12,7 @@ export const formatter: QueryFormatter<BlockAlert, AlertBlock> = (
   if (!entity) {
     return null
   }
+  console.log("block alert formatter", entity)
 
   return {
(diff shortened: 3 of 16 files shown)


laflannery:
@pwolfert On the alert component, we need to remove a couple properties:
- `uswds="false"`: I'm not actually sure this is doing anything detrimental because this isn't actually allowed anymore since all V1 (non-uswds) components have been deprecated but we should remove it anyway just in case
- `role="alert"`: Using this will make the alert intrusively announce for folks using screen readers and should only be used for very important content that, if missed, would be detrimental to the user. ([More info on roles and when to use them if you are curious](https://design.va.gov/components/alert/alert-expandable/#assign-an-appropriate-aria-role)) . For this content, that is not the case, a role is not required and we should remove it

pwolfert:
@laflannery, I found what you're referring to, and it's the `<va-alert data-template="paragraphs/alert"...`, right? That part I didn't add. That's how all Drupal alert paragraphs are rendered on the site. Would you like me to update the template/component for all Drupal Alert paragraphs?

What I changed was the `<va-additional-info disable-border="true" uswds="true"...` inside it. Those two attributes are coming from the content-build template. Are those ones correct?

pwolfert:
@laflannery, I'm removing `role="alert"` on both the alerts on the page. I was [reading the VADS guidance on it](https://design.va.gov/components/alert/#assign-an-appropriate-aria-role), and I think it's unlikely that any of the alerts rendered by the two shared components that govern the ones we see on this page will fall under the category of not being present on the page at the start. If we ever do come across one of those, we can add it for that specific alert.

brianseek (approved):
Lots going on here. I'm pretty sure I followed it all and it seems all well and good to me. More importantly, the thing is now doing what we expect. Nice job! This seems like it was a headache.

PR #1179, opened 2025-07-22
Upgrade next-drupal dependency to 2.0.0

cvalarida:
# Description

Like it says on the tin, this PR upgrades `next-drupal` to 2.0.0. See the [migration guide](https://next-drupal.org/docs/upgrade-guide) for details, but the TL;DR is that we didn't need to do much because we're not utilizing any of the features that experienced breaking changes.

That said, while I checked that it worked locally, this PR will do a static build that we can use to verify that it works correctly there too. I don't anticipate any problems, though.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21374


src/lib/drupal/drupalClient.ts
```diff
@@ -20,7 +20,8 @@ export const drupalClient = new DrupalClient(baseUrl, {
     process.env.USE_REDIS === 'true'
       ? redisCache(createRedisClient(process.env.REDIS_URL))
       : null,
-  previewSecret: process.env.DRUPAL_PREVIEW_SECRET,
+  // I think this is deprecated in next-drupal v2.0.0
+  // previewSecret: process.env.DRUPAL_PREVIEW_SECRET,
```
cvalarida, line 24: @timcosgrove Any idea how this was used before? From what I could tell, `next-drupal` [still uses `DRUPAL_PREVIEW_SECRET`](https://github.com/search?q=repo%3Achapter-three%2Fnext-drupal%20DRUPAL_PREVIEW_SECRET&type=code), but I'm not really sure how to test it out.

brianseek: I think we need to deploy this next branch on a Drupal tugboat to test the preview. I can test that.

src/lib/drupal/drupalClient.ts
```diff
@@ -18,7 +18,8 @@ export const drupalClient = new DrupalClient(baseUrl, {
...
```
timcosgrove, line 22:
```suggestion
```

PR #1180, opened 2025-07-23
Vets Center: Fix the hydration error by fixing some invalid HTML

pwolfert:
# Description

We can just use the `Address` component here because it already does everything we want—and with valid HTML. The `<p>` tags surrounding the `<div>` tags was invalid HTML, which is what was causing the hydration errors.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21739

## Testing Steps

Loading http://localhost:3999/secaucus-vet-center/ should show no Next.js hydration errors (a big pop-up dialog, see ticket).


brianseek (approved):
Nice!

PR #1186, opened 2025-07-25
Componentize the text-with-image layout pattern and use it

pwolfert:
# Description

## Summary

This PR introduces a new reusable `TextWithImage` component that standardizes the common pattern of displaying text content on the left and an image on the right (desktop) or stacked vertically (mobile). We then refactored multiple existing components to use this new pattern, improving code consistency and maintainability.

### New Component: `TextWithImage`

- **Location**: `src/templates/components/textWithImage/index.tsx`
- **Features**:
  - Responsive layout: stacked on mobile, side-by-side on desktop
  - Accepts any `ReactNode` for image content
  - Uses `children` prop for text content
  - Supports arbitrary HTML div attributes via `HTMLAttributes<HTMLDivElement>`

### Refactored Components

1. **FacilityListing** (`src/templates/components/facilityListing/index.tsx`)
2. **StoryTeaser** (`src/templates/layouts/vamcSystem/StoryTeaser.tsx`)
3. **Health Care Local Facility Layout** (`src/templates/layouts/healthCareLocalFacility/index.tsx`)
4. **Vet Center Layout** (`src/templates/layouts/vetCenter/index.tsx`)
5. **Vet Center Outstation Layout** (`src/templates/layouts/vetCenterOutstation/index.tsx`)
6. **NewsStoryTeaser** (`src/templates/components/newsStoryTeaser/index.tsx`)

## Benefits

1. **Consistent Layout Pattern**: All components now use the same responsive text/image layout
2. **Reduced Code Complexity**: Eliminated custom grid and flex layouts across multiple components
3. **Better Maintainability**: Changes to the layout pattern only need to be made in one place
4. **Preserved Functionality**: All original behavior, styling, and conditional rendering is maintained
5. **Responsive Design**: All components now benefit from consistent responsive behavior
6. **Reusable Pattern**: The `TextWithImage` component can be used anywhere this layout is needed

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21567


laflannery:
@pwolfert I have questions,

- This doesn't quite work for the Story listings because they have a different layout at the 481 breakpoint
- However, I am confused because I'm looking at all the layouts you mentioned this impacts and they have different classes, shouldn't they all be the same if they are using a shared component?

src/templates/layouts/vetCenterOutstation/index.tsx
```diff
@@ -181,67 +181,67 @@ export function VetCenterOutstation({
           >
             Location and contact information
           </h2>
-          <div
-            className="region-list usa-grid usa-grid-full vads-u-display--flex vads-u-flex-direction--column
-          mobile-lg:vads-u-flex-direction--row facility"
+          <TextWithImage
+            className="region-list facility"
+            hackyImageAndStaticMapHack
```
pwolfert, line 186: To answer your other question

> ...looking at all the layouts you mentioned this impacts and they have different classes, shouldn't they all be the same if they are using a shared component?

That's another point that makes standardization difficult: The _ImageAndStaticMap_ component requires a hacky layout solution around it. In order to display it at a proper column width, the parent of that image and static map component needs to know some implementation details of the image and static map, which makes the parent layout and that component tightly coupled. It's really not ideal. I tried to make the coupling really explicit with his prop I've called `hackyImageAndStaticMapHack` ([see additional comments and naming alternatives](https://github.com/department-of-veterans-affairs/next-build/pull/1186#discussion_r2231546476)), but perhaps a better solution would be to create an entirely separate component where all the hacky CSS can be somewhat contained—something like `TextWithImageAndStaticMap`.

pwolfert: I believe either the current solution or the one I just proposed is an improvement over what was there before, because the previous version of the `ImageAndStaticMap` component had special CSS classes that assumed it was being placed within a CSS layout framework that the parent had to define and that had to be duplicated across multiple parents of the `ImageAndStaticMap`, including the new `VetCenter` component.

brianseek (approved):
This all looks cool to me. If QA of the affected page types looks good to @laflannery, lets roll it out.

src/templates/layouts/vetCenterOutstation/index.tsx
laflannery: For the most part, everything seems to be using the same classes now but I feel like there can be more clean up. Some of the things I think we can do:
- The flex utility classes are duplicative because `vads-grid-row` inherently contains these styles.
- `region-list` and `facility` classes don't seem to be doing anything so they probably could be removed
- The `inline-table-helper` class just makes things go wild and we probably shouldn't have this with the new v3 grid
- I think I would probably remove the `facility-img` class because having this, without the `inline-table-helper` class makes the image expand outside the grid
- I think for tablet a 7/5 split with a grid gap of 2 works a little better than 8/4, grid gap 5. The image ends up a little bigger on desktop but I think it makes it so that the image doesn't get too small on tablet

pwolfert: I've experimented with it, and I think removing `inline-table-helper` and `facility-img` would work, but it would only work if I were to override the `.facility-img` rule for the map image inside the widget through an additional custom CSS rule.

Not having to deal with that and allowing columns to behave as expected would make things way simpler, though.

@laflannery

pwolfert: Pushed an update. It would be nice if we were able to test these widgets on Tugboat.

PR #1220, opened 2025-08-05
Vet Center Locations: Show info for the main location

pwolfert:
# Description

Adds a bunch of data types and transformation code for rendering the location information for the main vet center and others. Right now for this ticket I'm only using it for the _main location_, but I'll be using the same component to render information for satellite locations in the next ticket.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21641

## Testing Steps

- Compare http://localhost:3999/secaucus-vet-center/locations/ to https://www.va.gov/secaucus-vet-center/locations/
- https://www.va.gov/st-paul-vet-center/locations/

## Screenshots

Screenshot 2025-08-05 at 1 59 18 PM


src/data/queries/vetCenterLocationListing.ts
```diff
@@ -42,6 +53,67 @@ export const data: QueryData<
   return entity
 }
 
+// Single formatter function that handles all Vet Center variants
+const formatVetCenterVariant = (
```
pwolfert, line 57: I expect to pare this down further as I figure out what information I actually need for the whole locations page.

src/types/formatted/vetCenterLocationListing.ts
```diff
@@ -1,11 +1,70 @@
...
+  lastSavedByAnEditor: string | null
+  image: FormattedMediaImage
+  fieldFacilityLocatorApiId: string
+}
+
+/**
+ * Main Vet Center info type for location listings.
+ * This is a simplified version of VetCenter for listing purposes.
+ */
+export type VetCenterLocationInfo = CommonVetCenterFields & {
```
pwolfert, line 28: Like the formatter function, this data will also likely change as I figure out what's truly necessary.

src/templates/layouts/vetCenterLocationListing/VetCenterLocationInfo.tsx
```diff
@@ -0,0 +1,149 @@
...
+        )}
+      </section>
+
+      <section className="region-grid usa-width-one-half vads-u-order--first mobile-lg:vads-u-order--initial vads-u-margin-bottom--2">
+        {image && isMainOffice && (
+          <ImageAndStaticMap
+            image={image}
+            facilityId={fieldFacilityLocatorApiId}
+          />
+        )}
```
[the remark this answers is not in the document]
pwolfert: When I come back to this, I also need to support two new widget type names used by the original template. The rest of this is identitical to what we have already except for the string we pass the widget.

src/data/queries/vetCenterLocationListing.ts
```diff
@@ -42,6 +53,67 @@ export const data: QueryData<
...
+  entity: NodeVetCenter | NodeVetCenterMobileVetCenter | NodeVetCenterCap
```
brianseek, line 58: What is up with `NodeVetCenterCap`? I see the formatting of the other two in the data but not this one. Is this just work ahead for later work?

brianseek (approved):
Looks good to me!

src/data/queries/vetCenterLocationListing.ts
pwolfert: Yeah we're not using the mobile centers or CAP yet, but I will in the next PR. They all have similar but not quite the same info, and they get rendered in basically the same way on the locations page. There's a lot of branching logic in the original template. I haven't yet decided if I'll try to keep the logic separate or combine it—not until I see it all together.

laflannery:
@pwolfert Some comments:
- From the H4s can you remove `force-small-header` and `vads-u-line-height--1`? We want to use just the default h4 styles for size and font family
- From looking at just the code, I think that the address needs to be wrapped in a `<p>` rather than a `<div>` I don't believe that it's doing to resize and scale properly. Are you able to test this locally and change it if needed? (FWIW, this is how Prod is an it is not correct)
- And just a question I have, cause I can't tell from the code, if a location doesn't have an official name, does the `aria-describedby` attribute not show at all?

pwolfert:
@laflannery 

1. H4s:
    - Removed from the address h4 ✅ 
    - The h4 for hours is being rendered by a shared `Hours` component, which has those classes. Removing them from this component would affect other pages. Do you wish to continue? (Y/N)
2. Address
    - This is actually using the shared component, and further down the tree inside the `<address>` element we have a `<p>`. This came up in the last template where at first we were wrapping the whole thing in a `<p>` tag, but it was invalid HTML because `<p>` elements can't have divs and other ps in them. I expect this to scale correctly as is.
3. Is `aria-describedby` dropped when not needed?
    - Yes, it uses similar logic to the Vets Center template and drops it if it points to nothing. React doesn't render attributes that have an `undefined` value

laflannery (approved):
Approving based on code and comments. I imagine if there are any issue they will be small and we can handle them in the QA phase

PR #1222, opened 2025-08-06
Content release warnings

cvalarida:
# Description

When we get 403 errors for archived content during the content release process, we want to know about it. It shouldn't break the build, but we want to know about it. This sends a notification to Slack when we do have 403 errors.

It does it by writing warnings to files (one for each child process, since that's a thing), globs all those warnings from all those files together into a single file (sorted by timestamp for convenience), and if that file exists, sends a warning notification to Slack with its contents.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21757

### DIFF
--- a/envs/.env.dev
+++ b/envs/.env.dev
@@ -39,8 +39,8 @@ NEXT_PUBLIC_DATADOG_RUM_SESSION_SAMPLE_RATE=100
 
 # Feature flags for enabling content types. These should only be added when you are preparing to go to prod and are testing on dev.
 # It is better to test these from the CMS backend for dev than here.
-FEATURE_NEXT_BUILD_CONTENT_EVENT_LISTING=true
-FEATURE_NEXT_BUILD_CONTENT_EVENT=true
-FEATURE_NEXT_BUILD_CONTENT_HEALTH_CARE_LOCAL_FACILITY=true
+# FEATURE_NEXT_BUILD_CONTENT_EVENT_LISTING=true
+# FEATURE_NEXT_BUILD_CONTENT_EVENT=true
+# FEATURE_NEXT_BUILD_CONTENT_HEALTH_CARE_LOCAL_FACILITY=true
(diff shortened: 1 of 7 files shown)


envs/.env.dev
cvalarida, line 1: Just for testing the content release. This way, the releases are way shorter, making for a faster feedback loop. This needs to be reverted before we merge this sucker.

packages/proxy-fetcher/src/index.ts
cvalarida, line 1: Similarly, this is just for testing and should be reverted before merging.

Because it's a pain to _actually_ archive content in the middle of a content release just for testing, this simulates archived content 10% of the time (so I'm expecting about 14 messages for 403s sent to Slack 🤞).

scripts/yarn/export.js
```diff
@@ -2,16 +2,75 @@ import { processEnv } from 'env-loader'
...
+  for (const file of files) {
+    const filePath = path.join(logsDir, file)
+    const content = fs.readFileSync(filePath, 'utf8')
+    const lines = content.split('\n')
+
+    for (const line of lines) {
+      if (line.trim() === '') continue
+      // Match lines that start with a timestamp
+      // Alternatively, we could write JSON as log lines and parse them...if we
+      // had more metadata. ¯\_(ツ)_/¯
```
cvalarida, line 56: I could go either way, honestly. If for whatever reason it's malformed JSON, it'll throw an error during parsing, so I kept it like this for now.

One reason we might want to add more metadata to spiffy this up is if we wanted our Slack notification to glob warnings by warning type, sorted by content type. Or something fancy like that. I don't anticipate we'll see a need for this, but it's an option. 🤷

cvalarida:
> My one comment is that this message should be alerting `@cms-troubleshooting`. Are we able to do that?

Oh, we can prepend that text with `@cms-troubleshooting\n`. 👍

pwolfert (approved):
Looks like it's working 🤩. My question/suggestion above doesn't affect the actual functionality

PR #1231, opened 2025-08-11
Vets Center: Remove usa-width-three-fourths

pwolfert:
# Description

Takes care of a legacy grid class that I missed in [the initial migration](https://github.com/department-of-veterans-affairs/next-build/pull/1186). The whole page was still wrapped in a class that isn't necessary anymore because the `usa-content` gives the page content a max width anyway. There's nothing I need to actually replace it with, which means we can get rid of a `div` and reduce the indentation in this file.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21948

## Testing Steps

The max page width on http://localhost:3999/secaucus-vet-center/ should look basically the same as it does in production.

### DIFF
--- a/src/templates/common/featuredContent/index.tsx
+++ b/src/templates/common/featuredContent/index.tsx
@@ -7,7 +7,7 @@ export function FeaturedContent({
   link,
 }: ParagraphComponent<FormattedFeaturedContent>) {
   return (
-    <va-card class="vads-u-flex--fill vads-u-margin-bottom--2 vads-u-margin-right--0 medium-screen:vads-u-margin-x--0p5 spotlight-card">
+    <va-card class="vads-u-flex--fill vads-u-margin-bottom--2 vads-u-margin-right--0 tablet:vads-u-margin-x--0p5 spotlight-card">
       {title && (
         <>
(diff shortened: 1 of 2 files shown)


laflannery:
@pwolfert I'm doing full QA on the grid migration now and there are a couple other things that need to be adjusted:
- For the summary box, because of the 768 to 640 breakpoint change, there is no space between the summary box and the on this page component at 640 now.
   - Can we change `tablet:vads-u-margin-bottom--0` on the `<va-summary-box>` to `desktop:vads-u-margin-bottom--0`? I would rather have a large space between 768 and 640 than no space at 640
- On the spotlight `<va-card>` components, those still have `medium-screen:vads-u-margin-x--0p5` so at 640 there is no space between the cards. Can you change that to `tablet`?

pwolfert:
@laflannery, good solution to that first one 👍 

## Screenshots for summary box gap

### Before

Screenshot 2025-08-12 at 11 30 33 AM

### After

Screenshot 2025-08-12 at 11 29 36 AM

## Screenshots for featured content gap

### Before

Screenshot 2025-08-12 at 11 29 18 AM

### After

Screenshot 2025-08-12 at 11 29 27 AM

PR #1234, opened 2025-08-12
Vet Center Locations: Satellite locations

pwolfert:
# Description

Adds the "Satellite locations" and "Other nearby Vet Centers" sections.

> [!IMPORTANT]
> It took me a while to figure this out, but the map widgets for the satellite locations relies on global variables defined for the "Other nearby Vet Centers" widget, so these two things are intertwined. I've tried to make this secret coupling explicit by not coupling one sibling to another but move the definitions of the global variables up to the parent so the coupling is vertical instead of horizontal. It's not great, but it's possibly better. Open to suggestions here.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21642
Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21643

## Testing Steps

- An outstation and two mobile vet centers: [local](http://localhost:3999/white-river-junction-vet-center/locations/), [prod](https://www.va.gov/white-river-junction-vet-center/locations/)
- A lot of satellite locations that must be properly sorted: [local](http://localhost:3999/fargo-vet-center/locations/), [prod](https://www.va.gov/fargo-vet-center/locations/)
- Just a mobile vet center (bug in prod where the archived Rchimond Mobile Vet Center is shown): [local](http://localhost:3999/alexandria-va-vet-center/locations/), [prod](https://www.va.gov/alexandria-va-vet-center/locations/)

## Screenshots

Screenshot 2025-08-13 at 1 01 54 PM


src/templates/layouts/vetCenterLocationListing/index.tsx
```diff
@@ -1,13 +1,35 @@
...
+      window.mainVetCenterPhone = mainOffice.phoneNumber
+      window.mainVetCenterAddress = mainOffice.address
+      window.mainVetCenterId = mainOffice.fieldFacilityLocatorApiId
+
+      // Needed by both the facility map widgets and the nearby vet centers widget
+      window.satteliteVetCenters = satelliteLocations.map(
+        (location) => location.fieldFacilityLocatorApiId
+      )
+    }
+  }, [mainOffice, satelliteLocations])
```
pwolfert, line 26: Some more archeological information about this:

- [vets-website code that reads satelliteVetCenters to make map widgets](https://github.com/department-of-veterans-affairs/vets-website/blob/040e7c580b204a2481885fd3a9b3522e504a5c90/src/applications/static-pages/facilities/createFacilityPageSatelliteLocations.js#L18)
- [vets-website code that reads all of these variables to make the other widget](https://github.com/department-of-veterans-affairs/vets-website/blob/037403e2fd6f867a2eb13beb797ffcccc790aa79/src/applications/static-pages/facilities/vet-center/createNearByVetCenters.jsx#L16-L20)
- [last time that widget code was changed to fix the map not showing for outstations](https://github.com/department-of-veterans-affairs/vets-website/pull/35285) (but it's still broken for CAPs)
- [liquid template in content-build where the map widget element was being created](https://github.com/department-of-veterans-affairs/content-build/blob/eb8fd10c88f70f9856c905e11d0ec1706d7419e8/src/site/includes/image_and_static_map_dynamic.liquid#L12)
- [liquid template in content-build where the "other nearby" widget was being created and all the variables used by both widgets were being defined](https://github.com/department-of-veterans-affairs/content-build/blob/eb8fd10c88f70f9856c905e11d0ec1706d7419e8/src/site/includes/vet_centers/nearby.liquid) (the secret coupling)

pwolfert:
Temporarily [deployed it to the staging environment](https://github.com/department-of-veterans-affairs/next-build/actions/runs/16946316254) so I could see if the widgets would work. 

<details>
<summary>Here are before and after screenshots of the map widget not working for Outstations and then working:</summary>

## Before
Screenshot 2025-08-13 at 11 54 11 AM

## After
Screenshot 2025-08-13 at 12 24 12 PM

</details>

laflannery:
> What are your screenshots pointing out exactly? The label being "Address" vs "Located at"? I used the existing Address component we've been using for the other templates, and I was thinking the consistency would be a plus. I also noticed during my testing and see in your screenshot that there's a first line to the addresses on prod that isn't present in the address component. Do we also want to match prod here even though it seems like that first line is redundant?
@pwolfert I'm pointing at the missing first line of the address info - "Shea Center" and "Wayne County Veterans Services". I don't mind that we changed "Located at to "Address", I agree with you that consistency is good, but we need that first line of data. In some instances I can see what you mean about it being redundant (Shea Center) but there are many cases where it's not. If you look at the Satellite locations on Greenvilee, NC you can see that the info in this field is unique: https://www.va.gov/greenville-nc-vet-center/locations/

pwolfert:
I compared the global variables being passed to the "other nearby vet centers" widget in staging and prod, and I see that the address object is formatted differently:

Screenshot 2025-08-14 at 11 06 37 AM

Hopefully if I transform the data to what it expects the results will match

pwolfert:
Kicked off a deploy to Staging

pwolfert:
Whoops, I accidentally committed my v3 migration to this branch, and now Tugboat will rebuild

laflannery:
One small adjustment - can you add `vads-u-margin-top--0` to the H4 for "Phone" so the spacing matches Prod? I tested and if this is added for all the "Phone" headings, even the Main location this work fine so there shouldn't need to be any conditional, weird logic on it.

And then one question which I think the answer is - you touched nothing - but I feel like I need to ask just in case:

- The nearby locations are now matching but I'm seeing Operating statuses on Staging that are not there on Prod. But this is just a widget right, so we wouldn't control the data that each item pulls in right?
Screenshot 2025-08-14 at 3 26 26 PM

PR #1243, opened 2025-08-14
Rename the type `ParagraphCCVetCenterFaqs` to `ParagraphCCQaSection`

pwolfert:
# Description

- Renames the _centralized content_ (CC) type `ParagraphCCVetCenterFaqs` to `ParagraphCCQaSection` since it is used by more than just the Vet Centers
- Makes the relationship between `ParagraphQaSection` and `ParagraphQA` clearer by actually using the `ParagraphQA` in `ParagraphQaSection`

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21861

## Testing Steps

If your editor supports it or you have some other TypeScript tools, follow the usage for some of these types to get a feel for how they're used. What's a little confusing but is at least consistent is that most of the `Paragraph*` types are the untransformed Drupal data structures, whereas when they get "formatted", the name changes from something like `ParagraphQA` to `QaParagraph`; essentially the type name is flipped.

### DIFF
--- a/src/mocks/qaSection.mock.json
+++ b/src/mocks/qaSection.mock.json
@@ -75,10 +75,39 @@
         {
           "type": "paragraph--wysiwyg",
           "id": "f6dfd8da-877f-46bd-9707-28861c27003e",
-          "resourceIdObjMeta": {
-            "target_revision_id": 3011,
-            "drupal_internal__target_id": 2754
-          }
+          "drupal_internal__id": 2754,
+          "drupal_internal__revision_id": 3011,
+          "langcode": "en",
+          "status": true,
+          "created": "2019-03-04T22:24:09+00:00",
+          "parent_id": "2755",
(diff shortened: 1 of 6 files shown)


src/mocks/qaSection.mock.json
```diff
@@ -75,10 +75,39 @@
         {
           "type": "paragraph--wysiwyg",
           "id": "f6dfd8da-877f-46bd-9707-28861c27003e",
-          "resourceIdObjMeta": {
-            "target_revision_id": 3011,
-            "drupal_internal__target_id": 2754
-          }
+          "drupal_internal__id": 2754,
+          "drupal_internal__revision_id": 3011,
```
pwolfert, line 79: When I started using the specific paragraph type in `ParagraphQaSection`, these unhydrated mock objects would no longer cut it for keeping TypeScript happy.

PR #1264, opened 2025-08-25
Next.js 15 upgrade

cvalarida:
# Description

Upgrades Next.js from 14 to 15. Along the way, we also have to upgrade React to 19, which is where most of the substantive changes come from.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21377


src/templates/layouts/vamcSystem/index.test.tsx
```diff
@@ -196,7 +196,7 @@ describe('VamcSystem with valid data', () => {
       expect(vaLink).toBeInTheDocument()
       expect(vaLink).toHaveAttribute('href', `${mockData.path}/locations`)
       expect(vaLink).toHaveAttribute('text', 'See all locations')
-      expect(vaLink).toHaveAttribute('active', 'true')
+      expect(vaLink).toHaveAttribute('active', '')
```
cvalarida, line 199: I think this is due to a change in how React 19 handles boolean values for DOM attributes. I double-checked that it's working properly in the browser, and it's golden.

src/products/newsStory/formatted-type.ts
```diff
@@ -6,7 +6,7 @@ import { PublishedEntity } from '@/types/formatted/publishedEntity'
 import { MediaImage } from '@/types/formatted/media'
 
 export type NewsStoryTeaser = PublishedEntity & {
-  headingLevel?: ComponentType | keyof JSX.IntrinsicElements
+  headingLevel?: ComponentType | keyof React.JSX.IntrinsicElements
```
cvalarida, line 9: React 19 no longer uses the global `JSX` namespace, so all this is scoped under the `React` namespace.

src/templates/components/newsStoryTeaser/index.tsx
```diff
@@ -11,7 +11,7 @@ export const NewsStoryTeaser = ({
   introText,
 }: FormattedNewsStoryTeaser) => {
   const TitleTag = ({ children, className }) => {
-    const Heading = headingLevel ? headingLevel : 'h2'
+    const Heading = (headingLevel ? headingLevel : 'h2') as React.ElementType
```
cvalarida, line 14: Apparently React 19 has more strict type checking for what we can use as a `<Component>`. 🤷

src/types/jsx.d.ts
```diff
@@ -0,0 +1,24 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import 'react'
+
+declare module 'react' {
```
cvalarida, line 4: See https://react.dev/blog/2024/04/25/react-19-upgrade-guide#the-jsx-namespace-in-typescript for more details.

src/types/jsx.d.ts
```diff
@@ -0,0 +1,24 @@
...
+  namespace JSX {
+    interface IntrinsicElements {
+      'va-accordion': any
```
cvalarida, line 7: Ideally, we don't leave these as `any`, but I don't think they had proper types anyhow. 🤷 It'd be nice if we could import the prop (attribute) types for each web component, but I don't think the component library exports that. We don't want to use the prop types for the React bindings.

.nvmrc
```diff
@@ -1 +1 @@
-22.16.0
\ No newline at end of file
+24.1.0
```
cvalarida, line 1: So...yeah, this is just a cool two-major-version bump to Node. 🙈 

I had to bump it up to at least 23 for the linting to work. I'm not _really_ sure why, but I kept running into an error in CI:
```
TypeError: Failed to load plugin 'jsx-a11y' declared in '.eslintrc.json » eslint-config-next/core-web-vitals » /home/runner/work/next-build/next-build/node_modules/eslint-config-next/index.js': `[[GeneratorState]]` is not present on `O`
    at Object.assert (/home/runner/work/next-build/next-build/node_modules/eslint-plugin-jsx-a11y/node_modules/internal-slot/index.js:18:10)
    at GeneratorStart (/home/runner/work/next-build/next-build/node_modules/eslint-plugin-jsx-a11y/node_modules/es-iterator-helpers/aos/GeneratorStart.js:12:7)
    at CreateIteratorFromClosure (/home/runner/work/next-build/next-build/node_modules/eslint-plugin-jsx-a11y/node_modules/es-iterator-helpers/aos/CreateIteratorFromClosure.js:42:2)
    at Array Iterator.filter (/home/runner/work/next-build/next-build/node_modules/eslint-plugin-jsx-a11y/node_modules/es-iterator-helpers/Iterator.prototype.filter/implementation.js:75:15)
    at Object.<anonymous> (/home/runner/work/next-build/next-build/node_modules/eslint-plugin-jsx-a11y/lib/util/isInteractiveElement.js:72:70)
    at Module._compile (node:internal/modules/cjs/loader:1730:14)
    at Object..js (node:internal/modules/cjs/loader:1895:10)
    at Module.load (node:internal/modules/cjs/loader:1465:32)
    at Function._load (node:internal/modules/cjs/loader:1282:12)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
```

I didn't have this problem locally, and discovered that's because I was running v23 locally. v24.1.0 is the current stable release, so I figured I'd just jump us all the way here and see what happens. If CI passes and everything's good on Tugboat, I think we're good to go. If anybody knows otherwise, let's talk about it!

additional.d.ts
```diff
@@ -19,23 +19,3 @@ declare module 'debug'
 declare module 'mq-polyfill'
 declare module 'nock'
 declare module 'test-utils'
-
-declare namespace JSX {
```
cvalarida, line 23: I didn't actually need to move this to a new module, but I liked keeping it a little more separate. 🤷

lint-staged.config.js
```diff
@@ -6,9 +6,9 @@ const lintStaged = {
   // Lint and format TypeScript and JavaScript files
   '*.(ts|tsx|js|jsx|mjs)': (filenames) => [
     // some extra handling to use next's eslint https://github.com/vercel/next.js/issues/27997#issuecomment-900554790
-    `yarn lint --fix --file ${filenames
-      .map((file) => file.split(process.cwd())[1])
-      .join(' --file ')}`,
+    `yarn lint:files --fix ${filenames
+      .map((file) => file.split(`${process.cwd()}/`)[1])
+      .join(' ')}`,
```
cvalarida, line 11: I added a new `lint:files` script so we could pass individual files to it for use in `lint-staged`. This is because we previously were using `next lint` for our linting, but that's been deprecated in favor of straight-up `eslint`.

Along the way, I also discovered that the previous version of this was passing `/src/absolute/paths.ts` instead of relative paths, which understandably caused `eslint` to get confused. Apparently the `next lint` command didn't really care about passing absolute paths and just assumed you meant to give it a relative one? 🤷 At any rate, I had to make it pass relative paths by adding that trailing `/` to the `file.split()`.

tsconfig.json
```diff
@@ -30,7 +30,10 @@
       "@/types/*": ["./src/types/*"],
       "@/products/*": ["./src/products/*"],
       "@/playwright/*": ["./playwright/*"],
-      "react": ["./node_modules/@types/react"]
+      // Import the JS instead of the TS to avoid React 19 compatibility issues
+      "@department-of-veterans-affairs/component-library/dist/react-bindings": [
+        "./node_modules/@department-of-veterans-affairs/component-library/dist/react-bindings/index.js"
+      ]
```
cvalarida, line 36: There's a whole long Slack thread for this. The TL;DR is that the component library adds their `.ts` files to `dist/`, so when we import from the library, the module resolution favors the `.ts` files and type checking is done on those library files. React 19 introduces some stricter static types (apparently) and type checking these files inside the component library resulted in some errors. So instead of making it our upstream dependency a blocker to migration, I decided to just import the `.js` files and let that library take care of its own types.

From what I can tell, there's no compatibility problem with React 19 and our component library other than the static types, but we'll want to do some solid QA.

:bat-signal: @laflannery 😇

next.config.js
```diff
@@ -1,43 +1,79 @@
...
+  //arbitrary; 60 is default but it's too small
+  staticPageGenerationTimeout: 180,
+
   experimental: {
-    largePageDataBytes: 512 * 1000, // 512kb, is 128kb by default
+    // 512kb, is 128kb by default
+    largePageDataBytes: 512 * 1000,
+
+    // Enable React 19 optimizations
+    reactCompiler: true,
```
cvalarida, line 33: Calling this out. I'm not sure how much of the performance boost I saw was due to the compiler, but we can try it without if'n y'like.

src/assets/styles/globals.css
```diff
@@ -25,7 +25,7 @@ va-button#mdFormButton::part(button) {
 }
 
 .feedback-button {
-  background-color: var(color-primary);
+  background-color: var(--color-primary);
```
cvalarida, line 28: Webpack's CSS loader was apparently more forgiving than Turbopack's. This was the only problem I saw, though, so I think we're okay?

.nvmrc
brianseek: we should probably update the tugboat config to node 24 too. Line 25 of `.tugboat/config.yml` also some the GH actions should go to 24 as well

.github/workflows/preview.yml
```diff
@@ -52,7 +52,7 @@ jobs:
     - uses: actions/checkout@v5
     - uses: actions/setup-node@v4
       with:
-        node-version: 22
+        node-version: 24
```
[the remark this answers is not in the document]
brianseek: ➕ to @pwolfert's question

.nvmrc
brianseek: I think you need to rebuild without the base preview to see this take effect in tugboat.
Screenshot 2025-09-02 at 11 34 06 AM

pwolfert (approved):
Honestly not as many changes as I was expecting. The yarn cache updates made it look bigger.

brianseek (approved):
Other than some of the lingering questions around the node 24 stuff, this is looking good to me. Building and serving seem to be working normally

next.config.js
```diff
@@ -1,43 +1,79 @@
...
+      }),
     }
   },
+
+  turbopack: {
+    rules: {
+      '*.svg': {
+        loaders: ['@svgr/webpack'],
+        as: '*.js',
+      },
```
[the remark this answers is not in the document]
cvalarida: Next.js will use Webpack by default, yeah. Only when you pass in the `--turbopack` will it use Turbopack. We could probably get rid of the Webpack config altogether, but I don't know for sure if, for example, Turbopack will pick up any of that Webpack config or if we've got a rogue `next <something>` command lurking somewhere that takes advantage of the Webpack config.

I _think_ it's fine to remove, but 🤷

lint-staged.config.js
cvalarida: 🤔 That's not a bad idea. I only added that new script because I didn't even think about running _not_ a script in here. 😅 

If I had to think of an advantage, though, it'd be visibility + future changes. If we decide to modify the script to add some other CLI options / flags later. The `yarn eslint --fix ...` in this file is a little hidden unless you remember to go look for it. It's not a strong advantage, but if I had to think of one, that's all I've got. 😜

moanu404 (approved):
these changes LGTM!

timcosgrove (approved):

PR #1268, opened 2025-08-26
Component migration: Final cleanup

pwolfert:
# Description

This is another part of https://github.com/department-of-veterans-affairs/next-build/pull/1262.

Cleans up folders directly under `src` that are very sparse in order to reduce noise. See individual commit messages.

Proposal: What if we move `src/data/queries/index.ts` (the only file remaining under `src/data`) to `src/lib/drupal/queries.ts`?

## Ticket

Contributes to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/22001


laflannery:
Based on the files changed I only reviewed Stories and News releases. Those are both working as expected but let me know if I need to expand QA and review other templates

brianseek (approved):
Nice work cleaning up that unused hook and everything else looks good too! Also the proposal of moving `src/data/queries/index.ts` sounds fine to me too

pwolfert:
Okay, sweet about the moving of the queries index. Turns out I'll have to touch a lot of files, but at least it's a separate commit. I don't feel like creating a whole new PR

PR #1278, opened 2025-08-28
VACMS-22026: Added performance testing to template migration documentation.

brianseek:
# Description

Adds guidance for performance testing of new page layouts before releasing to prod.

## Ticket

Closes [22026](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/22026)

## Developer Task

## Screenshots

Screenshot 2025-08-28 at 4 32 28 PM
Screenshot 2025-08-28 at 4 32 15 PM

### DIFF
--- a/docs/template-migration/phases/active-migration/qa.md
+++ b/docs/template-migration/phases/active-migration/qa.md
@@ -8,6 +8,7 @@ The QA ticket contains the specific items that must be reviewed and tested durin
 - Functionality
 - Design
 - Accessibility
+- [Check for degraded page performance with a lighthouse scan](#how-to-check-for-degraded-page-performance)
 
 For each item within the ticket, you will:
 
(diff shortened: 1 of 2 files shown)


laflannery:
@brianseek this look good. I have a few thoughts:
- I am doing pretty extensive accessibility testing during QA, so we probably could uncheck Accessibility for Lighthouse as well, unless there is a specific reason we want or should include that with the performance testing here?
- I think we should run desktop and mobile testing, I could even make a case that mobile is more important because the majority of traffic is coming from mobile these days
- Do we think we need to do this as part of QA rather than launch prep? My thought here is if we do find degradation, we would want to fix that and therefore it would stop the launch. If we did it as part of the QA process, the launch wouldn't have been scheduled yet.

brianseek:
> @brianseek this look good. I have a few thoughts:
> 
> * I am doing pretty extensive accessibility testing during QA, so we probably could uncheck Accessibility for Lighthouse as well, unless there is a specific reason we want or should include that with the performance testing here?
> * I think we should run desktop and mobile testing, I could even make a case that mobile is more important because the majority of traffic is coming from mobile these days
> * Do we think we need to do this as part of QA rather than launch prep? My thought here is if we do find degradation, we would want to fix that and therefore it would stop the launch. If we did it as part of the QA process, the launch wouldn't have been scheduled yet.
@laflannery 
- Yeah, I agree we could turn off accessibility. I just figured it wouldn't hurt to have it there. The other two, especially SEO give very different results because of the dev environment. So I really wanted to cut down on the noise there.

- The reason that I have only desktop is because we are just measuring if there is any drift between the two, and we mostly rely on RWD for mobile. So its less "how can we make this better" and more "is this different than before". That said, I'm fine including both mobile and desktop

- This could for sure be part of QA, I just didn't want to put another task on your plate. If you are okay with that, I'll move this around in the doc while making the other updates.

brianseek:
> @brianseek Your explanations makes sense. For where in the process, I personally think QA makes more sense but I will be curious what @timcosgrove thinks about where in the process makes the most sense though, since you guys talked about this the other day.

@laflannery The other good thing about being in QA, is that it gives us time to fix anything if we find something. But we are really not expecting this to ever be much different. And just because its a QA task, it doesn't have to be you doing it, you can always assign this ticket out to whoever.

I'll make that change along with some of the other things we were discussing, and see what Tim has to say.

PR #1307, opened 2025-09-09
MCP experiment

cvalarida:
# Description

This is a first pass at making an MCP server to provide context otherwise unavailable to LLMs to help with template migration. Specifically, this exposes the JSON:API schema.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/22121
Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/22297


cvalarida:
An alternative to adding all those JSON files is, of course, fetching the data from Drupal for each call. Or we could add it to the .gitignore and have the tool fetch the JSON file if it's not already downloaded. (We can call it a "cache" and sound all fancy. 😜 )

cvalarida (commented):
Soooooo many files. 🫠

We can _not_ add those, but that'd mean querying for the JSON schema from Drupal, which is...not great? 🤷

packages/template-migration-mcp-server/README.md
```diff
@@ -0,0 +1 @@
+# Template Migration MCP Server
```
pwolfert, line 1: Could AI write some documenation here?

cvalarida: Just added it!

pwolfert (approved):
I'm still running into issues with the entity-fetch tool, but everything else works great!

cvalarida:
Awesome! Yeah, we can iterate on that separately. 👍

PR #1309, opened 2025-09-10
API Explorer: Save the queries to the url query string for sharing and persistence

pwolfert:
# Description

I was getting tired of having to re-enter my filters when exploring the API data. I also wanted to be able to share what I was seeing with teammates. Now you can copy the URL and share it.

## Screenshots

Before:

Screenshot 2025-09-10 at 9 02 10 AM

After:

Screenshot 2025-09-10 at 9 02 02 AM


cvalarida (approved):
💯

PR #1328, opened 2025-09-16
Add Lovell switching to the Billing & Insurance page

pwolfert:
# Description

What does this PR address?

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21972

## Testing Steps

http://localhost:3999/lovell-federal-health-care-va/billing-and-insurance/

## Screenshots

Screenshot 2025-09-16 at 3 04 35 PM


brianseek (approved):
🚀

pwolfert:
Tugboat is currently on its second rebuild after I rebased this branch from main. It keeps timing out, and it doesn't make sense to me that it would be any of the changes in this branch causing that, but we'll see 🤷‍♂️

PR #1330, opened 2025-09-17
VAMC policies: All CC text

cvalarida:
# Description

I plugged in the Liquid template, transmogrified it to React, and filled in all the CC text. There are placeholders and `TODO`s for stuff we still need to do, but this is the bulk (or all?) of the content on the page.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21992
Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21997
Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21998
Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21999
Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/22000

This also adds `field_vamc_other_policies`, which I didn't see ticketed out. 🙈

### DIFF
--- a/src/components/vamcSystemPoliciesPage/template.test.tsx
+++ b/src/components/vamcSystemPoliciesPage/template.test.tsx
@@ -1,11 +1,203 @@
 import React from 'react'
 import { render, screen } from '@testing-library/react'
 import { VamcSystemPoliciesPage } from './template'
+import { formatter } from './query'
+import { NodeVamcSystemPoliciesPage } from '@/types/drupal/node'
+import mockData from './mock.json'
 
+const mockEntity = mockData as unknown as NodeVamcSystemPoliciesPage
+const formattedData = formatter(mockEntity)
+
+/**
+ * Most of these tests are just making sure that the data is rendered correctly.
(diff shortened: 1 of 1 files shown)


laflannery:
This ticket was also done: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21994

pwolfert (approved):
Looks good to me as far as translating from what was already done in the query goes. I do want to point out in case we revisit this that we now have a simple way of converting any CC content into paragraph types, which can then be rendered by the `<Paragraph>` component.

pwolfert:
@cvalarida, yeah! We have a `normalizeEntityFetchedParagraphs` function now ([example here](https://github.com/department-of-veterans-affairs/next-build/blob/main/src/components/vamcSystemRegisterForCare/query.ts#L130)) which will prepare those CC fields to be passed into the standard `formatParagraph` function. If you use the `formatParagraph` function instead of a more specific one, you just need to type cast the output to the specific paragraph you want.

Issue #1299, opened 2025-09-08
Update faker to v10.0.0


brianseek:
This [dependabot PR](https://github.com/department-of-veterans-affairs/next-build/pull/1261) needs some extra help to make the migration to v10. I think it has something to do with [faker moving to ESM-only](https://v10.fakerjs.dev/guide/upgrading.html#node), but needs more investigation. 

Brian and Patrick took a look at this and from that discussion:

It looks like everywhere we're importing faker right now it's using ESM syntax, but that could be deceptive with Jest. Possible that it's not truly ESM. Looks like we're only using it in a couple of JS modules (not TypeScript)

There are two test modules that import them from our own test-utils.ts module, and one of those is some kind of "example test"?? There's one other place where we import from the Baker package directly, but it's an unused import that could be cleaned up

Looks like a good opportunity to clean up and convert those modules to TypeScript and see what happens

### Acceptance criteria

- [ ] Convert required modules to TypeScript
- [ ] Determine if there are errors
- [ ] If there are, can the errors be resolved within the scope of this ticket?
  - [ ] If yes, fix them
  - [ ] If no, determine next steps

PR #1333, opened 2025-09-18
Delete `example_tests`

pwolfert:
# Description

Removes the example tests because we don't use them. We talked about it as a team during office hours, and nobody has made use of them. We considered the value of keeping them around:

- It's more work to maintain them as documentation because they can get out of date
- While AI might be able to benefit from a succinct example of our test strategies, it can just as easily look to existing tests in the project now that the project is more mature. We considered for a moment whether we should consolidate the example tests into a single markdown file with snippets, but it would have the same risk of getting stale as previously mentioned.
- Nobody on the team has noticed them or used them, so what makes us think future teammates will?

## Ticket

Loosely related to https://github.com/department-of-veterans-affairs/next-build/issues/1299


brianseek (approved):
🪦

PR #1341, opened 2025-09-22
Upgrade faker library to 9.9.0. We'll leave it there

pwolfert:
# Description

We're just going to leave it at 9.9.0 and ignore future upgrades. The library has no dependencies, and we don't use it in production.

## Ticket

Closes https://github.com/department-of-veterans-affairs/next-build/pull/1261

## Testing Steps

Run the unit tests and typechecks.


brianseek (approved):
🏆

PR #1343, opened 2025-09-23
Add H2 IDs

cvalarida:
# Description

Originally, the goal was to add IDs to the H2s inside WYSIWYG paragraphs.

Along the way, to make that happen and allow us to optionally turn it off for certain elements, I added `type` properties to all the paragraph static types so we can use a "proper" discriminated union inside the `formatParagraph()` function. This lead to a bunch of other little changes like using type assertions in all those query tests files because importing the JSON doesn't preserve the `type: 'whatever'` string literal, which means TS gets angry because the types of the paragraphs are mismatched.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/22272

## Screenshots

image


pwolfert (approved):
Nice work! I love that we can use the `type` as a descriminator now!

cvalarida:
> * When do the IDs show up? What are the requirements?

Right now, whenever we use `getHtmlFromDrupalContent()` and don't tell it to _not_ add those IDs. This includes calls to `getHtmlFromField()` and `formatParagraph()`.

I didn't dig into where all we're _not_ using those functions when we could, but changing them _should_ be pretty straightforward. Probably. 😅 
 
> * I can't test these scenarios because these content types aren't in next yet but I want to confirm the following:
>   * We want the following within Drupal content to have IDs
>     * Alerts blocks
>     * Featured content WYSWIWYGs

That shouldn't be a problem!

>   * We do NOT want the following to have IDs
>     * Headings inside accordions

This also shouldn't be a problem. We should know what WYSIWIG contents go inside accordions, and just..._not_ add them. We'd do that by passing a `{ addH2Ids: false }` option to any of those functions I mentioned above. (This is likely in the query formatter function.)

> * Can we protect against duplicate headings and therefore duplicate IDs?
>   * Example: https://www.va.gov/texas-valley-health-care/work-with-us/ There are 2 "Jobs and Careers" headings. I'm not saying this is a good thing to do but the way the current content-build plugin works is that if a duplicate heading is found, a number is added to make the ID unique.
>   * Currently in next we are ending up with a duplicate ID and therefore the anchor links get confused: https://next-kfwepuxgjlz4plg1oxtlikkk7knqsi3g.demo.cms.va.gov/cincinnati-health-care/policies/

This is....problematic. 😬 Those functions I mentioned above don't actually have visibility into the contents for the entire page (to know if there's a duplicate heading ID), so...we've got a few options, I think, but I'm puzzling through them.

Options include:
1. Somehow post-processing an entire page to rejigger the heading IDs as needed
    - This is pretty gross for a few reasons, but the two I'm most concerned about are the architecture change required to make this post-processing possible and the performance implications that'd have on our static build times, since we'd need to parse and process the DOM on thousands of pages
2. Adding a unique counter to all generated IDs
    - We'd make the function a closure and append a number on all generated IDs. The number would either be universally unique (each heading gets +1), or increments only for duplicate headings
    - These counters would show up in the links (hover over it and see that it goes to `#privacy-and-patient-rights-1`, for example)
    - I think we'd need to always add the counter (starting at 1) in case the ID is specified in the WYSIWYG contents (not generated); if that happened, we'd still get duplicate IDs (though likely less frequently)
 3. Using UUIDs
     - Probably these will change every content release, which...isn't awesome for a few reasons.
     - There might be a way to generate UUIDs in a way that's deterministic, but that'll take some puzzling, and it'll still be kinda funky, and I can't guarantee they'll keep the same IDs across content releases
         - I'm thinking like a seeded UUID generator that we initialize at the beginning of the content release...or something. I dunno. It's wonky, and I'm not a big fan of this approach because it seems unnecessarily complicated.
4. Using some other deterministic, random ID based on the page data
    - But when that page data changes, so too will the IDs, which won't be good for saved deep links

laflannery:
Also just noting that Stories content didn't have IDs because it's not being formatted. That's just something we should do and I will ticket that but not an issue here since we don't actually have the ToC component

laflannery (approved):

PR #1347, opened 2025-09-24
VAMC Policies Lovell switcher

cvalarida:
# Description

Adds the Lovell switcher to the VAMC Policies page

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21995

## Screenshots

image


src/components/vamcSystemPoliciesPage/query.test.ts
```diff
@@ -68,4 +87,41 @@ describe('VamcSystemPoliciesPage formatter', () => {
...
+        isLovellVariantPage: true,
+      },
+    })
+
+    expect(result.lovellSwitchPath).toBe(
+      '/lovell-federal-health-care-tricare/policies'
+    )
+  })
+
+  // NOTE: There is no TRICARE variant of this content type
```
cvalarida, line 126: I'm pretty sure that's true, anyhow. CC @laflannery 

The `formatter()` would work just fine if it _did_ exist, though.

src/components/vamcSystemPoliciesPage/query.ts
```diff
@@ -89,6 +93,13 @@ export const formatter: QueryFormatter<
     bottomOfPageContent: formatCcWysiwyg(
       entity.field_cc_bottom_of_page_content
     ),
+    lovellVariant: lovell?.variant,
```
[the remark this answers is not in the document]
cvalarida: image
Ooh, good eye. I do need to update it.

cvalarida:
I updated the breadcrumbs for Lovell.
image

Note the differences from prod:
- 1 & 2: The breadcrumbs are suffixed with ` - VA` instead of ` va` (which I think is nicer?)
- 3: The Lovell switcher heading doesn't show up in the `On this page` section (which I think is good?)

src/components/vamcSystemPoliciesPage/query.test.ts
laflannery: Yes - no TRICARE variant, that is built with the VAMC Detail page. That is because that the CC content doesn't apply to TRICARE so it would have been too much customization.

PR #1349, opened 2025-09-25
Fix MCP server's `fetch-entity` tool

pwolfert:
# Description

Possibly because I had a different workspace configuration (both `next-build` and `content-build` folders in my workspace) or because I was using a different AI chat tool than @cvalarida, the `fetch-entity` tool of the new MCP server just didn't work. I had two primary problems:

1. Certs wouldn't load for interacting with the Drupal API through `proxy-fetcher`
    - To fix the certs issue, I've added a `cwd` property to the mcp server settings (in my tool and in the example in the README). This fixes it by making sure the script is run from the root of the repo so the cert paths in `proxy-fetcher` are correct.
    - When I first published this PR, my solution was to update the `proxy-fetcher` code to use `__dirname` in the cert paths, but the dev server actually calls the function in two different contexts, and for some reason the __dirname is messed up in the second.
2. Node would get tripped up on TypeScript things
    - While @cvalarida did address this with his shell script wrapper for his `fetch-entity` script, that script assumed I had `nvm` installed, and I actually use a different node version manager.
    - There was also the problem of making all of the TS code compatible with plain "type stripping" available in node 24. While we could add tsconfig settings to make sure we didn't do things incompatible with type stripping, it wasn't a foolproof solution. I ended up using `npx tsx` to execute the TypeScript in a way that was compatible with all versions of node (that have npx) and wouldn't require `tsx` to be installed globally.

Additionally, I updated the two TypeScript based script tools in `./scripts` to be executed with TSX as well for consistency and to simplify them.

Also this https://github.com/department-of-veterans-affairs/next-build/pull/1349/commits/457cdcc8b1ab9f6420c9302dd0cdd5bcc9199e85

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/22338

## Testing Steps

Follow the MCP server package's readme to set it up in your tool of choice, then try to prompt the AI to use it.

https://github.com/user-attachments/assets/d2625434-68fc-4074-bba3-884af93aaffe


scripts/fetch-entity/src/drupalClient.ts
```diff
@@ -5,34 +5,13 @@
...
 import { DrupalClient } from 'next-drupal'
-
-/* eslint-disable import/no-extraneous-dependencies */
-import { SocksProxyAgent } from 'socks-proxy-agent'
-import fetch from 'node-fetch'
-
-const syswideCas = await import('syswide-cas')
-syswideCas.addCAs('certs/VA-Internal-S2-RCA-combined.pem')
-syswideCas.addCAs('certs/rootCA.pem')
+import { getFetcher } from 'proxy-fetcher'
```
pwolfert, line 8: Simplified this to use `proxy-fetcher`

scripts/fetch-entity/src/index.ts
```diff
@@ -50,76 +27,21 @@ program
     'Limit the number of resources to fetch (only with --collection)',
     '1'
   )
-  .action(
```
pwolfert, line 53: Moved this into a separate module because I was at one point trying to load the TS file for this directly in the mcp server and call the function directly. I ended up not doing that in the end, but it seemed like an okay thing to separate into its own module to separate the command-line interface from the business logic of the tool itself.

pwolfert:
Well, I thought I had tested this well enough, but it looks like my changes have broken the dev server. I totally tested that, so I'm not sure what happened.

cvalarida (approved):
Gonna be honest: I didn't review this one thoroughly. 🙈 Seems fine, though.

Only thing that really raises an eyebrow for me is the `scripts/dependency-tree-visualizer/*` changes, which I think are totally unrelated?

scripts/dependency-tree-visualizer/README.md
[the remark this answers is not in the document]
pwolfert: We talked about bun being a dependency for that one and how I was going to update it to use TSX. I was just handling all those scripts migrating to TSX once

scripts/fetch-entity/src/drupalClient.ts
pwolfert: It's a good question. I didn't want to change too much at once, but I can try it now

PR #1344, opened 2025-09-23
VAMC health services: Services

pwolfert:
# Description

Yo dawg, adds services to the services page.

Something that I've changed since I first opened the PR: I'm no longer getting the `node--regional_health_care_service_des` nodes from the VAMC System's `field_clinical_health_services`. We found that that data was unreliable, so I'm instead querying for those `node--regional_health_care_service_des` directly and filtering on the VAMC System id. Because of this, I've also removed that field from the `NodeHealthCareRegionPage` definition, and there were several opportunities to delete mock data.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21803

## Testing Steps

http://localhost:3999/boston-health-care/health-services/

## Screenshots

<details>
<summary>Full page screenshot</summary>

localhost_3999_boston-health-care_health-services_
</details>

PR #1350, opened 2025-09-25
Fix type error with mock data that is failing CI tests

pwolfert:
# Description

This fixes a type-checking failure on the main branch.

I had merged my https://github.com/department-of-veterans-affairs/next-build/pull/1344 after https://github.com/department-of-veterans-affairs/next-build/pull/1343 was merged without having updated it from the main branch, so the CI tests never caught the new error.


brianseek (approved):
👍

PR #1352, opened 2025-09-26
Account for incomplete link teasers

pwolfert:
# Description

The page `/pittsburgh-health-care/research/safety-security` had some incomplete data. When you look at that page data, the related links section exists, but it just has a single link with no fields filled out. In prod the related links section isn't shown, so the content-build code must be ignoring the incomplete link and not showing the whole section.

I've updated our code for the `LinkTeaser` formatter so it will ignore link teasers with a `null` link, and if a `ListOfLinkTeasers` has no valid link teasers in its list, it will itself return `null`.

Related to https://github.com/department-of-veterans-affairs/next-build/pull/1351

## Testing Steps

This is what has allowed https://github.com/department-of-veterans-affairs/next-build/pull/1351 to build.

## QA step


src/components/listOfLinkTeasers/mock.json
```diff
@@ -1,67 +1,203 @@
 {
   "type": "paragraph--list_of_link_teasers",
-  "id": "77159",
-  "field_title": "More information",
+  "id": "fd090f35-1c89-4bc6-9525-3e8c8221a87e",
```
pwolfert, line 3: The actual change here is that I replaced this file with real data from the API

cvalarida (approved):

laflannery (approved):
Confirmed Related links section is showing properly where links do exist

PR #1374, opened 2025-10-03
Medical records page: Add FAQ section

pwolfert:
## Description

Adds the FAQs, which also involved fixing a bug in parsing the boolean value of `field_accordion_display`. 

The bug was that the `entity_field_fetch` plugin was failing to translate the `1` and `0` values from Drupal and instead transliterated it as a string value. Our formatter function expected a Drupal paragraph entity coming from the Drupal API where boolean values are transmitted as JavaScript booleans. Because the string `"0"` is truthy in JavaScript, that flag was always on. This bug seemed to exist in `content-build` as well, because it showed the FAQ section of this page as an accordion, even though in the Drupal editing interface the corresponding checkbox was unchecked. To fix it, I'm moving the parsing of these boolean values and number values to the `normalizeEntityFetchedParagraphs`, whose purpose is to undo the `entity_field_fetch` mutilation.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21981

## Testing Steps

http://localhost:3999/boston-health-care/medical-records-office/

## Screenshots

Screenshot 2025-10-03 at 10 22 56 AM

### DIFF
--- a/src/components/vamcSystemVaPolice/__snapshots__/query.test.ts.snap
+++ b/src/components/vamcSystemVaPolice/__snapshots__/query.test.ts.snap
@@ -27,10 +27,10 @@ exports[`VamcSystemVaPolice formatData outputs formatted data 1`] = `
   "entityId": 63857,
   "entityPath": "/central-arkansas-health-care/va-police",
   "faqs": {
-    "displayAccordion": "1",
+    "displayAccordion": true,
     "header": "Other questions you may have about VA police",
     "id": "141575",
-    "intro": [],
+    "intro": null,
     "questions": [
       {
         "answers": [
--- a/src/components/reactWidget/query.ts
+++ b/src/components/reactWidget/query.ts
@@ -2,13 +2,6 @@ import { ParagraphReactWidget } from '@/types/drupal/paragraph'
 import { QueryFormatter } from 'next-drupal-query'
 import { ReactWidget } from '@/components/reactWidget/formatted-type'
 
-const toBoolean = (value: string | boolean) => {
-  if (typeof value === 'string') {
-    return value === '1'
-  }
-  return value
-}
-
 export const formatter: QueryFormatter<ParagraphReactWidget, ReactWidget> = (
   entity: ParagraphReactWidget
 ) => {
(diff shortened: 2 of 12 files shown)


src/components/vamcSystemVaPolice/__snapshots__/query.test.ts.snap
```diff
@@ -27,10 +27,10 @@ exports[`VamcSystemVaPolice formatData outputs formatted data 1`] = `
   "entityId": 63857,
   "entityPath": "/central-arkansas-health-care/va-police",
   "faqs": {
-    "displayAccordion": "1",
+    "displayAccordion": true,
     "header": "Other questions you may have about VA police",
     "id": "141575",
-    "intro": [],
+    "intro": null,
```
pwolfert, line 33: Once the values started getting correctly parsed, these values changed to accurately reflect the data

laflannery:
@pwolfert These CC FAQ section are working when the accordion checkbox is checked, however I tested when it's unchecked and I don't get any of the FAQ content, all I get is the section heading and section into:
- Medical records: (this has no section intro content added) https://next-jaxpjdlzbm5ny5arkfclvpxgdz1geogn.demo.cms.va.gov/boston-health-care/medical-records-office/#questions-about-medical-records---another-test
- VA Police: https://next-jaxpjdlzbm5ny5arkfclvpxgdz1geogn.demo.cms.va.gov/providence-health-care/va-police/#other-questions-you-may-have-about-va-police---update
- Vet Center: https://next-jaxpjdlzbm5ny5arkfclvpxgdz1geogn.demo.cms.va.gov/boston-vet-center/#how-vet-centers-are-different-than-a-medical-clinic---test

Tugboats for checked:
- Medical records: https://pr1374-bre9s7v1szccqkpkrhjlcbwylngmsgvs.tugboat.vfs.va.gov/boston-health-care/medical-records-office/#questions-about-medical-records
- VA Police: https://pr1374-bre9s7v1szccqkpkrhjlcbwylngmsgvs.tugboat.vfs.va.gov/providence-health-care/va-police/#other-questions-you-may-have-about-va-police
- Vet Center: https://pr1374-bre9s7v1szccqkpkrhjlcbwylngmsgvs.tugboat.vfs.va.gov/boston-vet-center/#how-vet-centers-are-different-than-a-medical-clinic

pwolfert:
@laflannery, the problem here should be fixed too. I made it so it can fall back to a heading and paragraph variation of the question and answer sections that we already had but hadn't hooked up.

PR #1766, opened 2026-03-20
Make VRT more reliable

cvalarida:
## Description

This PR does a couple things:
1. Cleans up the environment parity testing package's hooks
   - Most of the hooks weren't being used, so I removed them until we need them
   - The hooks also weren't previously definable per path
   - The `beforeComparison` hook didn't pass in the Playwright `Page` object, so we couldn't do any DOM manipulation for normalization right before comparison
2. Adds a couple hooks to stabilize the tests

### Screenshots

| Path |  Env screenshot | Diff screenshot |
| --- | --- | --- | 
| `/find-locations` |  envA | diff |
| `/view-change-dependents` | envA | diff |


brianseek (approved):
Looks good to me, I think it will be much more reliable ignoring the loading of things we can't control on those couple of pages.

PR #1820, opened 2026-04-09
Document current state of staging -> prod workflow

pwolfert:
## Description

I was going to think about how to fix the current bug we have in the staging workflow notifications. It's not priority right now, but this documentation was an artifact of me getting up to speed with how the workflows work together.


brianseek (approved):
LGTM! thanks for adding this

PR #1833, opened 2026-04-15
Tcosgrove/workflow permissions

timcosgrove:
Potential fix for [https://github.com/department-of-veterans-affairs/next-build/security/code-scanning/173](https://github.com/department-of-veterans-affairs/next-build/security/code-scanning/173)

Add a **workflow-level `permissions` block** so every job without its own block defaults to least privilege.

## Testing steps
This only enforces lowest level permissions. Workflows that need elevated permissions for specific steps (i.e. production-tag) are granted it when needed.


brianseek (approved):
⭐️

PR #1871, opened 2026-04-30
Resources & Support Listing: lighten query

zachbutton:
## Description

This CI workflow failed: https://github.com/department-of-veterans-affairs/next-build/actions/runs/25191326766/job/73861881656

Theory: Since it's consistently failing on page 2 of `support_resources_detail_page`, it's likely because the amount of fields are too much for CMS to handle in a single request.

Since we're requesting a lot of separate pages to populate the listing index, yet we don't display a whole lot of data, there was room to make the request more sparse than it already was. 

It also lightens the search fields in deeply nested collapsable panel to account for the sparser request.

## Testing Steps

- Test that Resources & Support still works locally
- ...and in tugboat
- A staging content-release run completes successfully


pwolfert (approved):
Nice! More efficient!

PR #1876, opened 2026-05-01
Support & Resources article listing page

zachbutton:
## Description

Reverts https://github.com/department-of-veterans-affairs/next-build/pull/1874, re-introducing the Resources & Support listing page.

Also fixed:

- `FEATURE_NEXT_BUILD_CONTENT_SUPPORT_RESOURCES_ARTICLE_LISTING` was not being respected
- When building locally with `SSG_CHERRY_PICKED_PATHS='/resources/[P]' yarn build`, the cherry-picked path would collide with `/[P]` if P was an existing page. An example of this is `health-care`: both `/health-care` and `/resources/health-care` are valid pages.

## Testing Steps

- Looks the same as prod
- Should 404 without `FEATURE_NEXT_BUILD_CONTENT_SUPPORT_RESOURCES_ARTICLE_LISTING=true`
  - should also not build when running `yarn build`


pwolfert (approved):
The `FEATURE_NEXT_BUILD_CONTENT_SUPPORT_RESOURCES_ARTICLE_LISTING` flag is now respected in both the dev server and static export 👍

PR #1890, opened 2026-05-06
Fix circular dependency in paragraph renderer

pwolfert:
## Description

Alternative to https://github.com/department-of-veterans-affairs/next-build/pull/1882. That fix was way more concise. This will behave a little nicer with compilers and avoid the same error in other nested paragraphs.

There were a few other ways I tried to fix it with the same general strategy, and this current way is a little more holistic and concise than those.

### Summary

Breaks the **`Alert` ↔ `Paragraph`** circular dependency by introducing a **paragraph render context** at the app composition root, moving nested paragraph rendering behind **`useRenderParagraphs()`** so alert components no longer import the paragraph dispatcher. Follow-up: **`AlertBlock`** now delegates body content through the same hook path, **`ExpandableText`** owns stable `va-additional-info` ids, and **global CSS** scopes margin tweaks to **`va-alert va-additional-info`** instead of a one-off class.

### Motivation

The paragraph dispatcher imported `Alert`, and `Alert` imported `Paragraph` to render nested paragraphs—creating a **mutual module dependency** that can surface as fragile, order-dependent runtime issues (e.g. invalid element type / `undefined` components). The context-based approach keeps alert UIs **data-driven** without a render prop at every call site, as long as the tree is wrapped by the composition-root provider (e.g. **`RootLayout`**). A **per-`Paragraph` provider** is kept so unit tests that render **`Paragraph` / `ParagraphList`** without the full Next layout still get a working renderer for nested alert bodies (see JSDoc on `Paragraph`).

#### Paragraph render context (composition root) 
- Added `ParagraphRenderContext` with `ParagraphRenderProvider`, `useRenderParagraph`, and `useRenderParagraphs`.  
- Added `ParagraphRendererProvider` (maps formatted paragraphs → `<Paragraph />`) and mounted it in **`src/app/layout.tsx`** around `ClientMetrics` / `children`.

#### Alert / alert single
- **`Alert`** and **`AlertSingle`** use `useRenderParagraphs()` for nested paragraph lists and **no longer import** `Paragraph` from `paragraph/template.tsx`.

#### Paragraph dispatcher
- **`Paragraph`** wraps its switch output in a local `ParagraphRenderProvider` with the same renderer as the root, documented for **test convenience** vs. redundant-but-harmless nesting in production.

#### Prereq / follow-on cleanup (reusable alert block body)
- **`AlertBlock`** uses `renderParagraph(content)` instead of branching on WYSIWYG vs expandable text (mutually exclusive `content` union). Marked **`'use client'`** where hooks are required.  
- **`ExpandableText`**: `va-additional-info` gets **`id={\`additional-info-${id}\`}`** so alert and non-alert paths share one implementation.  
- **`globals.css`**: replaced **`.alert-with-additional-info`** with **`va-alert va-additional-info`** for the same margin behavior without a dedicated class on the web component.

#### Tests
- Wrapped **`Alert`**, **`AlertSingle`**, and **`AlertBlock`** tests with `ParagraphRendererProvider` where full rendering is needed.  
- **`ExpandableText`** test asserts the new `id` attribute.

### How to verify

- Pages that use **`paragraph--alert`** / **`paragraph--alert_single`** (including **`AlertSingle`** used directly on node templates, e.g. checklist) still render nested body content under **`RootLayout`**.  
- Reusable alert blocks still render WYSIWYG or expandable body via the paragraph dispatcher.  
- Run unit tests for alert, alert single, alert block, paragraph, and expandable text (or full CI).

### Notes / tradeoffs

- **Implicit dependency:** any environment that renders **`Alert` / `AlertSingle` / `AlertBlock`** without the root provider (or without going through **`Paragraph`**, which supplies the local provider) must wrap tests with **`ParagraphRendererProvider`**.  
- **Double provider in production:** root + per-`Paragraph` is intentional for tests; behavior is equivalent because both providers expose the same mapping.

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/24212


brianseek (approved):
everything works for me.

PR #1892, opened 2026-05-06
Fix archived benefits hub pages causing build errors.

pwolfert:
## Description

- Handle archived benefits hub nodes in VA Form Detail page so they don't crash the build
- Handle archived benefits hub nodes in Campaign Landing page so they don't crash the build

## Ticket

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/24237

## Testing Steps

Should build on Tugboat. If you want to test it locally, pull the latest CMS data as of today, and then visit the following pages: 

- http://localhost:3999/forms/10-10d/
- http://localhost:3999/initiatives/sign-in-securely-with-logingov/


brianseek (approved):
works locally for me

PR #1899, opened 2026-05-08
Opt-in App Router page chrome

pwolfert:
### Summary

This change stops treating the full VA.gov shell (global assets, web components, analytics, polyfills, header/footer layout, etc.) as something every App Router route implicitly inherits from the root layout. Instead, the minimal root layout only provides the universal HTML shell (`lang`, `dir`, charset, shared `metadata`). Routes that need the full experience explicitly wrap their content in `AppRouterPageLayout`, which composes the former root-layout concerns with `PageLayout` (`template.tsx`).

### Motivation

We want explicit, opt-in behavior so we can add lightweight static routes (e.g. a 503 page) that should not load header/footer, Datadog/GTM, web-components bundles, or other chrome. Previously that logic lived in the root App Router `layout`, making “everything everywhere” the default and forcing workarounds for exceptions.

### What changed

- Root `app/layout.tsx` — Reduced to a small universal shell plus shared `metadata` (e.g. `metadataBase`, handheld hints). No global CSS, asset links, or third-party scripts there by default.
- `AppRouterPageLayout` — Holds stylesheets, font preloads, icons, web-components scripts, DAP, polyfills (`nomodule`), announcement root, GTM `noscript`, custom element registration, `ClientMetrics`, and `PageLayout` wired together so callers pass `bannerData` / `footerData` / `megaMenuData` (and optional preview/resource) plus route-specific children.
- Colocation — `ClientMetrics`, `CustomElements`, and `PolyfillsScript` moved from `src/app/` into `src/components/pageLayout/` next to the layout they serve.

### Benefits

- Opt-in chrome — Only routes that import `AppRouterPageLayout` pay for that HTML/JS/CSS and behavior; other routes stay cheap and predictable.
- 503 / maintenance / bare routes — Easier to ship pages that intentionally skip header, footer, mega menu, and analytics without forking layout hacks or duplicating `<html>`.
- Clearer mental model — “Full VA.gov page” is one named component tree instead of “whatever the root layout happened to inject.”
- Cleaner `app/` directory — Fewer client-oriented modules at the router root; layout helpers live with `pageLayout` alongside `PageLayout`.
- Easier evolution — You can add alternate wrappers (e.g. “minimal shell” vs “full chrome”) or route-specific layouts without changing the universal root contract.
- Safer defaults for experiments — New App Router routes won’t accidentally inherit full analytics and asset loading until they deliberately opt in. (This also applies to **our dev routes** when we migrate those over to App Router!)
- Ownership — App Router–specific integration (GTM, RUM, polyfills, WC loader) sits next to the composed layout instead of mixing into the smallest possible root file.

### Notes / follow-ups

- Pages that should look like the rest of VA.gov should continue to use `AppRouterPageLayout` (as home and `not-found` do today).
- Any new static route that must stay minimal should omit `AppRouterPageLayout` and only use the root layout (plus whatever minimal assets that page needs).

## Ticket

No ticket

## Testing Steps

~The statically built home page and 404 page should be exactly the same as before.~ They'll be slightly different in terms of HTML output but should be equivalent.


pwolfert:
Some manual testing:

- Homepage and 404 page look good in Tugboat.
- With homepage flag off, it does not build the homepage file (tested on staging)
- With homepage flag on, it does build the homepage

[Slack thread with some video proof](https://dsva.slack.com/archives/C01SR56755H/p1778281699581279?thread_ts=1778281529.948359&cid=C01SR56755H)

zachbutton (approved):

